PR #15, opened 2023-02-01
Improve Inventory struct

fao89:


fao89:
tested with:
```
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneNode
metadata:
  name: openstackdataplanenode-sample
spec:
  node:
    hostName: openstackdataplanenode-sample.localdomain
    ansibleHost: 192.168.122.18
    ansibleUser: rhosp
    ansiblePort: 80
```

got:
```
kubectl get configmap dataplanenode-openstackdataplanenode-sample-inventory -o yaml
apiVersion: v1
data:
  inventory: |
    all:
      hosts:
        openstackdataplanenode-sample:
          ansible_host: openstackdataplanenode-sample.localdomain
          ansible_user: rhosp
          ansible_port: 80
kind: ConfigMap
metadata:
  creationTimestamp: "2023-02-01T18:44:27Z"
  name: dataplanenode-openstackdataplanenode-sample-inventory
  namespace: default
  resourceVersion: "1226"
  uid: afe624b1-2118-425c-a19b-b6066ad6cb86

```

controllers/openstackdataplanenode_controller.go
```diff
@@ -107,37 +106,17 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
...
-	host := make(map[string]map[string]string)
-	hostVars := make(map[string]string)
-	hostVars["ansible_host"] = instance.Spec.Node.HostName
-	hostVars["ansible_user"] = instance.Spec.Node.AnsibleUser
-	hostVars["ansible_port"] = strconv.Itoa(instance.Spec.Node.AnsiblePort)
-	host[instance.Name] = hostVars
-	all["hosts"] = host
-	inventory["all"] = all
-
+	var inventory = `all:
```
fultonj, line 113: In general I think it's better to build a data structure and then have a function render it into proper yaml. I don't think formatting a string by hand into proper yaml is a pattern we should promote.

fultonj (changes requested):

controllers/openstackdataplanenode_controller.go
fultonj: If I make this change:
```
[fultonj@osp-storage-01 dataplane-operator]$ git diff controllers/openstackdataplanenode_controller.go
diff --git a/controllers/openstackdataplanenode_controller.go b/controllers/openstackdataplanenode_controller.go
index cf04365..00ed9d3 100644
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -162,6 +162,7 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
                cm.Data = map[string]string{
                        "inventory": string(invData),
                }
+               fmt.Printf("inv: %s", string(invData))
                return nil
        })
        if err != nil {
[fultonj@osp-storage-01 dataplane-operator]$ 

```
Then I see the inventory:
```
inv: all:
  hosts:
    openstackdataplanenode-sample:
      ansible_host: openstackdataplanenode-sample.localdomain
      ansible_port: "0"
      ansible_user: ""
```
when I run the controller directly and run:
```
oc kustomize config/samples/ | oc apply -f -
```
So I see my port isn't getting set to 22 as it might be for the role but I assume that will be addressed when we have inheritance working. 

I'm not sure (short commit message) what this patch is trying to accomplish.

fultonj:
> 

I get the same results without your patch however. It looks like this is a re-implementation of the existing functionality which uses a string instead of a struct.

```
fultonj@osp-storage-01 dataplane-operator]$ cat test.yaml 
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneNode
metadata:
  name: openstackdataplanenode-sample-80
spec:
  node:
    hostName: openstackdataplanenode-sample-80.localdomain
    ansibleHost: 192.168.122.18
    ansibleUser: rhosp
    ansiblePort: 80
[fultonj@osp-storage-01 dataplane-operator]$ oc create -f test.yaml
openstackdataplanenode.dataplane.openstack.org/openstackdataplanenode-sample-80 created
[fultonj@osp-storage-01 dataplane-operator]$ oc get configmap dataplanenode-openstackdataplanenode-sample-80-inventory -o yaml
apiVersion: v1
data:
  inventory: |
    all:
      hosts:
        openstackdataplanenode-sample-80:
          ansible_host: openstackdataplanenode-sample-80.localdomain
          ansible_port: "80"
          ansible_user: rhosp
kind: ConfigMap
metadata:
  creationTimestamp: "2023-02-01T20:08:14Z"
  name: dataplanenode-openstackdataplanenode-sample-80-inventory
  namespace: openstack
  resourceVersion: "7744387"
  uid: fa3dd591-d77e-4e2d-beb2-73ff47623862
[fultonj@osp-storage-01 dataplane-operator]$ 
```

controllers/openstackdataplanenode_controller.go
SeanMooney: ya so i would either perefer to build the data structure and then render it too or use go templates to build the inventory file.

fao89: sorry about the commit message, my idea was just replacing the maps

fao89: I'm googling around, marshalling & unmarshalling structs with dynamic keys is painful. Will we use the inventory struct elsewhere? If not, I believe go templates will probably be the best option then

fultonj: > "I'm googling around, marshalling & unmarshalling structs with dynamic keys is painful."

Interesting. Would you mind sharing what you found? Is it just the amount of code required or something else?

fao89: the dynamic key would be the hostname
```yaml
all:
  hosts:
    << dynamic host name  >>:
        hostvars
```        
this is what I found:
https://stackoverflow.com/questions/32147325/how-to-parse-yaml-with-dynamic-key-in-golang


the https://github.com/relex/aini looks interesting, I'll do some experiments

SeanMooney: but the hostname is just the key in a map so 
my type 

type Group struct {
GroupVars map[string]interface{}
Hosts list[Host]
}

should actually be

type Group struct {
GroupVars map[string]interface{}
Hosts map[string]Host
}

where the key is just the hostname

SeanMooney: by the way we never have to parse this in go.

we will only ever need to write this so it make it a little simpler but modelling the dictionary filed as maps of sting to structs should make the hostnames simple to handle
if we did need to parse this.

fao89:
tested with:
```
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneNode
metadata:
  name: openstackdataplanenode-sample
spec:
  node:
    hostName: openstackdataplanenode-sample.localdomain
    ansibleHost: 192.168.122.18
    ansibleUser: rhosp
    ansiblePort: 80
```

got:
```
kubectl get configmap dataplanenode-openstackdataplanenode-sample-inventory -o yaml
apiVersion: v1
data:
  inventory: |
    groups:
      all:
        hosts:
          openstackdataplanenode-sample:
            vars:
              ansible_host: openstackdataplanenode-sample.localdomain
              ansible_port: "80"
              ansible_user: rhosp
kind: ConfigMap
metadata:
  creationTimestamp: "2023-02-02T15:42:35Z"
  name: dataplanenode-openstackdataplanenode-sample-inventory
  namespace: default
  resourceVersion: "1307"
  uid: d9377b17-bb57-47f8-ad53-6751addc3edf

```

controllers/openstackdataplanenode_controller.go
fao89: thank you @SeanMooney 
I've followed your tips, and tested with:
```
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneNode
metadata:
  name: openstackdataplanenode-sample
spec:
  node:
    hostName: openstackdataplanenode-sample.localdomain
    ansibleHost: 192.168.122.18
    ansibleUser: rhosp
    ansiblePort: 80
```

and got:
```
kubectl get configmap dataplanenode-openstackdataplanenode-sample-inventory -o yaml
apiVersion: v1
data:
  inventory: |
    groups:
      all:
        hosts:
          openstackdataplanenode-sample:
            vars:
              ansible_host: openstackdataplanenode-sample.localdomain
              ansible_port: "80"
              ansible_user: rhosp
kind: ConfigMap
metadata:
  creationTimestamp: "2023-02-02T15:42:35Z"
  name: dataplanenode-openstackdataplanenode-sample-inventory
  namespace: default
  resourceVersion: "1307"
  uid: d9377b17-bb57-47f8-ad53-6751addc3edf

```

olliewalsh: The ansible inventory structure has some quirks so probably need custom (un)marshalling logic to get this perfect, but for now something like this might work https://go.dev/play/p/92S1JDkHGDW

SeanMooney: ya that looks pretty good

controllers/openstackdataplanenode_controller.go
```diff
@@ -107,36 +107,39 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
...
+	host.Vars["ansible_user"] = instance.Spec.Node.AnsibleUser
+	host.Vars["ansible_port"] = strconv.Itoa(instance.Spec.Node.AnsiblePort)
+
+	group := Group{}
+	group.Hosts = make(map[string]*Host)
+	group.Hosts[instance.Name] = &host
+
+	inventory := Inventory{}
+	inventory.Groups = make(map[string]*Group)
+	inventory.Groups["all"] = &group
```
fultonj, line 142: This addresses my initial concerns about building a string vs a struct. Thanks, I think that's much better. 

I'm curious what @SeanMooney and @olliewalsh think.

SeanMooney: in terms of a minimal implemnation this is certenly much better then what was there before.

its a lot more readable 

i was originally using map[string]interface{}

so that if you wanted to set something to an int bool or null you could

currently, you don't have the ability to do that as all your types are converted to strings so i would personally prefer to do that over map[string]string for the vars

both i prefer this vs the current state prior to this pr.

SeanMooney: the main delta vs what ollie proposed in https://go.dev/play/p/92S1JDkHGDW
is that host and groups don't know there name and groups cant be nested.

ollies example is more complete in that respect but i don't really like that it need the fixnames function.

over all i think i prefer https://go.dev/play/p/92S1JDkHGDW but I'm not against this either.

fultonj (approved):

controllers/openstackdataplanenode_controller.go
```diff
@@ -107,36 +107,39 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
 	return nil
 }
 
-// Inventory struct
-// TODO: make use of the struct below
-//
-//nolint:unused
+// Host represents ansible host
+type Host struct {
+	Vars map[string]string `yaml:"vars,omitempty"`
```
olliewalsh, line 112: Is this missing the `inline` yaml flag?

IIRC format needs to be:
```
all:
  hosts:
    hostname:
      key1: val1
```

vs

```
all:
  hosts:
    hostname:
      vars:
        key1: val1
```

controllers/openstackdataplanenode_controller.go
```diff
@@ -107,36 +107,39 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
...
-	all struct {
-		hosts struct {
-			host struct {
-				hostVar struct {
-					hostVarValue string
-				}
-			}
-		}
-	}
+	Groups map[string]*Group
```
olliewalsh, line 123: also missing `inline` yaml flag

controllers/openstackdataplanenode_controller.go
olliewalsh: You mean I can't get away with a teeny-tiny little hack like the fixnames function :smiley:

FWIW I've a slightly better version that uses a custom unmarshaller to call fixnames - https://go.dev/play/p/AuJqU08RPmS

+1 for using `map[string]interface{}`. Can get around it with Itoa for the ansible_port but will be necessary I expect if more complex vars need to be set such as NetworkConfig

fao89: I've followed @olliewalsh suggestions here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/17

SeanMooney:
this is an example of what an inventory should look like  for a single host

https://github.com/openstack-k8s-operators/nova-operator/pull/233/files#diff-0ca239fcb0eecfef839bc6f51dd9c49e2e038de5ea0d77c4d3eb1f67c8156220

that might help as a reference
or
https://github.com/openstack-k8s-operators/nova-operator/pull/233/files#diff-732bbbe00e1f100a22f312a959b3fac6979dc99343a1eac9019b9e7c28b47a0f
although i don't expect use to pass much data via the inventory which is why i prefer the former

SeanMooney:
ollie pointed out the fix for this already above by the way 
https://github.com/openstack-k8s-operators/dataplane-operator/pull/15#discussion_r1095138062

PR #16, opened 2023-02-01
Document how inheritance from roles to nodes works

fultonj:
Document how inheritance from roles to nodes works and provide an example.

### DIFF
--- a/docs/inheritance.md
+++ b/docs/inheritance.md
@@ -1,0 +1,100 @@
+# Inheritance
+
+An `OpenStackDataPlaneNode` inherits any attribute of an
+`OpenStackDataPlaneRole` but those attributes may also be overridden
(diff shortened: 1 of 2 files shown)


docs/inheritance.md
```diff
@@ -0,0 +1,100 @@
+# Inheritance
+
+An `OpenStackDataPlaneNode` inherits any attribute of an
+`OpenStackDataPlaneRole` but those attributes my also be overridden on
```
SeanMooney, line 4: nit: `OpenStackDataPlaneRole` but those attributes may also be overridden on

fultonj: Thanks, fixed.

fao89 (approved):
Thank you!

PR #17, opened 2023-02-02
Improve the inventory data structure

fao89:
Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>


controllers/openstackdataplanenode_controller.go
```diff
@@ -19,10 +19,9 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"strconv"
 
 	"github.com/go-logr/logr"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
```
SeanMooney, line 24: i cant comment on this directly later because github suck for code review
but later we call "by the way i cant comment on this directly because github suck for code review
but later we call "invData, err := yaml.Marshal(inventory)"

if instead we add a Marshal function to the inventory object we can just do"

if instead we add a Marshal function to the inventory object we can just do

invData, err := inventory.Marshal()
and we can avoid importing gopkg.in/yaml.v3 in this file.

in other word i think we should make the Inventory struct be able to Marshal and UnMarshal itself and entirely hide the use of the yaml module in the inventory.go file.

SeanMooney (changes requested):

slagle:
I tested this PR along with https://github.com/openstack-k8s-operators/dataplane-operator/pull/15, and it looks like that is working fine for me. Sorry for reverting PR15, I can't entirely tell if this one is ready to merge, and they need to merge at the same time.

@SeanMooney are you still requesting more changes on this one? If not we can revert the revert, then merge this one.

SeanMooney:
so we moved the inventory generation out of the dataplane operator and to AnsibleEE operator
in https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/73
which has the fixes to add inline

https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/73#diff-3fa46e76c6a4f42394b2194abddf735e6792afa3ef91be18452d1c598cc65d23R24

https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/73#diff-3fa46e76c6a4f42394b2194abddf735e6792afa3ef91be18452d1c598cc65d23R24

where it is required and this then just uses that 

so i did not have any other requests for this path i think it was more or less ready.

rather then reverting the revert i would suggest that since this needs a rebase the simple solution is jut to squash merge in the previous patch into this so that we can just apply one commit.


> I tested this PR along with #15, and it looks like that is working fine for me. Sorry for reverting PR15, I can't entirely tell if this one is ready to merge, and they need to merge at the same time.
> 
> @SeanMooney are you still requesting more changes on this one? If not we can revert the revert, then merge this one.

slagle:
Closing this one out in favor of https://github.com/openstack-k8s-operators/dataplane-operator/pull/24 which combines the commits from https://github.com/openstack-k8s-operators/dataplane-operator/pull/15 and https://github.com/openstack-k8s-operators/dataplane-operator/pull/17.

PR #19, opened 2023-02-03
Create nodes from roles which inherit the role template

fultonj:
This patch creates nodes from roles which inherit the role template. It does the following:

- Creates CRs of type OpenStackDataPlaneNodes from the OpenStackDataPlaneRole's DataPlaneNodes list.
- Each node CR is populated with values from the node entry in the DataPlaneNodes list.
- Inheritance is implemented by having the node look up its role and setting Ansible inventory values from the role template in the node configMap.
- Node values take precedence over template values when the inventory configMap values are set.
- Creating a node directly and setting its role will result in the node inheriting that role's nodeTemplate in its inventory.

### DIFF
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -19,6 +19,8 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -112,6 +114,11 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		return result, err
 	}
 
+	instanceRole, err := r.GetInstanceRole(ctx, instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Always patch the instance status when exiting this function so we can
 	// persist any changes.
 	defer func() {
@@ -146,7 +153,6 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		// Register overall status immediately to have an early feedback e.g.
 		// in the cli
 		return ctrl.Result{}, nil
-
 	}
 
 	// check if provided network attachments exist
@@ -180,7 +186,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		}
 	}
 
-	inventoryConfigMap, err := r.GenerateInventory(ctx, instance)
+	inventoryConfigMap, err := r.GenerateInventory(ctx, instance, instanceRole)
 	if err != nil {
 		util.LogErrorForObject(helper, err, fmt.Sprintf("Unable to generate inventory for %s", instance.Name), instance)
 		return ctrl.Result{}, err
@@ -218,7 +224,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
 }
 
 // GenerateInventory yields a parsed Inventory
-func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneNode) (string, error) {
+func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) (string, error) {
 	var (
 		err      error
 		hostName string
@@ -227,11 +233,13 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
 	inventory := ansible.MakeInventory()
 	all := inventory.AddGroup("all")
 	host := all.AddHost(instance.Name)
-	host.Vars["ansible_user"] = instance.Spec.Node.AnsibleUser
 
-	if instance.Spec.Node.AnsiblePort != 0 {
-		host.Vars["ansible_port"] = instance.Spec.Node.AnsiblePort
-	}
+	host.Vars["ansible_user"] = r.GetAnsibleUser(instance, instanceRole)
+	host.Vars["ansible_port"] = r.GetAnsiblePort(instance, instanceRole)
+	host.Vars["managed"] = r.GetAnsibleManaged(instance, instanceRole)
+	host.Vars["management_network"] = r.GetAnsibleManagementNetwork(instance, instanceRole)
+	host.Vars["network_config"] = r.GetAnsibleNetworkConfig(instance, instanceRole)
+	host.Vars["networks"] = r.GetAnsibleNetworks(instance, instanceRole)
 
 	if instance.Spec.AnsibleHost == "" {
 		hostName = instance.Spec.HostName
@@ -241,7 +249,7 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
 	host.Vars["ansible_host"] = hostName
 
 	ansibleVarsData := make(map[string]interface{})
-	err = yaml.Unmarshal([]byte(instance.Spec.Node.AnsibleVars), ansibleVarsData)
+	err = yaml.Unmarshal([]byte(r.GetAnsibleVars(instance, instanceRole)), ansibleVarsData)
 	if err != nil {
 		return "", err
 	}
@@ -280,4 +288,80 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
 	}
 
 	return configMapName, nil
+}
+
+// GetInstanceRole returns the role of a node based on the node's role name
+func (r *OpenStackDataPlaneNodeReconciler) GetInstanceRole(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneNode) (*dataplanev1beta1.OpenStackDataPlaneRole, error) {
+	// Use the instances's role name to get its role object
+	var err error
+	instanceRole := &dataplanev1beta1.OpenStackDataPlaneRole{}
+	err = r.Client.Get(ctx, client.ObjectKey{
+		Namespace: instance.Namespace,
+		Name:      instance.Spec.Role,
+	}, instanceRole)
+	return instanceRole, err
+}
+
+// GetAnsibleUser returns the string value from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleUser(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.AnsibleUser != "" {
+		return instance.Spec.Node.AnsibleUser
+	}
+	return instanceRole.Spec.NodeTemplate.AnsibleUser
+}
+
+// GetAnsiblePort returns the string value from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsiblePort(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.AnsiblePort > 0 {
+		return strconv.Itoa(instance.Spec.Node.AnsiblePort)
+	}
+	return strconv.Itoa(instanceRole.Spec.NodeTemplate.AnsiblePort)
+}
+
+// GetAnsibleManaged returns the string (from boolean) value from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleManaged(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.Managed {
+		return strconv.FormatBool(instance.Spec.Node.Managed)
+	}
+	return strconv.FormatBool(instanceRole.Spec.NodeTemplate.Managed)
+}
+
+// GetAnsibleManagementNetwork returns the string value from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleManagementNetwork(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.ManagementNetwork != "" {
+		return instance.Spec.Node.ManagementNetwork
+	}
+	return instanceRole.Spec.NodeTemplate.ManagementNetwork
+}
+
+// GetAnsibleNetworkConfig returns a JSON string value from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleNetworkConfig(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.NetworkConfig != instanceRole.Spec.NodeTemplate.NetworkConfig {
+		return fmt.Sprintf("{template: %s}", instance.Spec.Node.NetworkConfig)
+
+	}
+	return fmt.Sprintf("{template: %s}", instanceRole.Spec.NodeTemplate.NetworkConfig.Template)
+}
+
+// GetAnsibleNetworks returns a JSON string mapping fixedIP and/or network name to their valules
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleNetworks(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	var network string // the resulting string containing each network
+	for _, netMap := range instance.Spec.Node.Networks {
+		if netMap.FixedIP != "" && netMap.Network == "" {
+			network += fmt.Sprintf("{%s: %s},", "fixedIP", netMap.FixedIP)
+		}
+		if netMap.FixedIP != "" && netMap.Network != "" {
+			network += fmt.Sprintf("{%s: %s, %s: %s},",
+				"fixedIP", netMap.FixedIP, "network", netMap.Network)
+		}
+	}
+	return fmt.Sprintf("[%s]", strings.TrimSuffix(network, ","))
+}
+
+// GetAnsibleVars returns a string value of ansible vars from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleVars(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.AnsibleVars != "" {
+		return instance.Spec.Node.AnsibleVars
+	}
+	return instanceRole.Spec.NodeTemplate.AnsibleVars
 }
--- a/controllers/openstackdataplanerole_controller.go
+++ b/controllers/openstackdataplanerole_controller.go
@@ -23,10 +23,12 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/go-logr/logr"
@@ -77,7 +79,6 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
-
 	err = r.ReconcileNodes(ctx, instance)
 	if err != nil {
 		util.LogErrorForObject(helper, err, fmt.Sprintf("Unable to reconcile nodes for %s", instance.Name), instance)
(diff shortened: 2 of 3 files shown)


controllers/openstackdataplanerole_controller.go
```diff
@@ -85,11 +86,78 @@ func (r *OpenStackDataPlaneRoleReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+		OpenStackDataPlaneNodeName := fmt.Sprintf("dataplanenode-inheritance-%d", i)
+
+		node := &dataplanev1beta1.OpenStackDataPlaneNode{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      OpenStackDataPlaneNodeName,
+				Namespace: instance.Namespace,
+			},
+		}
+		// Set new node with values from its entry on the DataPlaneNodes list
+		node.Spec.Role = instance.Name
```
fultonj, line 109: Because the node is getting its role set here (line 109), I could changes lines 114 to 144 in this file to remove every `else` but keep every `if` so that each node is created with node properties only (besides the single reference to its role from line 109). Then on the node level, it can look up the value from the role and pass it to the Ansible call without setting it directly in the object itself. This assumes that we have the Ansible call from the node level but no Ansible call from the role level. I can try that in my next patch set.

controllers/openstackdataplanerole_controller.go
```diff
@@ -85,11 +86,78 @@ func (r *OpenStackDataPlaneRoleReconciler) SetupWithManager(mgr ctrl.Manager) er
...
 // ReconcileNodes ensure the desired state for the nodes
 func (r *OpenStackDataPlaneRoleReconciler) ReconcileNodes(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneRole) error {
-	// loop over r.Spec.DataPlaneNodes:
+	// loop over instance.Spec.DataPlaneNodes:
 	//   for each node:
 	//     (1) complete node.Spec based r.Spec.nodeTemplate, and values set on the
 	//         node (values on the node take precedence over those from the template)
 	//     (2) Create a CR of OpenStackDataPlaneNode from the node
 
+	for i := range instance.Spec.DataPlaneNodes {
```
fmount, line 95: Do you need the index here or is it better using the current item returned by the iterator? 

e.g.

```
for i, cur := range instance.Spec.DataPlaneNodes {
 /***
  * e.g. node.Spec.Node.AnsibleHost = cur.Node.AnsibleHost
  ***/ 
}
```

This would help to make L114-L144 more robust.
I see you need the index on L100, I'm wondering if the index is required for node Name generation or we should have a diff solution.

controllers/openstackdataplanerole_controller.go
fultonj: I've basically done this in the latest patch set. An Ansible call is not necessary to prove the inheritance works. Because each node automatically creates its own inventory in a configmap, you can observe the inventory configmap per node to see inheritance working.

controllers/openstackdataplanerole_controller.go
fultonj: OK, I made that change.

fultonj:
Still needs an update to correctly handle dataPlaneNodes with nodeFrom entries.

controllers/openstackdataplanenode_controller.go
```diff
@@ -135,16 +141,59 @@ type Inventory struct {
...
+	} else {
+		hostVars["management_network"] = instanceRole.Spec.NodeTemplate.ManagementNetwork
+	}
+	if instance.Spec.Node.NetworkConfig != instanceRole.Spec.NodeTemplate.NetworkConfig {
+		hostVars["network_config"] = fmt.Sprintf("{template: %s}",
+			instance.Spec.Node.NetworkConfig)
+	} else {
+		hostVars["network_config"] = fmt.Sprintf("{template: %s}",
+			instanceRole.Spec.NodeTemplate.NetworkConfig.Template)
+	}
```
slagle, line 179: what about using getters here? Each "if" statement would be in a getter function:

`func (instance *dataplanev1beta1.OpenStackDataPlaneNode) GetAnsibleUser() string { ... }`

controllers/openstackdataplanerole_controller.go
```diff
@@ -96,11 +97,60 @@ func (r *OpenStackDataPlaneRoleReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+		OpenStackDataPlaneNodeName := fmt.Sprintf("%s-node-%d", instance.Name, i)
+		newNode := &dataplanev1beta1.OpenStackDataPlaneNode{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      OpenStackDataPlaneNodeName,
+				Namespace: instance.Namespace,
+			},
+		}
+		// Set new node with values from its entry on the DataPlaneNodes list
+		newNode.Spec.Role = instance.Name
+		newNode.Spec.Node.AnsibleHost = node.Node.AnsibleHost
```
slagle, line 116: AnsibleHost might not be required, so we should probably add a check for it. When we generate the inventory, we use AnsibleHost if set, otherwise HostName.
And could also make HostName optional, and default to just Name.

docs/inheritance.md
```diff
@@ -98,3 +98,30 @@ updating only that node (e.g. with `oc edit`). In that case we'd see
...
+```
+Then the role openstackdataplanerole-sample will be updated so that
+its `dataPlaneNodes` list contains the new node.
+```yaml
+  dataPlaneNodes:
+    - name: openstackdataplanenode-sample-3-from
+      nodeFrom: openstackdataplanenode-sample-3-from
+```
+and the node openstackdataplanenode-sample-3-from will inherit values
+from `nodeTemplate`.
```
slagle, line 127: :+1: that's how I'm thinking it will work as well

controllers/openstackdataplanerole_controller.go
```diff
@@ -96,11 +97,60 @@ func (r *OpenStackDataPlaneRoleReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+
+		_, err := controllerutil.CreateOrUpdate(ctx, r.Client, newNode, func() error {
+			newNode.TypeMeta = metav1.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "OpenStackDataPlaneNode",
+			}
+			newNode.ObjectMeta = metav1.ObjectMeta{
+				Name:      OpenStackDataPlaneNodeName,
+				Namespace: instance.Namespace,
+			}
```
gibizer, line 147: I think the newNode.Spec needs to be set within this function. The doc says:
```
// CreateOrUpdate creates or updates the given object in the Kubernetes
// cluster. The object's desired state must be reconciled with the existing
// state inside the passed in callback MutateFn.
//
// The MutateFn is called regardless of creating or updating an object.
```

controllers/openstackdataplanenode_controller.go
fmount: As per our conversation, given the scope of this change it's ok having a getter like:

```
func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleUser(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {...}
```

that lives in the controller, where the two data structures can be processed.

controllers/openstackdataplanenode_controller.go
```diff
@@ -166,3 +183,71 @@ func (r *OpenStackDataPlaneNodeReconciler) ConfigureNetwork(ctx context.Context,
...
+	return instanceRole.Spec.NodeTemplate.AnsibleUser
+}
+
+// GetAnsiblePort retruns the string value from the template unless it is set in the node
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsiblePort(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.AnsiblePort > 0 {
+		return strconv.Itoa(instance.Spec.Node.AnsiblePort)
+	}
+	return strconv.Itoa(instanceRole.Spec.NodeTemplate.AnsiblePort)
```
fao89, line 212: Should we raise an error when fields are empty in both node and role?
```suggestion
	if instanceRole.Spec.NodeTemplate.AnsiblePort > 0 {
		return strconv.Itoa(instanceRole.Spec.NodeTemplate.AnsiblePort), nil
	}
	return nil, fmt.Errorf("Please provide the AnsiblePort")
```

fultonj: I don't disagree about trying to capture that type of error. However, I think we should do it here in the controller and instead do it as a validation in the API.

https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanenode_types.go#L70

fmount: I agree with @fultonj that we might want to do that check elsewhere.
I think we don't have to check if the value is `> 0`: in the API spec [1] we can use `kubebuilder` to validate the provided input, which can be set as `22` as default.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanenode_types.go#L70

fultonj: I'm doing this check (>0) because I want to know if the value has been set to something other than the default. 

I was thinking of doing a separate PR to add the validation and set a new default of 22. Though, if the ansible port is not specified in an inventory, I think ansible chooses 22. Thus, we don't want to always set it to 22 in the inventory. We only want to set it in the inventory if it's non standard. Otherwise it will just be noise. We can add a PR with the validation and continue this discussion there.

controllers/openstackdataplanenode_controller.go
```diff
@@ -166,3 +187,99 @@ func (r *OpenStackDataPlaneNodeReconciler) ConfigureNetwork(ctx context.Context,
...
+}
+
+// EnsureNodeOnRoleList will ensure that a node is on its role's dataPlaneNodes list
+func (r *OpenStackDataPlaneNodeReconciler) EnsureNodeOnRoleList(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) error {
+	// 1. When a node is created by a role it will already be on the list
+	// 2. When a node is directly created with a role name, it won't be on the list
+	// In the second case we have to add the node directly and set nodeFrom field
+
+	var err error
+	var nodeOnList bool // default is false
```
gibizer, line 266: you can  codify that comment  with `nodeOnList := false` and that is even shorter :)

controllers/openstackdataplanenode_controller.go
```diff
@@ -118,15 +129,25 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
 }
 
 // GenerateInventory yields a parsed Inventory
-func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneNode) error {
+func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) error {
```
bogdando, line 132: wrt batching/aggregating inventories for future use - we might also want to put each node into a children  group defined from its parent role CR's metadata name.

then, after individual inventories for DP Node CRs are created, we can leverage dynamic merging feature for AEE. For example, there could be a volumeMount for each inventory file in AEE job pod aggregated in the  '/inventories' directory. Then it can be batched just as running ansible with -i /inventories

fultonj: > wrt batching/aggregating inventories for future use - we might also want to put each node into a children group defined from its parent role CR's metadata name.

So right now there's one inventory created per node. To do that we'd need an inventory created per role. I like this idea, though I think we should do it in a separate PR after we have the Ansible execution working with 1 node to 1 inventory. 

> then, after individual inventories for DP Node CRs are created, we can leverage dynamic merging feature for AEE. For example, there could be a volumeMount for each inventory file in AEE job pod aggregated in the '/inventories' directory. Then it can be batched just as running ansible with -i /inventories

Yeah, I think that would good; something to do in a future PR.

controllers/openstackdataplanenode_controller.go
fultonj: Thanks, I've done that in my code and it will be reflected in the next patch set that I push.

controllers/openstackdataplanenode_controller.go
bogdando: I meant to say we can still create a single node inventories, just adding each node there into its corresponding group. Not combining all nodes of a role into inventory file - but rely on automatic merging it by ansible

fultonj: > I meant to say we can still create a single node inventories, just adding each node there into its corresponding role group. Not combining all nodes of a role into inventory file - but rely on automatic merging it by ansible

So when I add one node (which creates an inventory) I also add an inventory group based on it's role. If N nodes are added then we have N inventory config-maps. One Ansible runner could then mounts all N inventories in a directory and the target it of the run could be all members of that group. Ansible will then merge the inventories because of the way `ansible -i /path/to/dir` behaves.

controllers/openstackdataplanenode_controller.go
```diff
@@ -236,3 +248,139 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
...
+		// seem to define what will be in the mutate function.
+		// This works:
+		//
+		//   instanceRole.Spec.DataPlaneNodes[0] = *nodeSection
+		//
+		// But I want to append to len()+1
+		// But that fails since cap == len
+		// I know append() will increase the slice's cap, e.g.
+		//
+		//   instanceRole.Spec.DataPlaneNodes.append(*nodeSection)
```
fultonj, line 365: Looks like `OpenStackDataPlaneNodeSpec` needs an append method.

fao89: not sure it works:
```go
instanceRole.Spec.DataPlaneNodes = append(instanceRole.Spec.DataPlaneNodes, nodeSection)
```

fultonj: @fao89 I tried that but I got the following:
```
# github.com/openstack-k8s-operators/dataplane-operator/controllers
vet: controllers/openstackdataplanenode_controller.go:368:79: cannot use nodeSection (variable of type *"github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1".OpenStackDataPlaneNodeSpec) as "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1".OpenStackDataPlaneNodeSpec value in argument to append
```
I'm working on updating one of the following to define a function to do the append on the list.

```
	modified:   api/v1beta1/openstackdataplanenode_types.go
	modified:   api/v1beta1/openstackdataplanerole_types.go
```

fultonj: This worked:
`instanceRole.Spec.DataPlaneNodes = append(instanceRole.Spec.DataPlaneNodes, *nodeSection)`

Thank you @fao89 !

controllers/openstackdataplanenode_controller.go
```diff
@@ -236,3 +248,139 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
...
+	var err error
+	instanceRole := &dataplanev1beta1.OpenStackDataPlaneRole{}
+	err = r.Client.Get(ctx, client.ObjectKey{
+		Namespace: instance.Namespace,
+		Name:      instance.Spec.Role,
+	}, instanceRole)
+	return instanceRole, err
+}
+
+// GetAnsibleUser retruns the string value from the template unless it is set in the node
```
odyssey4me, line 264: s/retruns/returns/ (several more instances of this later on)

docs/inheritance.md
```diff
@@ -98,3 +98,35 @@ updating only that node (e.g. with `oc edit`). In that case we'd see
 an `ansiblePort` set directly in that node's CR. This allows the user
 to change the `nodeTemplate` after creation and once reconciliation is
 completed all existing nodes will inherit the new value.
+
```
odyssey4me, line 101: Slight tangent. We should perhaps add a note that any top level property in the node overrides the whole property in the role. In other words - there is no need merging of any sub-properties.

This is by design the same as the way Ansible merges variables to help keep it intuitive and simple.

fultonj:
Here's some examples that I tested: https://github.com/fultonj/zed/blob/main/inheritance/README.md

fao89 (commented):
+2

fultonj:
Just a rebase after PR56 merged.

slagle:
This looks pretty good to me. I tested it out and I do get an error though trying to add a node to an existing role, e.g., the role is already created, and then I create a DataPlaneNode CR with the role set as the already existing role.

```
1.6776071386009212e+09  ERROR   controllers.OpenStackDataPlaneRole      Unable to reconcile nodes for openstackdataplanerole-sample     {"ObjectType": "*v1beta1.OpenStackDat
aPlaneRole", "ObjectNamespace": "openstack", "ObjectName": "openstackdataplanerole-sample", "error": "openstackdataplanenodes.dataplane.openstack.org \"openstackdataplanerol
e-sample-node-0\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update"}                                                                
github.com/openstack-k8s-operators/lib-common/modules/common/util.LogErrorForObject                                                                                          
        /home/jslagle/go/pkg/mod/github.com/openstack-k8s-operators/lib-common/modules/common@v0.0.0-20230220181542-93a9b32e5ec9/util/log.go:65                              
github.com/openstack-k8s-operators/dataplane-operator/controllers.(*OpenStackDataPlaneRoleReconciler).Reconcile                                                              
        /home/jslagle/code/github/openstack-k8s-operators/dataplane-operator/controllers/openstackdataplanerole_controller.go:84                                             
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile                                                                                               
```

controllers/openstackdataplanenode_controller.go
```diff
@@ -266,3 +278,111 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
...
+			break
+		}
+		if node.AnsibleHost == instance.Spec.AnsibleHost {
+			nodeOnList = true
+			break
+		}
+	}
+	if !nodeOnList {
+		nodeSection := &dataplanev1beta1.OpenStackDataPlaneNodeSpec{}
+		_, err := controllerutil.CreateOrUpdate(ctx, r.Client, instanceRole, func() error {
```
fultonj, line 374: This needs to not fail as seen here  https://paste.opendev.org/show/bSM4Uy4xUQutVMFQH4lD/

fao89: see:
- https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/patch_csv.yaml#L12-L15
- https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/bindata_util/merge.go#L7-L21

fao89: you probably need `MergeMetadataForUpdate` before calling the `Update`

fultonj: Maybe this is creating an infinite loop: reconcilenode --> reconcilerole which calls reconcilenode.

I'm going to remove `EnsureNodeOnRoleList` from the next update to this PR since it's not essential for the `dataPlaneNodes` list to reflect all of the nodes within the role.

fultonj: Thanks for the suggestions @fao89 maybe we can get `EnsureNodeOnRoleList` in a follow up PR.

fultonj:
> This looks pretty good to me. I tested it out and I do get an error though trying to add a node to an existing role, e.g., the role is already created, and then I create a DataPlaneNode CR with the role set as the already existing role.
> 
> ```
> 1.6776071386009212e+09  ERROR   controllers.OpenStackDataPlaneRole      Unable to reconcile nodes for openstackdataplanerole-sample     {"ObjectType": "*v1beta1.OpenStackDat
> aPlaneRole", "ObjectNamespace": "openstack", "ObjectName": "openstackdataplanerole-sample", "error": "openstackdataplanenodes.dataplane.openstack.org \"openstackdataplanerol
> e-sample-node-0\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update"}                                                                
> github.com/openstack-k8s-operators/lib-common/modules/common/util.LogErrorForObject                                                                                          
>         /home/jslagle/go/pkg/mod/github.com/openstack-k8s-operators/lib-common/modules/common@v0.0.0-20230220181542-93a9b32e5ec9/util/log.go:65                              
> github.com/openstack-k8s-operators/dataplane-operator/controllers.(*OpenStackDataPlaneRoleReconciler).Reconcile                                                              
>         /home/jslagle/code/github/openstack-k8s-operators/dataplane-operator/controllers/openstackdataplanerole_controller.go:84                                             
> sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile                                                                                               
> ```
> 
> I've not seen anything like that before, I'm looking into the cause.

The latest PR no longer hits the above issue. As per our conversation I've removed `EnsureNodeOnRoleList()`. Maybe something like that can happen in a future PR if we figure out how to avoid the above.

fao89 (commented):
+2

slagle (changes requested):
instance.Spec.Node.AnsibleVars needs to use a Getter as well, so that it gets inherited from the role if set. This is at line 237 (I couldn't figure out how to comment inline on a line not part of the diff :shrug: )

fultonj:
> instance.Spec.Node.AnsibleVars needs to use a Getter as well, so that it gets inherited from the role if set. This is at line 237 (I couldn't figure out how to comment inline on a line not part of the diff shrug )

This has now been addressed and I tested it with the following example:

https://github.com/fultonj/zed/commit/fc0af3dcc913f03abfd0b0f0444faba9733e1d1b

controllers/openstackdataplanenode_controller.go
```diff
@@ -241,7 +249,7 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
 	host.Vars["ansible_host"] = hostName
 
 	ansibleVarsData := make(map[string]interface{})
-	err = yaml.Unmarshal([]byte(instance.Spec.Node.AnsibleVars), ansibleVarsData)
+	err = yaml.Unmarshal([]byte(r.GetAnsibleVars(instance, instanceRole)), ansibleVarsData)
```
fultonj, line 252: @slagle here's the Getter you suggested.

slagle: Great, my test looks like it's working now as well. :+1:

slagle (approved):

PR #22, opened 2023-02-09
Revert "Improve Inventory struct"

slagle:
Reverts openstack-k8s-operators/dataplane-operator#15

"groups" is not a valid key in the inventory, the groups are just listed as keys.
"vars" is not a key under an individual host either, only a group. Otherwise, ansible interprets "vars" as a host variable called vars.

### DIFF
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -118,39 +118,36 @@ func (r *OpenStackDataPlaneNodeReconciler) Provision(ctx context.Context, instan
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


rebtoor (approved):

SeanMooney:
i would have prefered to not revert this and just add the missing "inline"

but i guess we can just do everything in https://github.com/openstack-k8s-operators/dataplane-operator/pull/17

slagle:
> i would have prefered to not revert this and just add the missing "inline"
> 
> but i guess we can just do everything in #17

sorry, it wasn't clear to me what the fix was.

PR #23, opened 2023-02-09
Add initial ansible execution and use it for network config

slagle:
Adds the initial handling of ansible execution, within
pkg/util/ansible_execution.go, which will create CR's of AnsibleEE.

The network configuration illustrates how the ansible execution is used.
The deployment package will contain other files that trigger ansible
executions for different stages of the deployment.


slagle:
This includes the revert at https://github.com/openstack-k8s-operators/dataplane-operator/pull/22 just so I could test this out for the time being. I will rebase it once that PR is merged.

config/crd/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
fao89, line 6: it is reverting the kubebuilder version, please remove your `./bin/controller-gen` so you can get an updated one

controllers/openstackdataplanenode_controller.go
```diff
@@ -60,7 +62,7 @@ type OpenStackDataPlaneNodeReconciler struct {
 // For more details, check Reconcile and its Result here:
```
fao89, line 62: we need to add (line 50)
```
//+kubebuilder:rbac:groups=ansibleee.openstack.org,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete
```

config/crd/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
slagle: Done.

controllers/openstackdataplanenode_controller.go
slagle: Done.

fao89 (approved):

PR #25, opened 2023-02-10
OpenStackDataPlaneNode role should not be a templateRef

fultonj:
OpenStackDataPlaneNode types had a typo in its JSON tag where the role string was set to a templateRef. This was from an earlier prototype. Set it to to the role value.


fultonj:
This change fixed this problem for me for the role: https://paste.opendev.org/show/818684/

fao89 (approved):

PR #26, opened 2023-02-10
OpenStackDataPlaneNode Network should not be a template

fultonj:
Change OpenStackDataPlaneNode NetworksSection JSON tag from "template" to "network". Without this change, the network name "ctlplane" in the `config` sample for this type is not retained when the sample is created.


fultonj:
Fixes the network problem in https://paste.opendev.org/show/818684/

fao89 (approved):

PR #33, opened 2023-02-15
Deployment coordination for Network

slagle:
- Rework AnsibleExecution to use CreateOrPatch
- Add Deploy boolean and refactor deployment calls
- Add ValidateNetwork and coordination with ConfigureNetwork


pkg/deployment/deployment.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+
+	// GetConfigureNetwork
+	configureNetwork, err := GetConfigureNetwork(obj, ctx, helper)
+	if err != nil {
+		return err
+	}
+
+	// Wait on ConfigureNetwork to be complete
```
slagle, line 46: still need to add a wait here, instead of jumping ahead and checking the condtions.

pkg/util/ansible_execution.go
```diff
@@ -30,80 +36,150 @@ import (
...
+	return ansibleEE, nil
+
+}
+
+// GetAnsibleExecutionJob
+func GetAnsibleExecutionJob(instance client.Object, ctx context.Context, helper *helper.Helper, label string) (*batchv1.Job, error) {
+	ansibleEE, err := GetAnsibleExecution(instance, ctx, helper, label)
+	if err != nil {
+		return nil, err
+	}
```
slagle, line 176: need a wait here as there is no guarantee the Job is already created.

fultonj:
The basic idea here looks good to me. I'd like to test it in my environment.

slagle:
The latest push reworks this a fair bit from what I had originally. I've switched it over to use conditions, with requeue'ing for the reconcile loop. I think this is more of an accepted pattern in operator, instead of waits and explicit watches built into the operator. This is working for me pretty well now. And translating the rest of the playbooks over to this pattern I think would be pretty straightforward.

pkg/deployment/deployment.go
```diff
@@ -0,0 +1,124 @@
...
+		} else {
+			log.Info("ConfigureNetworkReadyCondition not yet ready, requeueing")
+			return ctrl.Result{RequeueAfter: time.Second * 2}, nil
+		}
+
+	} else if status.Conditions.IsTrue(dataplanev1beta1.ConfigureNetworkReadyCondition) {
+		log.Info("ConfigureNetworkReadyCondition already ready")
+	}
+
+	// ValidateNetwork
```
slagle, line 81: the 2 sections here for ConfigureNetwork and ValidateNetwork can be refactored out into a separate function I think, where we'd just need to pass the condition names and label.

api/v1beta1/conditions.go
```diff
@@ -0,0 +1,43 @@
...
+	// ValidateNetworkReadyCondition Status=True condition indicates if the
+	// network validation is finished and successful.
+	ValidateNetworkReadyCondition condition.Type = "ValidateNetworkReady"
+
+	// ValidateNetworkReadyMessage ready
+	ValidateNetworkReadyMessage = "ValidateNetwork ready"
+
+	// ValidateNetworkReadyWaitingMessage not yet ready
+	ValidateNetworkReadyWaitingMessage = "ValidateNetwork not yet ready"
+)
```
fao89, line 43: 👏🏽

pkg/util/ansible_execution.go
fao89: now you can just watch the conditions, right?

fao89 (commented):
it is looking good, I would approve if it wasn't a draft

pkg/deployment/deployment.go
slagle: resolved with using conditions in latest patch

fao89:
from `ReadyCondition` https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md#type
I believe we need an `IsReady` method:
https://github.com/openstack-k8s-operators/keystone-operator/blob/master/api/v1beta1/keystoneapi_types.go#L253-L257

slagle:
> from `ReadyCondition` in https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md#type I believe we need an `IsReady` method: https://github.com/openstack-k8s-operators/keystone-operator/blob/master/api/v1beta1/keystoneapi_types.go#L253-L257

I added this in the last push.

fao89 (commented):
+2

I'm not formally approving because of the bot

api/v1beta1/openstackdataplanenode_types.go
```diff
@@ -38,6 +39,10 @@ type OpenStackDataPlaneNodeSpec struct {
 	// +kubebuilder:validation:Optional
 	// AnsibleHost SSH host for Ansible connection
 	AnsibleHost string `json:"ansibleHost,omitempty"`
+
+	// +kubebuilder:default=true
+	// Deploy boolean to trigger ansible execution
+	Deploy bool `json:"deploy"`
```
olliewalsh, line 45: What is the purpose of this? Dev/testing?

api/v1beta1/conditions.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2022.
```
olliewalsh, line 2: nit: 2023 now

api/v1beta1/openstackdataplanenode_types.go
slagle: Partially, yes. The nova-operator is also adding something similar. We may eventually need something finer grained, but for the time being it also allows for orchestration between baremetal and the ansible, in that you might just want to deploy baremetal, and then stop. Then deploy the ansible at a later time.

api/v1beta1/conditions.go
slagle: Done.

fao89 (approved):

PR #36, opened 2023-02-17
Add OS config

fao89:
ref: https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/edpm_ansible/playbooks/deploy-edpm-os-configure.yml

closes OSP-22319


slagle:
This is looking pretty good. If we like what's in https://github.com/openstack-k8s-operators/dataplane-operator/pull/33, then this one would need to be redone to match that pattern, but I think that would be pretty straightforward.

fao89:
> This is looking pretty good. If we like what's in #33, then this one would need to be redone to match that pattern, but I think that would be pretty straightforward.

I'm also working on https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/88 to get a playbook translation

pkg/dataplane/deployment/os.go
```diff
@@ -0,0 +1,106 @@
...
+
+	role := ansibleeev1alpha1.Role{
+		Name:           "Deploy EDPM Operating System Configure",
+		Hosts:          "all",
+		Strategy:       "free",
+		Become:         true,
+		AnyErrorsFatal: true,
+		Tasks: []ansibleeev1alpha1.Task{
+			{
+				Name: "Configure edpm_podman",
```
fmount, line 40: I would refactor this function to avoid duplicating the block:

```
{
    Name: "Configure edpm_podman",
    ImportRole: ansibleeev1alpha1.ImportRole{
        Name:      "<ROLE_NAME>",
        TasksFrom: "<ROLE_YAML_FILE_WITH_TASKS>",
     },
    Tags: []string{"<TAGS>"},
},
```

and provide a data structure (e.g. a `map`) as input of this function that you can iterate, producing the blocks you need.
Not 100% convinced the right way is to do the `OS` part this way, I would have created a playbook containing all the roles added here, and called it from this point w/ a set of `tags/variables` to properly drive its execution, but if you prefer to keep this approach is fine, just some optimizations are required I guess

fmount (changes requested):

pkg/dataplane/deployment/os.go
fao89: I don't think a map would provide too much value,   we would be moving the duplication from `ansibleeev1alpha1.Task` to `map[string]string` 
```go
var importRole = []map[string]string {
        {
	  "ROLE_NAME": "name",
          "ROLE_YAML_FILE_WITH_TASKS": "yaml",
          "TAGS": "tag1, tag2",
        },
       {
	  "ROLE_NAME": "name2",
          "ROLE_YAML_FILE_WITH_TASKS": "yaml2",
          "TAGS": "tag1, tag2",
        },
    }
  ```
  
  About creating a playbook, do you mean something like this? https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/examples/openstack-ansibleee-play.yaml#L8-L14

Comment (fmount) [reply pkg/dataplane/deployment/os.go:40]: > I don't think a map would provide too much value, we would be moving the duplication from `ansibleeev1alpha1.Task` to `map[string]string`
> 
> ```go
> var importRole = []map[string]string {
>         {
> 	  "ROLE_NAME": "name",
>           "ROLE_YAML_FILE_WITH_TASKS": "yaml",
>           "TAGS": "tag1, tag2",
>         },
>        {
> 	  "ROLE_NAME": "name2",
>           "ROLE_YAML_FILE_WITH_TASKS": "yaml2",
>           "TAGS": "tag1, tag2",
>         },
>     }
> ```
> 
> About creating a playbook, do you mean something like this? https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/examples/openstack-ansibleee-play.yaml#L8-L14

As you wish, in general I prefer working with a more complex data structure instead of making the code verbose and with a lot of duplication, but I guess it's just an opinionated approach.
e.g.

```
type Task struct {
	Name string
	Path string
	Tags string
}

	tx := Task { Name:"podman", Path: "podman.yaml", Tags: []string{"podman", "edpm_os"} }
	...
	...

	var v = map[string]Tu {
		...
                ...
		"taskX": tx,
		...
		...
	}

```
What I meant regarding the playbook is that calling something like [1], where `roles` are imported on a file like [1] (which is just an example) allows you to just call an `os.yaml` as a playbook instead of building it in a Go function.


[1] https://github.com/openstack/tripleo-ansible/blob/master/tripleo_ansible/playbooks/cli-deployed-ceph.yaml#L191-L287

Comment (fao89) [reply pkg/dataplane/deployment/os.go:40]: Gotcha! I'm following the pattern @slagle started with network https://github.com/openstack-k8s-operators/dataplane-operator/pull/33 
I believe we should have the same pattern for all the playbooks: https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/edpm_ansible/playbooks
@slagle what do you think about the playbook approach?

Comment (fao89) [reply pkg/dataplane/deployment/os.go]: I'm seeing other playbooks that have tasks that use ansible modules: https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/edpm_ansible/playbooks/deploy-edpm-os-install.yml
So going with playbook seems to be the best way

FYI: @slagle @fultonj @rebtoor

Comment (slagle) [reply pkg/dataplane/deployment/os.go]: I'm fine with any approach to avoid the code duplication. It's really just a list of role names and tasks_from.

As for playbooks vs roles, I'd tend to avoid playbooks. We want to orchestrate from the operator, and be able to configure how we call the roles. If we used playbooks, we'd have to use playbook templates anyway. This will allow us to do more advanced things later on, like executing roles in parallel (multiple AnsibleEE's). I'm not saying it will *never* be playbooks, or other more specific configurations like ceph can't use playbooks, just that as of now, I'd like to drive it all with the existing standalone roles for the tasks that dataplane-operator is directly orchestrating.

For the tasks in the playbooks from edpm-ansible, we just need to move those into an existing standalone role in edpm-ansible. For the ones from os-install in particular, I'd just copy them straight into the edpm_bootstrap role.

Comment (fao89) [reply pkg/dataplane/deployment/os.go]: I'm moving them: https://github.com/openstack-k8s-operators/edpm-ansible/pull/45

File: pkg/deployment/os.go [hunk 2]
```diff
@@ -0,0 +1,214 @@
...
+		{
+			Name:          "Install edpm_ovn",
+			RoleName:      "edpm_ovn",
+			RoleTasksFrom: "install.yml",
+			Tags:          []string{"edpm_ovn"},
+		},
+	}
+	role := ansibleeev1alpha1.Role{
+		Name:           "Deploy EDPM Operating System Install",
+		Hosts:          "{{ primary_role_name | default([]) }}:{{ deploy_target_host | default('overcloud') }}",
```
slagle, line 70: this should just be "all". These other group names are carry overs from TripleO that we don't need

pkg/deployment/os.go
```diff
@@ -0,0 +1,214 @@
...
+	tasks := []dataplaneutil.Task{
+		{
+			Name:          "Install edpm_bootstrap",
+			RoleName:      "edpm_bootstrap",
+			RoleTasksFrom: "bootstrap.yml",
+			Tags:          []string{"edpm_bootstrap"},
+		},
+		{
+			Name:     "Install edpm_kernel",
+			RoleName: "edpm_kernel",
```
slagle, line 40: missing a RoleTasksFrom here which causes a traceback:

`1.6770903274761724e+09  ERROR   Unable to execute Ansible for InstallOS {"controller": "openstackdataplanenode", "controllerGroup": "dataplane.openstack.org", "controllerKin
d": "OpenStackDataPlaneNode", "OpenStackDataPlaneNode": {"name":"openstackdataplanenode-sample-deployment","namespace":"openstack"}, "namespace": "openstack", "name": "opens
tackdataplanenode-sample-deployment", "reconcileID": "1d465598-2b43-4528-b357-5eca6004f550", "error": "OpenStackAnsibleEE.ansibleee.openstack.org \"dataplane-deployment-inst
all-osgg26f\" is invalid: spec.roles.tasks[1].import_role.tasks_from: Required value"}                                                                                       
github.com/openstack-k8s-operators/dataplane-operator/pkg/deployment.InstallOS                                                                                               
        /home/jslagle/code/github/openstack-k8s-operators/dataplane-operator/pkg/deployment/os.go:80                                                                         
github.com/openstack-k8s-operators/dataplane-operator/pkg/deployment.ConditionalDeploy                                                                                       
        /home/jslagle/code/github/openstack-k8s-operators/dataplane-operator/pkg/deployment/deployment.go:119                                                                
github.com/openstack-k8s-operators/dataplane-operator/pkg/deployment.Deploy                                                                                                  
`

api/v1beta1/conditions.go
```diff
@@ -47,4 +47,34 @@ const (
 
 	// ValidateNetworkReadyWaitingMessage not yet ready
 	ValidateNetworkReadyWaitingMessage = "ValidateNetwork not yet ready"
+
+	// InstallOSReadyCondition Status=True condition indicates if the
+	// OS configuration is finished and successful.
+	InstallOSReadyCondition condition.Type = "InstallOSReady"
```
slagle, line 53: The 3 new Ready conditions need to be initialized at https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanenode_controller.go#L114

see "Initialize Status" comment in the code

slagle (changes requested):
With my recommended changes, this worked. nice!
I did notice that we also need to switch from the WaitingMessage to the ReadyMessage once a condition is set to true, but this needs to be fixed for all deployments, and not specific to your patch.

api/v1beta1/conditions.go
fao89: done!

pkg/deployment/os.go
fao89: done!

slagle:
> With my recommended changes, this worked. nice! I did notice that we also need to switch from the WaitingMessage to the ReadyMessage once a condition is set to true, but this needs to be fixed for all deployments, and not specific to your patch.

I switched to using the right ReadyMessage's here:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/47

slagle (approved):

pkg/deployment/os.go
```diff
@@ -0,0 +1,215 @@
...
+			Tags:          []string{"edpm_podman"},
+		},
+		{
+			Name:          "Install edpm_sshd",
+			RoleName:      "edpm_sshd",
+			RoleTasksFrom: "install.yml",
+			Tags:          []string{"edpm_sshd"},
+		},
+		{
+			Name:          "Install chrony",
```
slagle, line 57: this role got renamed to edpm_chrony, so that needs to be updated everywhere.
https://github.com/openstack-k8s-operators/edpm-ansible/pull/46

slagle (commented):
Can you do a followup PR? I noticed these changes are still needed after I had already approved:

Need to update again for the chrony -> edpm_chrony role rename.

Other than that, I got to the configure os step, which fails b/c the ansible var tenant_ip is not set. That's a deployment specific setting, so we could add that to config/samples/dataplane_v1beta1_openstackdataplanenode_deployment.yaml under ansibleVars, set it to 192.168.24.100 or similar.

edpm_ovn_dbs also needs to be set.

Overall it's these 2 lines:

`
Wed Feb 22 03:20:48 (jslagle@teletran-1) [~/code/github/openstack-k8s-operators/dataplane-operator] (os_config)$ git diff config/samples/
diff --git a/config/samples/dataplane_v1beta1_openstackdataplanenode_deployment.yaml b/config/samples/dataplane_v1beta1_openstackdataplanenode_deployment.yaml
index 677d5d6..782e3a5 100644
--- a/config/samples/dataplane_v1beta1_openstackdataplanenode_deployment.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenode_deployment.yaml
@@ -19,5 +19,7 @@ spec:
       ctlplane_dns_nameservers:
       - 192.168.122.1
       dns_search_domains: []
+      tenant_ip: 192.168.24.100
+      edpm_ovn_dbs: 192.168.24.1
     deploy: true
     ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
`

pkg/deployment/os.go
fao89: I've addressed it here https://github.com/openstack-k8s-operators/dataplane-operator/pull/48

PR #44, opened 2023-02-21
Adding missing values to network role

fao89:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/edpm_ansible/playbooks/deploy-edpm-network-configure.yml

### DIFF
--- a/pkg/deployment/network.go
+++ b/pkg/deployment/network.go
@@ -30,9 +30,10 @@ import (
 func ConfigureNetwork(ctx context.Context, helper *helper.Helper, obj client.Object, sshKeySecret string, inventoryConfigMap string) error {
 
 	role := ansibleeev1alpha1.Role{
-		Name:     "edpm_network_config",
-		Hosts:    "all",
-		Strategy: "linear",
+		Name:        "edpm_network_config",
+		Hosts:       "all",
+		Strategy:    "linear",
+		GatherFacts: false,
 		Tasks: []ansibleeev1alpha1.Task{
 			{
 				Name: "import edpm_network_config",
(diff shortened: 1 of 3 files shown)


fultonj (commented):
+2

pkg/deployment/network.go
```diff
@@ -30,16 +30,19 @@ import (
 func ConfigureNetwork(ctx context.Context, helper *helper.Helper, obj client.Object, sshKeySecret string, inventoryConfigMap string) error {
 
 	role := ansibleeev1alpha1.Role{
-		Name:     "edpm_network_config",
-		Hosts:    "all",
-		Strategy: "linear",
+		Name:        "edpm_network_config",
+		Hosts:       "all",
+		Strategy:    "linear",
```
fao89, line 35: should this be `free`? 
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/edpm_ansible/playbooks/deploy-edpm-network-configure.yml#L19

slagle: No, I'd prefer to keep the strategy as linear. It makes things much easier to debug when looking at the ansible log. Unless we need to optimize, I'd say stick with linear.

pkg/deployment/network.go
```diff
@@ -30,16 +30,19 @@ import (
...
+		GatherFacts: false,
 		Tasks: []ansibleeev1alpha1.Task{
 			{
 				Name: "import edpm_network_config",
 				ImportRole: ansibleeev1alpha1.ImportRole{
 					Name:      "edpm_network_config",
 					TasksFrom: "main.yml",
 				},
+				Vars: "  edpm_network_config_with_ansible: true",
```
slagle, line 44: The var isn't really needed as it's already defaulted to true

slagle (changes requested):

PR #53, opened 2023-02-27
Allow attaching service pods to additional networks

stuggi:
For network isolation outgoing traffic, service pods need to be attached additional networks. To achive this, the
`k8s.v1.cni.cncf.io/networks` annotation needs to be set on the pods. This adds a networkAttachments parameter to the CRD where it can be specified to which additional networks the service pods should be attached to.
Its a list of pre-created net-attach-defs names. Per default the pods won't get attached. spec:

```
  ...
  networkAttachments:
  - internalapi
```

Depends-On: openstack-k8s-operators/lib-common#133
Depends-On: openstack-k8s-operators/openstack-ansibleee-operator#93


stuggi:
need update of go.mod when https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/93 merged

pkg/deployment/deployment.go
```diff
@@ -121,15 +129,28 @@ func Deploy(ctx context.Context, helper *helper.Helper, obj client.Object, sshKe
...
+	sshKeySecret string,
+	inventoryConfigMap string,
+	status *dataplanev1beta1.OpenStackDataPlaneNodeStatus,
+	readyCondition condition.Type,
+	readyMessage string,
+	readyWaitingMessage string,
+	deployFunc deployFuncDef,
+	deployName string,
+	deployLabel string,
+	networkAttachments []string,
```
stuggi, line 145: we might want to introduce a custom struct type to pass information instead of that many params

stuggi:
> need update of go.mod when [openstack-k8s-operators/openstack-ansibleee-operator#93](https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/93) merged

done

fao89 (commented):
+2

slagle:
@stuggi to be clear, this is for the pod running ansible-runner to connect over SSH to the edpm nodes correct? The PR body is not clear to me, given it talks about service pods and shows internalapi.

Is it the case that certain OCP worker nodes have a network configured on them (via the nmstate operator) that allows SSH access to edpm nodes? The set NetworkAttachmentDefinition will then make sure that the ansible-runner pods are created on the right OCP worker nodes such that ansible can SSH to the EDPM nodes.

stuggi:
> @stuggi to be clear, this is for the pod running ansible-runner to connect over SSH to the edpm nodes correct? The PR body is not clear to me, given it talks about service pods and shows internalapi.
> 
> Is it the case that certain OCP worker nodes have a network configured on them (via the nmstate operator) that allows SSH access to edpm nodes? The set NetworkAttachmentDefinition will then make sure that the ansible-runner pods are created on the right OCP worker nodes such that ansible can SSH to the EDPM nodes.

Your understanding is correct. I re-used the commit message from my other PRs to the ctlplane services. Worker nodes will be configured to be connected to the isolated networks using NNCP, via network-attachment-definitions pods then can be attached to those networks. This PR allows to reference those network-attachment-definitions where the ansible-runner pod needs to be attached to for connecting to the edpm nodes via ssh.

PR #54, opened 2023-02-27
AnsibleSSHPrivateKeySecret to follow kubernetes.io/ssh-auth type

### LINKED ISSUE #52: the ssh private key secret shoudl be of type 'kubernetes.io/ssh-auth'
currently we are using a generic secret for the ansible ssh key we should use --type 'kubernetes.io/ssh-auth'

this requires the private and public keys to have a specific name "ssh-privatekey" and "ssh-publickey"

https://kubernetes.io/docs/concepts/configuration/secret/#use-case-pod-with-ssh-keys

currently we use ssh_private_key

https://github.com/openstack-k8s-operators/dataplane-operator/blob/9291ba1ab72a812a33172a5296ea72d393e04697/api/v1beta1/openstackdataplanenode_types.go#L89-L94

which is allowed because we were using generic secret types but is not following standard convention.

we should adapt to 'kubernetes.io/ssh-auth' instead.

fao89:
https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets

closes #52


pkg/util/ansible_execution.go
```diff
@@ -71,7 +71,7 @@ func AnsibleExecution(ctx context.Context, helper *helper.Helper, obj client.Obj
 					SecretName: sshKeySecret,
 					Items: []corev1.KeyToPath{
 						{
-							Key:  "ssh_private_key",
+							Key:  "ssh-privatekey",
```
SeanMooney, line 74: currenlty im am adding both the ssh privatekey and the corresponding public key.

if we wanted to use the same secret with the baremetal operator we would probably want to require both filed are present in the secret.

i think that in general the operator is missing input validation for the secret.

we should probably move this to libcommon eventually 

https://github.com/openstack-k8s-operators/nova-operator/blob/master/controllers/common.go#L82-L147

https://github.com/openstack-k8s-operators/nova-operator/blob/1b8be195d2b37cc93ec33a63cb02e7ea507fbd83/controllers/novaapi_controller.go#L150-L167

but you probably need to add input validation that ensure the required keys are present and report this in the status when not.

SeanMooney (changes requested):
im not sure if we want to do this in this pr or as a separate issue but
the dataplane operator should be doing validation fo required fields on the seacret passed to the dataplane CR. see comment inline.

pkg/util/ansible_execution.go
SeanMooney: if we want to declar it out of scope fo this PR and do it in a different one that is fine but we should never get as far as invoking the AnsibleEE runner container before we have validated the input of the secret that we are using.

fao89: +1 to moving to lib-common

fultonj: I assume we'll merge this with the validation in this repo's common.go and then in a future PR add the key validation to lib-common and update this patch to use it.

controllers/openstackdataplanenode_controller.go
```diff
@@ -94,6 +96,20 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
+	_, result, err = ensureSecret(
```
fultonj, line 99: So we'll ensure the secret meets our constraints or throw an error here before the execution below.

api/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
```diff
@@ -106,6 +106,8 @@ spec:
                           type: string
                       type: object
                     type: array
+                required:
+                - ansibleSSHPrivateKeySecret
```
fultonj, line 110: Agreed, the SSH key is required to create data-plane{,role,node} objects. Else, we can't really manage them.

fultonj (commented):
I'm +2 on this based on my reading of it. I didn't test it in my environment though.

controllers/openstackdataplanenode_controller.go
fao89: yes!
it will ensure the secret contains `ssh-privatekey`

gibizer (commented):
looks good. I'm +1 on eventually moving ensureSecret to lib-common but I have no spare cycle at the moment to do it.

pkg/util/ansible_execution.go
fao89: I started this: https://github.com/openstack-k8s-operators/lib-common/pull/201

fao89: @SeanMooney the lib-common PR will need more discussion, could you please review this PR as is?
I'll do a follow up PR once the lib-common PR is merged

slagle (approved):

PR #55, opened 2023-02-27
Fix minor issues on EDPM diagram

fao89:


docs/diagrams/src/edpm.puml
```diff
@@ -28,20 +28,20 @@ folder AEEconfigmap #line.bold
...
+ODPRCR --> DOp #line:brown;line.bold;text:brown : consumed
+NCR --> NOp #line:brown;line.bold;text:brown : consumed
 DOp --> AEECR #line:purple;line.bold;text:purple  : creates
 DOp --> NECCR #line:purple;line.bold;text:purple  : creates
-AEECR --> AEEOp #line:brown;line.bold
-NECCR --> NOp #line:brown;line.bold
-AEEOp --> AJI #line:blue;line.bold
-AEEOp --> AJNC #line:blue;line.bold
+AEECR --> AEEOp #line:brown;line.bold;text:brown : consumed
+NECCR --> NOp #line:brown;line.bold;text:brown : consumed
```
SeanMooney, line 42: i woudl use manages in instead of consumes and i think the arrow is reversed.

so

NOp --> NECCR #line:brown;line.bold;text:brown : manages

and i think the same applies to all the lines you added consumed too

i think the relation ship should be reversed and we should use manages instead of consumed.

SeanMooney: actully reconsiles might also be a better name then manages

SeanMooney (commented):
this looks good to me

fultonj (approved):
+2 thanks

PR #58, opened 2023-03-01
Use VerifySecret from lib-common

fao89:


fultonj (commented):
+2

The dependent patch has also merged in lib-common:

  https://github.com/openstack-k8s-operators/lib-common/pull/201

slagle (approved):

PR #59, opened 2023-03-01
Make AnsibleSSHPrivateKeySecret optional and inherited

slagle:
This field should be inherited from the role if set. It also should be
optional within the NodeSection, otherwise it's required to be set on
every node.

The lib-common VerifySecret function will ensure that the secret is
actually a value and exists.

### DIFF
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -101,9 +101,15 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
+	instanceRole, err := r.GetInstanceRole(ctx, instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	ansibleSSHPrivateKeySecret := r.GetAnsibleSSHPrivateKeySecret(instance, instanceRole)
 	_, result, err = secret.VerifySecret(
 		ctx,
-		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Node.AnsibleSSHPrivateKeySecret},
+		types.NamespacedName{Namespace: instance.Namespace, Name: ansibleSSHPrivateKeySecret},
 		[]string{
 			"ssh-privatekey",
 		},
@@ -112,11 +118,6 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 	)
 	if err != nil {
 		return result, err
-	}
-
-	instanceRole, err := r.GetInstanceRole(ctx, instance)
-	if err != nil {
-		return ctrl.Result{}, err
 	}
 
 	// Always patch the instance status when exiting this function so we can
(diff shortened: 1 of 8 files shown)


controllers/openstackdataplanenode_controller.go
```diff
@@ -365,3 +366,11 @@ func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleVars(instance *dataplanev1b
 	}
 	return instanceRole.Spec.NodeTemplate.AnsibleVars
 }
+
+// GetAnsibleSSHPrivateKeySecret returns the secret name holding the private SSH key
+func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleSSHPrivateKeySecret(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) string {
+	if instance.Spec.Node.AnsibleSSHPrivateKeySecret != "" {
+		return instance.Spec.Node.AnsibleSSHPrivateKeySecret
+	}
+	return instanceRole.Spec.NodeTemplate.AnsibleSSHPrivateKeySecret
```
fao89, line 375: it should raise an error when both are empty

fultonj: > it should raise an error when both are empty

Don't we already get error raising behavior if both are empty?

The lib-common VerifySecret function will ensure that the secret is actually a value and exists. It's called immediately after this new GetAnsibleSSHPrivateKeySecret function. Thus, this function will return "" if both are empty and then VerifySecret will raise the error.

https://github.com/fao89/lib-common/blob/8d4d1a38b86d970608abfa205c94881d14c9ca86/modules/common/secret/secret.go#L391

For this reason I'm OK with it as is so +2, but please let me know if the above reasoning has a problem.

fultonj (commented):
+2

controllers/openstackdataplanenode_controller.go
fao89: I don't know how `VerifySecret` deals with empty string as `secretName`, maybe this validation should be there

fao89: @fultonj https://github.com/openstack-k8s-operators/lib-common/pull/203

fao89 (approved):
you just need to run:
```
make manifests bundle generate
```
to make CI happy

controllers/openstackdataplanenode_controller.go
fultonj: I assumed this code would encounter an `IsNotFound()` error if `secretName` was `""` but that's just my assumption. I didn't test it.
```
	err := reader.Get(ctx, secretName, secret)
	if err != nil {
		if k8s_errors.IsNotFound(err) {
                ...
```

https://github.com/fao89/lib-common/blob/8d4d1a38b86d970608abfa205c94881d14c9ca86/modules/common/secret/secret.go#L399-L407

Chat GPT says it will return an error but I'm not saying we have to trust it.

https://photos.app.goo.gl/ScXsKLcZQpRe2nNXA

fao89: oh! nice pair programming, I'll try that next time

fultonj: Confirmed by the link to the source you found :) 

https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Akubernetes+%22name+must+be+specified%22

slagle:
> you just need to run:
> 
> ```
> make manifests bundle generate
> ```
> 
> to make CI happy

Done.

PR #68, opened 2023-03-03
Stop overall deployment if one of the deployment steps fails

fao89:


controllers/openstackdataplanenode_controller.go
```diff
@@ -303,10 +303,12 @@ func (r *OpenStackDataPlaneNodeReconciler) GetInstanceRole(ctx context.Context,
...
-	err = r.Client.Get(ctx, client.ObjectKey{
-		Namespace: instance.Namespace,
-		Name:      instance.Spec.Role,
-	}, instanceRole)
+	if len(instance.Spec.Role) > 0 {
+		err = r.Client.Get(ctx, client.ObjectKey{
+			Namespace: instance.Namespace,
+			Name:      instance.Spec.Role,
+		}, instanceRole)
+	}
```
fultonj, line 311: I agree about only running Get if there was a role set. Regardless of if we want to make the role mandatory or optional, either way we should have this code (306-311).  If we go optional, then we could have something like this.

  https://github.com/openstack-k8s-operators/dataplane-operator/pull/69

pkg/deployment/deployment.go
```diff
@@ -364,13 +364,14 @@ func ConditionalDeploy(
...
 			log.Info(fmt.Sprintf("%s error", readyCondition))
-			err = fmt.Errorf("job: job.namespace %s job.name %s failed", ansibleEEJob.Namespace, ansibleEEJob.Name)
+			err = fmt.Errorf("failed: job.name %s job.namespace %s", ansibleEEJob.Name, ansibleEEJob.Namespace)
 			status.Conditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.ErrorReason,
 				condition.SeverityWarning,
 				readyErrorMessage,
 				err.Error()))
+			return ctrl.Result{}, err
```
fultonj, line 374: So if the job fails, then we'll return a failure. I agree.

fultonj (commented):
+2
Let's fail early instead of giving the user the impression we're going to do something.

Since this includes failing early if the role is undefined we should consider if we want the role to be optional or required. If required, then we should update the API too so that the node CR can't be created unless it has a role. Alternatively, we can allow the role to be optional as per the following. 

https://github.com/openstack-k8s-operators/dataplane-operator/pull/69

Either way, this patch (68) and the above (69) are compatible.

controllers/openstackdataplanenode_controller.go
slagle: +1

slagle (approved):

PR #69, opened 2023-03-03
Make node's role field optional

fultonj:
If an OpenStackDataPlaneNode CR is created before any OpenStackDataPlaneRole exists then do not fail with the error below.

This patch assumes we want to support creating nodes without roles. If we want to refuse to create nodes unless there is a role, then a validation should be added to the API so that the role field of the node is required.
```
{
  "controller": "openstackdataplanenode",
  "controllerGroup": "dataplane.openstack.org",
  "controllerKind": "OpenStackDataPlaneNode",
  "OpenStackDataPlaneNode": {
    "name": "network-edpm-compute-0",
    "namespace": "openstack"
  },
  "namespace": "openstack",
  "name": "network-edpm-compute-0",
  "reconcileID": "ada7594e-65b2-446e-b4bc-49d8989d5ad3",
  "error": "OpenStackDataPlaneRole.dataplane.openstack.org \"\" not found"
}
```

### DIFF
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -102,10 +102,8 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
-	instanceRole, err := r.GetInstanceRole(ctx, instance)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
+	// An instance can be created without a role so ignore error
+	instanceRole, _ := r.GetInstanceRole(ctx, instance)
 
 	ansibleSSHPrivateKeySecret := r.GetAnsibleSSHPrivateKeySecret(instance, instanceRole)
 	_, result, err = secret.VerifySecret(


fultonj:
Maybe this should be combined with https://github.com/openstack-k8s-operators/dataplane-operator/pull/68 

We also have to decide if we want to make the role mandatory or optional

fao89 (dismissed):

controllers/openstackdataplanenode_controller.go
```diff
@@ -102,10 +102,8 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
-	instanceRole, err := r.GetInstanceRole(ctx, instance)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
+	// An instance can be created without a role so ignore error
+	instanceRole, _ := r.GetInstanceRole(ctx, instance)
```
fao89, line 106: we need to handle the error for when the user provides a wrong name

fultonj: Yes, I agree. So a wrong name might be a role that doesn't exist.

I think I'll rebase this patch on https://github.com/openstack-k8s-operators/dataplane-operator/pull/68 and then handle that case by failing. If the role name isn't set, then by PR 68 the Get won't be run so we need to not fail in that case.

fao89: if we decide the role is required, I just need to revert that commit

slagle: I don't think role should be required. Let's make it optional.

fultonj: Agree, it should be optional. Now that we have https://github.com/openstack-k8s-operators/dataplane-operator/pull/68 it effectively is.

fultonj:
The error below from the role not existing no longer happens
now that PR 68* merged. The Get function failures can still
cause a failure but not when the role is not provided because
Get is only called if a role is specified. Thus, PR 68*
effectively made the role optional as was the original intent.

Also, we don't want to remove the error handling this patch
was adding because if the role name is provided but we can't
get it, then we should still through an error. Thus, I'm
closing this PR.

* https://github.com/openstack-k8s-operators/dataplane-operator/pull/68

```
{
  "controller": "openstackdataplanenode",
  "controllerGroup": "dataplane.openstack.org",
  "controllerKind": "OpenStackDataPlaneNode",
  "OpenStackDataPlaneNode": {
    "name": "network-edpm-compute-0",
    "namespace": "openstack"
  },
  "namespace": "openstack",
  "name": "network-edpm-compute-0",
  "reconcileID": "ada7594e-65b2-446e-b4bc-49d8989d5ad3",
  "error": "OpenStackDataPlaneRole.dataplane.openstack.org \"\" not found"
}
```

PR #70, opened 2023-03-03
Fix operator-sdk markers

fao89:


slagle (dismissed):

fultonj (changes requested):
IMO this should only change the deployStrategy for the sdk-markers.

I don't think we should change the default ansible port from 0 to 22. This patch will result in the ansible port always being specified in the inventory. However, that is not necessary since 22 is the default that ansible will use if we don't specify anything. Thus, all of our config will have an extra line which isn't necessary. As the code is currently written (without the port change of this patch) the port is only set in the inventory if someone sets it and we should expect them to do this only if they want a non-standard port.

If we're still going to change the ansible port, then this code needs to be updated to handle the new default.

https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanenode_controller.go#L331

slagle:
> IMO this should only change the deployStrategy for the sdk-markers.
> 
> I don't think we should change the default ansible port from 0 to 22. This patch will result in the ansible port always being specified in the inventory. However, that is not necessary since 22 is the default that ansible will use if we don't specify anything. Thus, all of our config will have an extra line which isn't necessary. As the code is currently written (without the port change of this patch) the port is only set in the inventory if someone sets it and we should expect them to do this only if they want a non-standard port.
> 
> If we're still going to change the ansible port, then this code needs to be updated to handle the new default.
> 
> https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanenode_controller.go#L331

Good point John, I agree

slagle (commented):
+2

fultonj (approved):

PR #73, opened 2023-03-03
Group nodes on roles with only the OpenStackDataPlaneNode.Spec.Role field

slagle:
Removes the list of nodes in a role that was set with the
OpenStackDataPlaneRole.Spec.DataPlaneNodes field. Instead, the node and
role grouping association will only be captured in the
OpenStackDataPlaneNode.Spec.Role field.

This simplifies the grouping logic and ensures that the single source of
truth is only kept in one place. It removes the need to keep the list of
nodes in sync on the role. Nodes are also no longer reconciled by a Role
reconcile reequest. Instead, nodes can be created in bulk using the
OpenStackDataPlaneNodeList type.

To simplify querying or filtering for nodes in a role, the
"openstackdataplanerole" label is added to nodes that are in a role. The
label's value is the role name.

This patch also updates the README.md with this change, and adds
additional docs to describe the behavior.


README.md
```diff
@@ -4,45 +4,65 @@ The dataplane-operator automates the deployment of an OpenStack dataplane.
...
-operates on a single node CR, and manages that node in isolation, performing
-all the necessary baremetal provisioning and software deployment/configuration
-on that node.
+`openstackdataplane=<dataplane-name>`
 
-The OpenStackDataPlane CRD and controller orchestrates deploying roles, and
-provides synchronization points between role deployment.
+The dataplane does not maintain the list of roles in that dataplane, to ensure
+that the source source of dataplane and role grouping is stored only in the
+`DataPlane` field on the role.
```
fultonj, line 34: Perhaps re-write 32,33,34 as:

The dataplane does not maintain a list of roles. The source of dataplane and role grouping is stored in the `DataPlane` field of the role.

README.md
```diff
@@ -4,45 +4,65 @@ The dataplane-operator automates the deployment of an OpenStack dataplane.
...
 
-## Description
-// TODO(user): An in-depth paragraph about your project and overview of use
+The OpenStackDataPlaneRole CRD provides for a logical grouping of nodes of a
+similar type within a role. Similarities within a role are defined by the
+user, and could be of a small scope (ansible port), or a large scope (same
+network config, nova config, provisioning config, etc). The similar properties
+that all nodes in a role share set in the `NodeTemplate` field on the role's
```
fultonj, line 40: Perhaps re-write 39,40,41 as 

The properties that all nodes in a role share is set in the `NodeTemplate` field of the role's `Spec`.

controllers/openstackdataplanenode_controller.go
```diff
@@ -102,6 +102,17 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
+	if len(instance.Spec.Role) > 0 {
+		if instance.ObjectMeta.Labels == nil {
+			instance.ObjectMeta.Labels = make(map[string]string)
+		}
+		r.Log.Info(fmt.Sprintf("Adding label %s=%s", "openstackdataplanerole", instance.Spec.Role))
+		instance.ObjectMeta.Labels["openstackdataplanerole"] = instance.Spec.Role
+	} else if instance.ObjectMeta.Labels != nil {
```
fultonj, line 111: OK, I see we're removing a possible role just in case the instance didn't specify it.

fultonj (commented):
This is quite a big change but it does simplify and remove redundant data structures. Having the node hold all the (growing) node fields (instead of the role and node) and then having only the node point to the role is probably a good move. I'm going to test the patch in my environment with the updated examples and then complete my review.

fultonj (commented):
I've played with the samples and I'm +2 on this change. 

  https://paste.opendev.org/show/bMRilG3G8yOcdcxHNrbQ/

Maybe just fix up a few typos in the proposed docs patch and get input from @fao89

config/samples/dataplane_v1beta1_openstackdataplanerole.yaml
```diff
@@ -3,16 +3,6 @@ kind: OpenStackDataPlaneRole
...
-  dataPlaneNodes:
-    - name: openstackdataplanenode-sample
-      hostName: openstackdataplanenode-sample.localdomain
-      ansibleHost: 192.168.122.18
-      node:
-        networks:
-          - network: ctlplane
-            fixedIP: 192.168.122.18
-    - name: openstackdataplanenode-sample-from
-      nodeFrom: openstackdataplanenode-sample-from
```
fao89, line 15: I like the idea of declaring multiple node in one shot like this, but I understand the complexity of dealing with multiple reconcile loops

fao89 (commented):
the code LGTM, but I'm wondering how many nodes a real setup normally have and what would be the user experience of having a CR for each of them. How was it done before? Was it done through yaml?

README.md
slagle: Done.

README.md
slagle: Done.

fao89 (approved):

PR #74, opened 2023-03-07
Initial IPAM support

rabi:


stuggi:
just a general note, in general its not important where the CRDs are, but since its also used by ctlplane components, I think we want the IPAM to be located in the infra-operator with other network related controllers like DNS.

rabi:
>  but since its also used by ctlplane components

Which ctlplane components do you think would use our baked-in ipam? Do we plan to use it 'only' for the VIPs in ctlplane? Would not all other ctlplane components use whereabouts cni plugin? I thought we would keep in dataplane as it's specific to dataplane requirements.

controllers/dataplaneipset_controller.go
```diff
@@ -0,0 +1,247 @@
...
+			ipReservation, updatedReservations, err := ipam.AssignIP(aid, net.ParseIP(network.FixedIP))
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			ipReservation.Gateway = subnet.Gateway
+			ipReservation.Routes = subnet.Routes
+
+			r.Log.Info("Got Reservations")
+			aid.Reservelist = updatedReservations
+			osnet.Status.Reservations = append(osnet.Status.Reservations, ipReservation)
```
stuggi, line 164: a different controller should not update other controllers CR status

api/v1beta1/openstacknet_types.go
```diff
@@ -0,0 +1,158 @@
...
+	Gateway string `json:"gateway,omitempty"`
+	// +kubebuilder:validation:Optional
+	Routes []Route `json:"routes,omitempty"`
+}
+
+// OpenStackNetStatus defines the observed state of OpenStackNet
+type OpenStackNetStatus struct {
+	// Already reserved IPs
+	// +kubebuilder:validation:Optional
+	Reservations []IPReservation `json:"reservations,omitempty"`
```
stuggi, line 113: how would we handle backup/restore with the Reservations stored in status? you can not just save the yaml of an osnet with its status and apply it back to the env. it would required special restore to also set the status to a state. thats why when we used the osnet in OSPDO we stored the reservations in the spec and a reservation was successfully stored if it was written to the spec, but also when the osnet controller successfully has updated its status with a different reservation format.

stuggi:
> > but since its also used by ctlplane components
> 
> Which ctlplane components do you think would use our baked-in ipam? Do we plan to use it 'only' for the VIPs in ctlplane? Would not all other ctlplane components use whereabouts cni plugin? I thought we would keep in dataplane as it's specific to dataplane requirements.

Some components need static IPs, like ovs to make sure that they don't change on a pod restart. with whereabouts you can not register static IPs. the only way to use it with whereabouts might be to use a network-attachment-definition with a single IP, but then we'd still need to track (IPAM) which IP is used for which instance/NAD.

pkg/ipam/ipam.go
```diff
@@ -0,0 +1,238 @@
...
+		if iPToBigInt(ip).Cmp(i) == 0 {
+			for _, rs := range assignIPDetails.Reservelist {
+				if rs.IP == ip.String() {
+					return false, dataplanev1beta1.IPReservation{}, fmt.Errorf("Already reserved %s", ip.String())
+				}
+			}
+			return true, dataplanev1beta1.IPReservation{IP: ip.String(), VIP: false}, nil
+		}
+	}
+	return false, dataplanev1beta1.IPReservation{}, fmt.Errorf("Does not exist in range %s", ip.String())
```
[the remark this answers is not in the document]
bshephar: Eh, we already have net.IPNet sorry. So we should have the IP and Mask in the IPNet field of the existing struct that we can use to check if the IP address falls within the subnet. My bad on that bit, I missed that even when copying and adding the Range.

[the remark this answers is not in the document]
rabi: I've not looked at optimizing the code yet (just got something working). We have allocation range which we need to check, only using Mask won't be enough.  We can easily avoid the nested for loop with a flag.  I had a discussion with @stuggi and @abays yesterday as we can't probably update the status of osnet for reservations as there would be an issue with standard backup and restore. Though I'm not convinced to use a spec field for reservations (which can be updated by any random api call), unless there is a better way we would  probably use that.

```
	inRange := false
	for i := iPToBigInt(firstip); iPToBigInt(lastip).Cmp(i) == 1 || iPToBigInt(lastip).Cmp(i) == 0; i.Add(i, big.NewInt(1)) {
		if iPToBigInt(ip).Cmp(i) == 0 {
			inRange = true
			break
		}
	}
	if inRange {
		for _, rs := range assignIPDetails.Reservelist {
			if rs.IP == ip.String() {
				return false, dataplanev1beta1.IPReservation{}, fmt.Errorf("Already reserved %s", ip.String())
			}
		}
		return true, dataplanev1beta1.IPReservation{IP: ip.String(), VIP: false}, nil
	}
	return false, dataplanev1beta1.IPReservation{}, fmt.Errorf("Does not exist in range %s", ip.String())
```

rabi:
> Some components need static IPs, like ovs to make sure

Should we leave it in dataplane-operator for the time being as usage of it in controlplane is an exception rather than norm? If we've concerns with that I can propose a new PR to infra-operator instead.

controllers/dataplaneipset_controller.go
rabi: It works though and I did not find any references which suggests not to do it. Though I agree on the issues with custom backup and restore requirement we discussed yesterday which can be avoided.

stuggi:
> > Some components need static IPs, like ovs to make sure
> 
> Should we leave it in dataplane-operator for the time being as usage of it in controlplane is an exception rather than norm? If we've concerns with that I can propose a new PR to infra-operator instead.

The plan was to have a network group in the infra-operator and have all networking related CRDs which we have to manage in one place. I think it's cleaner when ctlplane and dataplane both use it and the ctlplane components don't rely on the dataplane. 

@dprince @abays @olliewalsh your thoughts on this?

rabi:
> The plan was to have a network group in the infra-operator and have all networking related CRDs which we have to manage in one place.

OK, submitted to infra-operator https://github.com/openstack-k8s-operators/infra-operator/pull/13 and changed to use spec field for reservations, rather than status.

PR #79, opened 2023-03-09
Expose ExtraMounts in OpenStackDataPlaneNode

fultonj:
OpenStackDataPlaneNode (DPN) is already using ExtraMounts when it calls OpenStackAnsibleEE (AEE). This patch allows ExtraMounts to be set in a DPN CR so that its Ansible pod can access those files in order to configure a node.

A subsequent patch will use ExtraMounts so the AEE pod can access the same `extraVolType: Ceph` used by control plane pods and apply the same configuration to DPNs for Ceph client configuration. Additional applications are possible.

Related: [OSP-22956](https://issues.redhat.com//browse/OSP-22956) [OSP-19983](https://issues.redhat.com//browse/OSP-19983)

### DIFF
--- a/pkg/util/ansible_execution.go
+++ b/pkg/util/ansible_execution.go
@@ -49,6 +49,7 @@ func AnsibleExecution(
 	networkAttachments []string,
 	openStackAnsibleEERunnerImage string,
 	ansibleTags string,
+	extraMounts []storage.VolMounts,
 ) error {
 
 	var err error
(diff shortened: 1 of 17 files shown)


pkg/util/ansible_execution.go
```diff
@@ -127,7 +128,7 @@ func AnsibleExecution(
 		ansibleEEMounts.Mounts = append(ansibleEEMounts.Mounts, sshKeyMount)
 		ansibleEEMounts.Mounts = append(ansibleEEMounts.Mounts, inventoryMount)
 
-		ansibleEE.Spec.ExtraMounts = []storage.VolMounts{ansibleEEMounts}
+		ansibleEE.Spec.ExtraMounts = append(extraMounts, []storage.VolMounts{ansibleEEMounts}...)
```
fultonj, line 131: Merge `extraMounts` into the existing `ansibleEE.Spec.ExtraMounts` like the sshKeyVolme etc.

fmount: Looks good

fmount (approved):
The patch adds the ExtraMounts struct to the DPO API and we're able to call AEE passing the ExtraMounts defined in the top-level crd.
Looks good to me, thanks John!

PR #81, opened 2023-03-10
PoC: Adding deploy hooks

fao89:
PoC: pre and post hooks for executing AEE.play from configmaps


fao89:
closing as the third party integration needs more design

PR #85, opened 2023-03-13
Configure OpenStackDataPlaneNode as Ceph clients

fultonj:
If an OpenStackDataPlaneNode has any extraMounts items where extraVolType is Ceph, then call the Ansible Role edpm_ceph_client_files and pass it the mounted files. This results in nodes having Ceph configuration files (e.g. ceph.conf or a cephx key) so that the OpenStack services on those nodes can be clients of a Ceph cluster.

Closes: [OSP-22956](https://issues.redhat.com//browse/OSP-22956)

### DIFF
--- a/pkg/deployment/ceph_client.go
+++ b/pkg/deployment/ceph_client.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deployment
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	dataplaneutil "github.com/openstack-k8s-operators/dataplane-operator/pkg/util"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/storage"
+	ansibleeev1alpha1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
+)
+
+// ConfigureCephClient ensures the Ceph client configuration files are on data plane nodes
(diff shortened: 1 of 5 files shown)


pkg/deployment/ceph_client.go
```diff
@@ -0,0 +1,68 @@
...
+	haveCephSecret := false
+	for _, extraMount := range extraMounts {
+		if extraMount.ExtraVolType == "Ceph" {
+			haveCephSecret = true
+			break
+		}
+	}
+	if !haveCephSecret {
+		helper.GetLogger().Info("Skipping execution of Ansible for ConfigureCephClient because extraMounts does not have an extraVolType of Ceph.")
+		return nil
```
fultonj, line 42: I want this `return nil` to result in the ansibleEE for `dataplane-deployment-configure-ceph-client*` not running. However, when I pass a `DataPlaneNode` CR without an `extraMount` the flow gets stuck re-queueing. 

```
1.6787992745080173e+09	INFO	ConfigureCephClientReady not yet ready, requeueing	{"controller": 
"openstackdataplanenode", "controllerGroup": "dataplane.openstack.org", "controllerKind": 
"OpenStackDataPlaneNode", "OpenStackDataPlaneNode": {"name":"edpm-compute-
0","namespace":"openstack"}, "namespace": "openstack", "name": "edpm-compute-0", "reconcileID": "7ab3d851-
a3a1-4b23-bfc4-44d09509479e"}
```
When I pass a `DataPlaneNode` CR with an `extraMount` the patch works as I had hoped and moves on to `InstallOpenStack`. 

How can I have two flows based on what's in the CR?

```
if extraMounts:
  RunOS --> ConfigureCephClient --> InstallOpenStack
else:
  RunOS --> InstallOpenStack
```

fmount: you can skip in `deployment.go` the `conditionalDeploy` (where deployName =  ConfigureCephClient) if extraMounts is empty.

pkg/deployment/deployment.go
```diff
@@ -219,6 +219,37 @@ func Deploy(
...
+		deployName,
+		deployLabel,
+		networkAttachments,
+		openStackAnsibleEERunnerImage,
+		ansibleTags,
+		extraMounts,
+	)
+	if err != nil || result.RequeueAfter > 0 {
+		return result, err
+	}
```
fao89, line 251: it is easier to put this under an if,
as the flow is built here

fultonj: done.

fultonj:
This works with extraMounts for 1 or more Ceph clusters.
The ceph client ansible pod is not created if `extraVolType: Ceph` is not in `extraMounts`.

  https://github.com/fultonj/zed/blob/main/edpm/ceph.md#results

fao89 (commented):
+2

fmount (approved):

PR #87, opened 2023-03-14
Remove omitempty from fields with defaults

slagle:
make operator-lint now fails if a field has both a default value and is
marked omitempty. Remove omitempty in these cases.

### DIFF
--- a/api/v1beta1/openstackdataplanenode_types.go
+++ b/api/v1beta1/openstackdataplanenode_types.go
@@ -102,7 +102,7 @@ type NodeSection struct {
 	// ExtraMounts containing files which can be mounted into an Ansible Execution Pod
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default={}
-	ExtraMounts []storage.VolMounts `json:"extraMounts,omitempty"`
+	ExtraMounts []storage.VolMounts `json:"extraMounts"`
 }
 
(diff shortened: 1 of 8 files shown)


rebtoor (approved):

api/v1beta1/openstackdataplanenode_types.go
```diff
@@ -125,7 +125,7 @@ type NetworkConfigSection struct {
 	// +kubebuilder:default=templates/net_config_bridge.j2
 	// Template - ansible j2 nic config template to use when applying node
 	// network configuration
-	Template string `json:"template,omitempty" yaml:"template,omitempty"`
+	Template string `json:"template" yaml:"template"`
```
fao89, line 128: I think we should remove the default, otherwise we may have nodes/roles set with different templates

slagle: I'm not sure how, but ok. We're not actually using this field for anything yet. We might can remove it in a separate patch.

fao89: we use when building the inventory: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanenode_controller.go#L277

fao89 (approved):

PR #91, opened 2023-03-14
Remove description from DataPlane CRD

fmount:
This patch rebuilds the DataPlane CRDs removing the description to avoid a potential issue when the bundle is deployed via OLM by the meta operator and the following is seen:

'''
openstack/install-7ws9s" failed: failed to update installplan bundle lookups: etcdserver: request is too large
'''


rebtoor (commented):
+2 for me

PR #92, opened 2023-03-15
Move common data structures to common.go

fao89:
so we can fix the role docs


fultonj (commented):
Makes sense to me, though I didn't test it. +2

slagle (approved):

PR #93, opened 2023-03-15
Add DeployNovaExternalCompute to role deploy

slagle:
As the last step in the role deploy, DeployNovaExternalCompute is called
to deploy the nova services. This commit refactors the function since
the returns need to be handled differently whether it's a node deploy or
a role deploy.

During a node deploy, only the single created (or patched)
NovaExternalCompute needs to be checked for errors or the
ReadyCondition. However, during a role deploy, each NovaExternalCompute
for each corresponding node needs to be checked. If any of the
NovaExternalCompute CR's is in error, or not ready, then the entire role
is set as in error or not ready.

The call to DeployNovaExternalCompute is refactored out of the Deploy
func and called directly from the node and role controller instead so
that the return handling can be different.

This change also switches to use the standard ReadyCondition from
lib-common instead of the custom types DataPlaneNodeReadyCondition and
DataPlaneRoleReadyCondition. The custom types are not needed as a custom
message can already be set on the standard ReadyCondition. The standard
ReadyCondition is also used to reflect the overall status in the CLI and
UI.

### DIFF
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -140,7 +140,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		if instance.IsReady() {
 			instance.Status.Conditions.MarkTrue(condition.ReadyCondition, dataplanev1beta1.DataPlaneNodeReadyMessage)
 		}
-		c := instance.Status.Conditions.Mirror(dataplanev1beta1.DataPlaneNodeReadyCondition)
+		c := instance.Status.Conditions.Mirror(condition.ReadyCondition)
 		if c.Reason == condition.ErrorReason {
 			instance.Status.Conditions.MarkFalse(
 				condition.ReadyCondition,
@@ -161,7 +161,6 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		instance.Status.Conditions = condition.Conditions{}
 
 		cl := condition.CreateList(
-			condition.UnknownCondition(dataplanev1beta1.DataPlaneNodeReadyCondition, condition.InitReason, condition.InitReason),
 			condition.UnknownCondition(dataplanev1beta1.ConfigureNetworkReadyCondition, condition.InitReason, condition.InitReason),
 			condition.UnknownCondition(dataplanev1beta1.ValidateNetworkReadyCondition, condition.InitReason, condition.InitReason),
 			condition.UnknownCondition(dataplanev1beta1.InstallOSReadyCondition, condition.InitReason, condition.InitReason),
@@ -224,7 +223,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		if err != nil {
 			util.LogErrorForObject(helper, err, fmt.Sprintf("Unable to deploy %s", instance.Name), instance)
 			instance.Status.Conditions.Set(condition.FalseCondition(
-				dataplanev1beta1.DataPlaneNodeReadyCondition,
+				condition.ReadyCondition,
 				condition.ErrorReason,
 				condition.SeverityError,
 				dataplanev1beta1.DataPlaneNodeErrorMessage,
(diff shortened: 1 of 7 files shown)


controllers/openstackdataplanenode_controller.go
```diff
@@ -161,7 +161,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		instance.Status.Conditions = condition.Conditions{}
 
 		cl := condition.CreateList(
-			condition.UnknownCondition(dataplanev1beta1.DataPlaneNodeReadyCondition, condition.InitReason, condition.InitReason),
+			condition.UnknownCondition(condition.ReadyCondition, condition.InitReason, condition.InitReason),
```
fao89, line 164: you can remove this line, `Conditions.Init(&cl)` already adds it: https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/condition/funcs.go#L36

controllers/openstackdataplanerole_controller.go
```diff
@@ -198,14 +200,15 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 
 	if instance.Spec.DeployStrategy.Deploy {
 
-		r.Log.Info("Set DataPlaneRoleReadyCondition false")
-		instance.Status.Conditions.Set(condition.FalseCondition(dataplanev1beta1.DataPlaneRoleReadyCondition, condition.InitReason, condition.SeverityInfo, dataplanev1beta1.DataPlaneRoleReadyWaitingMessage))
+		r.Log.Info("Starting DataPlaneRole deploy")
+		r.Log.Info("Set ReadyCondition false")
+		instance.Status.Conditions.Set(condition.FalseCondition(condition.ReadyCondition, condition.InitReason, condition.SeverityInfo, dataplanev1beta1.DataPlaneRoleReadyWaitingMessage))
```
fao89, line 205:
```suggestion
		instance.Status.Conditions.Set(condition.FalseCondition(condition.ReadyCondition, condition.RequestedReason, condition.SeverityInfo, dataplanev1beta1.DataPlaneRoleReadyWaitingMessage))
```
Init is for Unknown: https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/condition/conditions.go#L74

controllers/openstackdataplanerole_controller.go
```diff
@@ -216,21 +219,72 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+				ansibleSSHPrivateKeySecret,
+				nodeConfigMapName,
+				&instance.Status,
+				instance.Spec.NetworkAttachments,
+				instance.Spec.OpenStackAnsibleEERunnerImage)
+			if err != nil {
+				novaErrors = append(novaErrors, err)
+			}
+			novaReadyCondition := novaExternalCompute.Status.Conditions.Get(condition.ReadyCondition)
+			if novaReadyCondition != nil && novaReadyCondition.Status == corev1.ConditionTrue {
```
fao89, line 241:
```suggestion
			if novaExternalCompute.IsReady() {
```

fao89 (approved):
just a few suggestions

controllers/openstackdataplanerole_controller.go
```diff
@@ -216,21 +219,72 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+
+			}
+		}
+
+		// When any errors are found, wrap all into a single error, and return
+		// it
+		errStr := "DeployNovaExternalCompute error:"
+		if len(novaErrors) > 0 {
+			for _, err := range novaErrors {
+				errStr = fmt.Sprintf("%s: %s", errStr, err.Error())
```
fao89, line 253: I believe the final error message would be something like this:
`DeployNovaExternalCompute error: stacktrace1: stacktrace2:...:stacktraceN`

controllers/openstackdataplanenode_controller.go
slagle: Done.

controllers/openstackdataplanerole_controller.go
slagle: Done.

controllers/openstackdataplanerole_controller.go
slagle: I think that's ok. Not sure how else to represent it.

fultonj (commented):
/lgtm

+2

controllers/openstackdataplanenode_controller.go
```diff
@@ -234,17 +233,60 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
+		} else {
+			r.Log.Info("NovaExternalCompute ReadyCondition not yet set")
+		}
+
+		mirroredCondition := novaExternalCompute.Status.Conditions.Mirror(dataplanev1beta1.NovaComputeReadyCondition)
+		if mirroredCondition != nil {
+			instance.Status.Conditions.Set(mirroredCondition)
+		}
+
+		if condition.IsError(instance.Status.Conditions.Get(dataplanev1beta1.NovaComputeReadyCondition)) {
```
bogdando, line 263: @gibi Are these checks expect the result of the set mirroredCondition to be in place yet?
Just to clarify, there should be no such expectations due to eventual consistency.
If these check work as expected with not yet updated older status conditions, then it is totally fine though.

bogdando: please disregard, I believe I got it now...

controllers/openstackdataplanerole_controller.go
```diff
@@ -216,21 +218,72 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+		instance.Status.Deployed = true
+		r.Log.Info("Set ReadyCondition true")
+		instance.Status.Conditions.Set(condition.TrueCondition(condition.ReadyCondition, dataplanev1beta1.DataPlaneRoleReadyMessage))
+
+		// Explicitly set instance.Spec.Deploy = false
+		// We don't want another deploy triggered by any reconcile request, it should
+		// only be triggered when the user (or another controller) specifically
+		// sets it to true.
+		r.Log.Info("Set DeployStrategy.Deploy to false")
+		instance.Spec.DeployStrategy.Deploy = false
```
bogdando, line 276: resetting this back to False, after a user requested a deploy could be confusing

PR #95, opened 2023-03-16
Validate NetworkAttachments and OpenStackAnsibleEERunnerImage

fao89:


fultonj (commented):
In general I'm +2 on this. 

One idea for refactoring to consider.

These lines:
  api/v1beta1/openstackdataplanerole_types.go lines 122-128
  api/v1beta1/openstackdataplanenode_types.go 99-105

Are the same excpet the items which are compared differ:
  instance.Spec.OpenStackAnsibleEERunnerImage, node.Spec.OpenStackAnsibleEERunnerImage
  imageNode.Spec.OpenStackAnsibleEERunnerImage, instance.Spec.OpenStackAnsibleEERunnerImage

Does it make sense to have a general function which both can call and pass different arguments?
As validations grow beyond just OpenStackAnsibleEERunnerImage and NetworkAttachments this pattern might be worth it.

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -82,3 +85,41 @@ func init() {
...
+
+		for _, node := range nodes {
+			if len(nodeMap["openStackAnsibleEERunnerImage"].Name) > 0 && len(node.Spec.OpenStackAnsibleEERunnerImage) > 0 && node.Spec.OpenStackAnsibleEERunnerImage != nodeMap["openStackAnsibleEERunnerImage"].Spec.OpenStackAnsibleEERunnerImage {
+				errorMsgs = append(errorMsgs, fmt.Sprintf("openStackAnsibleEERunnerImage mismatch between node: %s and node: %s", node.Name, nodeMap["openStackAnsibleEERunnerImage"].Name))
+			}
+
+			if len(nodeMap["networkAttachments"].Name) > 0 && len(node.Spec.NetworkAttachments) > 0 && !reflect.DeepEqual(node.Spec.NetworkAttachments, nodeMap["networkAttachments"].Spec.NetworkAttachments) {
+				errorMsgs = append(errorMsgs, fmt.Sprintf("networkAttachments mismatch between node: %s and node: %s", node.Name, nodeMap["networkAttachments"].Name))
+			}
+			// get the first non-empty
```
fultonj, line 111: I'm assuming the first time through the loop that the`nodeMap` is empty and so the first two conditions on lines 104 and 108 won't be true until after the first iteration. Thus, you're building the map in order to check that each item is the same as the previous item.

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -82,3 +85,41 @@ func init() {
...
 	return instance.Status.Conditions.IsTrue(condition.ReadyCondition)
 }
+
+// Validate - validates the shared data between role and nodes
+func (instance OpenStackDataPlaneRole) Validate(nodes []OpenStackDataPlaneNode) error {
+	var errorMsgs []string
+
+	if len(instance.Spec.OpenStackAnsibleEERunnerImage) > 0 || len(instance.Spec.NetworkAttachments) > 0 {
+		for _, node := range nodes {
+			errorMsgs = AssertNodeAndRoleUniqueness(node, instance, errorMsgs)
```
fultonj, line 95: It's good you can re-use this function here.

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -82,3 +85,41 @@ func init() {
...
+		}
+	}
+
+	// Compare nodes when role fields are empty
+	if len(instance.Spec.OpenStackAnsibleEERunnerImage) == 0 || len(instance.Spec.NetworkAttachments) == 0 {
+		nodeMap := make(map[string]OpenStackDataPlaneNode)
+
+		for _, node := range nodes {
+			if len(nodeMap["openStackAnsibleEERunnerImage"].Name) > 0 && len(node.Spec.OpenStackAnsibleEERunnerImage) > 0 && node.Spec.OpenStackAnsibleEERunnerImage != nodeMap["openStackAnsibleEERunnerImage"].Spec.OpenStackAnsibleEERunnerImage {
```
fultonj, line 104: There's still some duplicated logic on lines 104 and 108. We could have two functions in common.go 
`AssertNodeAndRoleUniqueness` and `AssertNodeAndNodeUniqueness` where the later is passed two nodes  but compares the same attributes (`RunnerImage` or `NetworkAttachments`).

api/v1beta1/common.go
```diff
@@ -102,3 +105,15 @@ type NetworksSection struct {
 	// FixedIP - Specific IP address to use for this network
 	FixedIP string `json:"fixedIP,omitempty" yaml:"fixedIP,omitempty"`
 }
+
+// AssertNodeAndRoleUniqueness - assert uniqueness between node and role fields
```
fultonj, line 109: `AssertNodeandRoleUniqueness` and `AssertNodeAndNodeUniqueness` could both work on taking the different types as input (role or node) and extracting the items to compare (`RunnerImage` or `NetworkAttachments`) but then both could pass to the same third function (e.g. `AssertUniqueness`) which ensures that the `RunnerImage` or `NetworkAttachments` do not differ.

If the goal is to not have duplicate code that's one way I see to do it.

fultonj (commented):
I'm glad this fixes the problem so we could merge it (+2).

I have one more idea to remove a little duplicate logic you could consider.

api/v1beta1/common.go
fao89: I'll try

api/v1beta1/common.go
```diff
@@ -102,3 +104,25 @@ type NetworksSection struct {
...
+
+	var errorMsgs []string
+	for _, field := range UniqueSpecFields {
+		value := vSpec.FieldByName(field)
+		otherValue := vOtherSpec.FieldByName(field)
+		if value.IsZero() || otherValue.IsZero() {
+			continue
+		}
+		if !reflect.DeepEqual(value.Interface(), otherValue.Interface()) {
+			errorMsgs = append(errorMsgs, field+" mismatch between "+suffix)
```
fultonj, line 124: Nice!

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -82,3 +85,46 @@ func init() {
...
+	for _, field := range UniqueSpecFields {
+		if reflect.ValueOf(instance.Spec).FieldByName(field).IsZero() {
+			containsEmptyField = true
+			break
+		}
+	}
+
+	if !containsEmptyField {
+		for _, node := range nodes {
+			suffix := fmt.Sprintf("node: %s and role: %s", node.Name, instance.Name)
```
fultonj, line 102: nice

api/v1beta1/common.go
```diff
@@ -102,3 +104,25 @@ type NetworksSection struct {
 	// FixedIP - Specific IP address to use for this network
 	FixedIP string `json:"fixedIP,omitempty" yaml:"fixedIP,omitempty"`
 }
+
+// UniqueSpecFields - the array of fields that must be unique between role and nodes
+var UniqueSpecFields = []string{"OpenStackAnsibleEERunnerImage", "NetworkAttachments"}
+
+// AssertUniquenessBetween - compare specs for uniqueness
+func AssertUniquenessBetween(spec interface{}, otherSpec interface{}, suffix string) []string {
```
fultonj, line 112: This refactoring makes it cleaner IMO. Thanks!

fultonj (commented):
+2

slagle (approved):

SeanMooney (commented):
i somewhat understand why you want to use reflection but the lack of test coverage in this patch is concerning.

i really hope we can get to the point in Q2 where we will not merge patches without test coverage. with the context of the upcoming demo i know why we are moving fast but we need to slow down and start writing test with each change especially when using tools like reflection or generics which are less obviously correct by simple inspection.

api/v1beta1/common.go
fultonj: @SeanMooney I agree that we want more test coverage

Can you give me an example of how to better implement `AssertUniquenessBetween`? Something which uses 
https://go.dev/doc/tutorial/generics ?

fao89: I've filed this issue: https://github.com/openstack-k8s-operators/dataplane-operator/issues/114

controllers/openstackdataplanenode_controller.go
```diff
@@ -382,6 +382,9 @@ func (r *OpenStackDataPlaneNodeReconciler) GetInstanceRole(ctx context.Context,
 			Namespace: instance.Namespace,
 			Name:      instance.Spec.Role,
 		}, instanceRole)
+		if err == nil {
+			err = instance.Validate(*instanceRole)
```
bogdando, line 386: I'm lat to this, sorry. So, if there are fields no longer unique between role and nodes, this would lock the controller out from reconciling any changes. Is that what we expect? My understanding is that any validation should happen on the API change submission, by validating/mutating webhooks. Not sure about the need for futher validation of internal state after weebhooks did the job, probably such extra validations in the controller code should only warn users, at the most?

controllers/openstackdataplanerole_controller.go
```diff
@@ -129,6 +129,10 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 			return ctrl.Result{}, err
 		}
 	}
+	err = instance.Validate(nodes.Items)
```
bogdando, line 132: ditto, a validation error for a node, should not be locking out the whole role from futher reconciliation?

api/v1beta1/common.go
bogdando: this can be moved into lib-common later on

controllers/openstackdataplanenode_controller.go
fao89: we can't do the API validation, because node and role share the same data structure, but I didn't know about the webhooks, could you please share an example of a webhook validation?

PR #97, opened 2023-03-16
Add sample for a full edpm_compute_0

slagle:
This example for edpm_compute_0 can be used without first creating a
role. A full set of required ansibleVars is set on the node, so nothing
needs to be inherited from the role.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml
@@ -1,0 +1,27 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNode
+metadata:
+  name: edpm-compute-0
+spec:
+  hostName: edpm-compute-0
+  ansibleHost: 192.168.122.100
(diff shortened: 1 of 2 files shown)


config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNode
+metadata:
+  name: edpm-compute-0
+spec:
+  hostName: edpm-compute-0
+  ansibleHost: 192.168.122.100
+  ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+  node:
```
SeanMooney, line 9: by the way i assume at somepoint we can update that with the NovaExternalComputeTemplate

its not currently in the CRD but we will need that sooner rather then later

slagle:
> please add it here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/kustomization.yaml

Done.

fao89 (approved):

PR #104, opened 2023-03-18
Use NewNovaExternalComputeSpec to get proper defaults

gibizer:
Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/293


gibizer:
I tested this together with https://github.com/openstack-k8s-operators/nova-operator/pull/293 and it resolves the defaulting issue

fao89 (approved):

api/go.mod
```diff
@@ -17,17 +17,19 @@ require (
...
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	k8s.io/api v0.26.1 // indirect
+	k8s.io/api v0.26.2 // indirect
 	k8s.io/klog/v2 v2.80.1 // indirect
 	k8s.io/utils v0.0.0-20230115233650-391b47cb4029 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace github.com/openstack-k8s-operators/nova-operator/api => github.com/gibizer/nova-operator/api v0.0.0-20230318182909-7589c2d0c019
```
slagle, line 35: we can remove this :smile:

go.mod
```diff
@@ -82,3 +82,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/openstack-k8s-operators/nova-operator/api => github.com/gibizer/nova-operator/api v0.0.0-20230318182909-7589c2d0c019
```
slagle, line 86: ditto

slagle (changes requested):
just need to cleanup the "replace" statements, otherwise LGTM.

api/go.mod
slagle: ah, I see your comment now about updating this once the nova-operator PR lands.

gibizer: yepp the nova PR landed so I will respin this to depend directly to the merged version. Then when this PR merges I will propose a bump in the openstack-operator that takes both the recent nova-operator and the recent dataplane-operator version used there so we can move in sync.

gibizer:
/unhold
The needed nova-operator PR has been landed

PR #110, opened 2023-03-21
Include openstack services containers related vars

raukadah:
In Upstream and Downstream CI, we pull containers
from different registery, namespace and image tag.

In order to use the same, we need add these vars
in dataplane sample crd. So that it gets
consumed in CI.


fao89 (commented):
+2

slagle (approved):

PR #113, opened 2023-03-21
Set controller reference on nodes and roles created by dataplane

### LINKED ISSUE #111: When OpenStackDataPlane is deleted the child OpenStackDataPlaneRole and OpenStackDatalPlaneNode CRs are not deleted.
It seems neither `CreateDataPlaneNode` and `CreateDataPlaneRole` set controller reference to the created resource:

https://github.com/openstack-k8s-operators/dataplane-operator/blob/aa6aba7338fa999ed2fd8205d8772c79446bf059/controllers/openstackdataplane_controller.go#L256-L316

I also noticed that these functions use client.Create which means this code cannot update the already created resource in a later reconciliation. I suggest to use `controllerutil.CreateOrPatch` instead of the simple Create.

fao89:
closes #111


slagle (approved):

gibizer (approved):
thanks

PR #118, opened 2023-03-22
Use correct conditions for DataPlane

slagle:
The DataPlane controller should use the standard ReadyCondition from
lib-common. The condition messages should also be DataPlane specific,
and not the role messages.


slagle:
> you need to drop DataPlaneRoleReadyCondition from: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplane_types.go#L92 otherwise it will keep as unknown

Done.

fao89 (approved):

PR #120, opened 2023-03-22
Use CreateOrPatch to sync roles and nodes from DataPlane

slagle:
The DataPlane controller now uses CreateOrPatch to sync the roles and
nodes instead of just calling Create when they don't exist. This fixes
setControllerReference not taking effect on the Node since it was
only set on the struct, and never passed to the API.

To avoid infinite looping, the role's DeployStrategy is explicitly
omitted from patching. Since the DataPlane controller already can set
role.Spec.DeployStrategy.Deploy in a different CreateOrPatch, infinite
looping between the DataPlane and DataPlaneRole controller could result
if there was another spot where role.Spec.DeployStrategy.Deploy was
being set to a different value.


fao89 (commented):
LGTM

fao89 (approved):

PR #122, opened 2023-03-23
Bump the number of concurrent Reconciles

fao89:
https://sdk.operatorframework.io/docs/building-operators/golang/tutorial/#controller-configurations

### DIFF
--- a/controllers/openstackdataplane_controller.go
+++ b/controllers/openstackdataplane_controller.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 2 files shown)


controllers/openstackdataplane_controller.go
```diff
@@ -226,6 +227,7 @@ func (r *OpenStackDataPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error
 		For(&dataplanev1beta1.OpenStackDataPlane{}).
 		Owns(&dataplanev1beta1.OpenStackDataPlaneNode{}).
 		Owns(&dataplanev1beta1.OpenStackDataPlaneRole{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 2}).
```
fao89, line 230: I believe we need 2 for dataplane and dataplanerole reconciles

gibizer: I believe that even if it set to 1 it should queue up the reconcile request and execute them sequentially.

slagle: I'm not sure about this. I don't see why we would want >1 reconciles.
FWIW, I'm not seeing the deploy not starting issue when I test with https://github.com/openstack-k8s-operators/dataplane-operator/pull/120 locally.

fao89: I'll close it then

PR #124, opened 2023-03-23
Remove variables from sample

fao89:
ref: https://github.com/openstack-k8s-operators/openstack-operator/pull/247

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplane.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane.yaml
@@ -16,7 +16,7 @@ spec:
           internal_api_ip: 172.17.0.100
           storage_ip: 172.18.0.100
           tenant_ip: 172.10.0.100
-          fqdn_internal_api: '{{ ansible_fqdn }}'
+          fqdn_internal_api: edpm-compute-0.example.com
         ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
       deployStrategy:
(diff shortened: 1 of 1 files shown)


config/samples/kustomization.yaml
```diff
@@ -4,7 +4,7 @@ resources:
 - dataplane_v1beta1_openstackdataplanenode_edpm_compute_0.yaml
 - dataplane_v1beta1_openstackdataplanenode_edpm_compute_1.yaml
 - dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml
-- dataplane_v1beta1_openstackdataplane.yaml
+- dataplane_v1beta1_openstackdataplane_sample.yaml
```
fao89, line 7: dataplane_v1beta1_openstackdataplane does have jinja vars: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane.yaml#L114-L119
I made a copy of it, but removing the vars and the CI got fixed

slagle: Nice find. Instead of removing the vars though, can we just expand them to what their default values would be? Then include that version in the CSV.

slagle (changes requested):

config/samples/kustomization.yaml
fao89: done!

gibizer (approved):

PR #126, opened 2023-03-23
Trigger role reconcile for NovaExternalCompute update

slagle:
OpenStackDataPlaneRole was not being reconciled when the associated
NovaExternalCompute's it had started were updated. This was due to 2
reasons.

First, the manager was not setup to Own the resources. This patch also
adds the other missing Owns, and adds the kubebuilder rules for the
other resources that both OpenStackDataPlaneRole and
OpenStackDataPlaneNode should own.

Second, the controller reference set on NovaExternalCompute was for the
node. In the case of a role deploy, this needs to be the role, so that
the role reconcile is triggered. The function signature of
DeployNovaExternalCompute needed to change to take an additional owner
parameter so that the owner can be set as either the node or role
depending on which type of deploy is triggered.

With these changes, roles are now reconciled a final time after all
NovaExternalCompute's are Ready, which is needed to set the final ready
state to true on the role.

### DIFF
--- a/controllers/openstackdataplanerole_controller.go
+++ b/controllers/openstackdataplanerole_controller.go
@@ -43,6 +43,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	novav1beta1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
+	"github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
 )
 
 // OpenStackDataPlaneRoleReconciler reconciles a OpenStackDataPlaneRole object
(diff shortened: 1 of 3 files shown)


fao89 (commented):
+2

controllers/openstackdataplanerole_controller.go
```diff
@@ -264,8 +272,8 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		// Return when any condition is not ready, otherwise set the role as
 		// deployed.
 		if len(novaReadyConditionsTrue) < len(nodes.Items) {
-			r.Log.Info("Not all NovaExternalCompute ReadyConditions are true")
-			return result, nil
+			r.Log.Info("Not all NovaExternalCompute ReadyConditions are true.")
```
gibizer, line 275: As a future enhancement we could log which NovaExternalCompute is not ready to easy the troubleshooting.

gibizer (approved):
It seems it works as expected and the OpenStackDataPlane now becomes Ready as well as the Role and NovaExternalCompute. So I approve this. 

Unfortunately the OpenstackDataPlaneNode still not Ready when the Node is deployed via a Role. It is pretty missleading to see the conditions on the Node being in init while the real node is already deployed:

```
$ oc get -o yaml  OpenStackDataPlaneNode | yq '.items[].status.conditions[] | select(.type == "Ready")'
{
  "lastTransitionTime": "2023-03-24T10:36:06Z",
  "message": "Init",
  "reason": "Init",
  "status": "Unknown",
  "type": "Ready"
}
{
  "lastTransitionTime": "2023-03-24T10:36:06Z",
  "message": "Init",
  "reason": "Init",
  "status": "Unknown",
  "type": "Ready"
}
gibi@bedrock:~/upstream/git/openstack-k8s-operators/install_yamls D[master]                                                                                                                          [000] [11:50:26]
$ oc get -o yaml  OpenStackDataPlaneRole | yq '.items[].status.conditions[] | select(.type == "Ready")'
{
  "lastTransitionTime": "2023-03-24T10:43:00Z",
  "message": "DataPlaneRole ready",
  "reason": "Ready",
  "status": "True",
  "type": "Ready"
}
```
We probably need to improve the status of the Node based on the Role it belongs.


There is still some requeue returns that can be removed and rely on the fact that Role / Node is reconciled with the job it started changes state. This would reduce the amount of logging we emit periodically.

PR #128, opened 2023-03-24
Update node status when deployed by role

fao89:


controllers/openstackdataplanerole_controller.go
```diff
@@ -281,6 +281,14 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		instance.Status.Deployed = true
 		r.Log.Info("Set ReadyCondition true")
 		instance.Status.Conditions.Set(condition.TrueCondition(condition.ReadyCondition, dataplanev1beta1.DataPlaneRoleReadyMessage))
+		for _, node := range nodes.Items {
+			node.Status.Conditions.MarkTrue(condition.ReadyCondition, fmt.Sprintf("Deployed by role.Name %s role.Namespace %s", instance.Name, instance.Namespace))
+			node.Status.Deployed = true
+			err = helper.PatchInstance(ctx, &node)
```
gibizer, line 287: will this trigger a Reconcile on the OpenstackDataPlaneNode object? If so then that Reconcile might reset the status condition.

fao89:
move to draft as I haven't tested yet

controllers/openstackdataplanerole_controller.go
abays: This [1] makes it seem like updates (patches too?) to status do trigger a reconcile, but it's from 2019.

[1] https://github.com/kubernetes-sigs/kubebuilder/issues/618

fultonj:
FWIW I tested this patch and it worked for me.

slagle (changes requested):

controllers/openstackdataplanerole_controller.go
```diff
@@ -281,6 +281,14 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+			if err != nil {
+				r.Log.Error(err, "Error on updating node status for ", node.Name)
```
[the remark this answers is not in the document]
fultonj: I hit that too but didn't hit it after this change. 

Are you saying you have that and you still get that error? 

https://github.com/openstack-k8s-operators/dataplane-operator/compare/433d64926e0e37c48732055d434f37f4d65bf58d..b0737b7bd60f82f62f6bd6ce8952906d3e3da97e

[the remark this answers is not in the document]
fultonj: I see. You hit a similar issue on a different line. 289 != 254. So this patch still needs that fix.

controllers/openstackdataplanenode_controller.go
```diff
@@ -138,7 +138,8 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 	// persist any changes.
 	defer func() {
 		// update the overall status condition if service is ready
-		if instance.IsReady() {
+		if instance.IsReady() || instanceRole.IsReady() {
+			instance.Status.Deployed = true
```
slagle, line 142: there's nothing to trigger a node reconcile so that this would get run

slagle (changes requested):

controllers/openstackdataplanenode_controller.go
slagle: Yea, i think we need that loop back. This worked for me:

```
Mon Mar 27 01:55:14 (jslagle@teletran-1) [~/code/github/openstack-k8s-operators/dataplane-operator] (nodestatus)$ git diff
diff --git a/controllers/openstackdataplanerole_controller.go b/controllers/openstackdataplanerole_controller.go
index aa35f12..1af8bbf 100644
--- a/controllers/openstackdataplanerole_controller.go
+++ b/controllers/openstackdataplanerole_controller.go
@@ -282,6 +282,15 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
                r.Log.Info("Set ReadyCondition true")
                instance.Status.Conditions.Set(condition.TrueCondition(condition.ReadyCondition, dataplanev1beta1.DataPlaneRoleReadyMessage))
 
+               for _, node := range nodes.Items {
+                       _, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), &node, func() error {
+                               node.Status.Deployed = true
+                               return nil
+                       })
+                       if err != nil {
+                               return ctrl.Result{}, err
+                       }
+               }
                // Explicitly set instance.Spec.Deploy = false
                // We don't want another deploy triggered by any reconcile request, it should
                // only be triggered when the user (or another controller) specifically
```

fao89: done!

fao89:
CI fix: https://github.com/openstack-k8s-operators/dataplane-operator/pull/140

PR #129, opened 2023-03-24
Use node or role name for OpenStackAnsibleEE

slagle:
Instead of using a randomly generated name for the OpenStackAnsibleEE,
use a name containing the role or node name. This makes the names
predictable and helps with discoverability of the right job for the
corresponding node/role.


fultonj (approved):
+2

PR #150, opened 2023-04-12
Combine OpenStack roles into OS roles

slagle:
The OpenStack roles are combined into the OS role executions. Since the
NovaExternalCompute CRD handles deploying nova libvirt and compute,
there isn't really a reason for a separate OpenStack execution within
the dataplane-operator itself. This also results in one less
OpenStackAnsibleEE resource and ansible execution.


bshephar (approved):
lgtm

fultonj (commented):
It looks like this will be obsoleted by the following

https://github.com/openstack-k8s-operators/dataplane-operator/pull/158

SeanMooney:
well the reason to seperate them was originally  because we agreed to have 2 phases bootstrap (all the work to configure the host that is not related to the OpenStack services) and then the OpenStack phase.

but as long as you ensure the ordering it won't break anything right now although it will effectively move the nova execution later if additional roles are added via the composable service interface unless those always happen after the NovaExternalCompute Cr is created.

I don't know of any reason why you could not define that the third-party roles always run in a third phase I guess which would alleviate that issues but i think there is an argument to be made that its clean to separate os configuration, OpenStack configuration and third-party integations into separate phases with a strict order but perhaps that is too rigid.

PR #155, opened 2023-04-19
Add handling for custom nic config

bshephar:
Users should be able to provide their own custom jinja2 or static nic-config template to be used during the os-net-config stage. This change adds handling for a nic-config template file, and wires in the relevant mounts to ensure the file is present within the AEE container.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplane_customnetworks.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane_customnetworks.yaml
@@ -1,0 +1,174 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlane
+metadata:
+  name: openstack-edpm
+spec:
+  deployStrategy:
+    deploy: false
+  nodes:
+    edpm-compute-0:
+      role: edpm-compute
+      hostName: edpm-compute-0
+      ansibleHost: 192.168.1.5
+      node:
+        ansibleVars: |
+          ctlplane_ip: 192.168.1.5
+          internal_api_ip: 172.17.0.101
+          storage_ip: 172.18.0.101
+          tenant_ip: 172.10.0.101
+          external_ip: 172.20.12.76
+          fqdn_internal_api: edpm-compute-1.example.com
+          ansible_ssh_transfer_method: scp
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+        ansibleUser: root
+      deployStrategy:
+        deploy: true
+  roles:
+    edpm-compute:
+      env:
+        - name: ANSIBLE_FORCE_COLOR
+          value: "True"
+        - name: ANSIBLE_ENABLE_TASK_DEBUGGER
+          value: "True"
+        - name: ANSIBLE_VERBOSITY
+          value: 4
+        - name: ANSIBLE_HOST_KEY_CHECKING
+          value: false
+      nodeTemplate:
+        managed: false
+        managementNetwork: ctlplane
+        ansibleUser: root
+        ansiblePort: 22
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+        networkConfig:
+          template: |
+            ---
+            network_config:
+            - type: interface
+              name: nic2
+              mtu: 1500
+              addresses:
+                - ip_netmask:
+                    {{ ctlplane_ip }}/{{ ctlplane_subnet_cidr }}
+            - type: ovs_bridge
+              name: {{ neutron_physical_bridge_name }}
+              mtu: 1500
+              use_dhcp: false
+              dns_servers: {{ ctlplane_dns_nameservers }}
+              domain: []
+              addresses:
+              - ip_netmask: {{ lookup('vars', networks_lower["External"] ~ '_ip') }}/{{ lookup('vars', networks_lower["External"] ~ '_cidr') }}
+              routes: [{'ip_netmask': '0.0.0.0/0', 'next_hop': '192.168.1.254'}]
+              members:
+              - type: interface
+                name: nic1
+                mtu: 1500
+                # force the MAC address of the bridge to this interface
+                primary: true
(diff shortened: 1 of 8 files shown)


config/samples/dataplane_v1beta1_openstackdataplane_customnetworks.yaml
```diff
@@ -0,0 +1,174 @@
...
+          value: 4
+        - name: ANSIBLE_HOST_KEY_CHECKING
+          value: false
+      nodeTemplate:
+        managed: false
+        managementNetwork: ctlplane
+        ansibleUser: root
+        ansiblePort: 22
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+        networkConfig:
```
kajinamit, line 43: Wondering if we want to allow users to use built-in templates instead of requiring the whole templates always.

config/samples/dataplane_v1beta1_openstackdataplane_customnetworks.yaml
```diff
@@ -0,0 +1,174 @@
...
+          template: |
+            ---
+            network_config:
+            - type: interface
+              name: nic2
+              mtu: 1500
+              addresses:
+                - ip_netmask:
+                    192.168.1.5/24
```
kajinamit, line 52: Can we use vars in this template (eg. `{{ ctlplane_ip }}` , `{{ ctlplane_mtu}}` )

bshephar: Yeah, it can be a jinja2 template. It's just converted to a string, written to the ConfigMap and then mounted in the AEE environment. So you can provide a Jinja2 formatted template like we normally would.

config/samples/dataplane_v1beta1_openstackdataplane_customnetworks.yaml
bshephar: It does seem as though there is precedent for setting Ansible vars like this, so I added handling here:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/155/files#diff-506be0d5f31d8a915bfd2ae4c8b305856930dc22e9d8a443813b90e9a8b01356R407

Maybe that would be preferable. Then if there is NOT a `NetworkConfig.Template` file provided, then we will revert to having the user define one of the default templates?

controllers/openstackdataplanenode_controller.go
```diff
@@ -316,11 +316,13 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
...
 	host := all.AddHost(instance.Name)
 
+	networkConfig := r.GetAnsibleNetworkConfig(instance, instanceRole)
+
 	host.Vars["ansible_user"] = r.GetAnsibleUser(instance, instanceRole)
 	host.Vars["ansible_port"] = r.GetAnsiblePort(instance, instanceRole)
 	host.Vars["managed"] = r.GetAnsibleManaged(instance, instanceRole)
 	host.Vars["management_network"] = r.GetAnsibleManagementNetwork(instance, instanceRole)
-	host.Vars["network_config"] = r.GetAnsibleNetworkConfig(instance, instanceRole)
+	host.Vars["network_config"] = deployment.NicConfigTemplateFile
```
bshephar, line 325: This won't work if we want to allow users to choose a default template actually. Might need to conditionally set this based on the value of `networkConfig`. Which is where it previously got its value from, only that the return value of that function now returns the actual template.

So if `networkConfig != ""`. Then we set this to `deployment.NicConfigTemplateFile` else, we don't set it?

slagle (commented):
the approach looks good to me here

pkg/util/ansible_execution.go
```diff
@@ -119,11 +123,17 @@ func AnsibleExecution(
 			MountPath: "/runner/inventory/hosts",
 			SubPath:   "inventory",
 		}
+		networkConfigMount := corev1.VolumeMount{
+			Name:      "inventory",
+			MountPath: "/runner/network/nic-config-template",
```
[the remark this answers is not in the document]
bshephar: The constant is defined in the `deployment` package and we aren't importing it here, importing it would create a cyclic import. Since it was only used once here, I just followed what we did for the inventory/hosts file above rather than redefining a non-exported const in the utils package.

controllers/openstackdataplanenode_controller.go
```diff
@@ -363,6 +365,7 @@ func (r *OpenStackDataPlaneNodeReconciler) GenerateInventory(ctx context.Context
 		}
 		cm.Data = map[string]string{
 			"inventory": string(invData),
+			"network":   string(networkConfig.Template),
```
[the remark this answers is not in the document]
bshephar: Yeah, the only reason I chose to append to this one instead of creating a new one is due to the design of https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/deployment.go#L79

I didn't want to add a new param to each and every call to `ConditionalDeploy`. 

We can rename it, although since we're just defining a template file used by Ansible, I thought it was still essentially in-line with everything else that configmap already contained. 

If the preference is to rename though, I can do that.

config/samples/dataplane_v1beta1_openstackdataplane_customnetworks.yaml
```diff
@@ -0,0 +1,174 @@
...
+          storage_mtu: 1500
+          storage_vlan_id: 21
+          storage_cidr: '24'
+          storage_host_routes: []
+          tenant_mtu: 1500
+          tenant_vlan_id: 22
+          tenant_cidr: '24'
+          tenant_host_routes: []
+          role_networks:
+          - InternalApi
```
kajinamit, line 124: This is off-topic: We no longer need these network names in heat resources and can probably get rid of the redundant conversion using networks_lower, though the final var structure would be dependent on what we get from baremetal provisioning.

kajinamit (approved):

PR #158, opened 2023-04-20
Add Composable Service Interface

slagle:
- Add Composable Service CRD and controllers
- Add OpenStackDataPlaneService deployment
- Migrate ConfigureNetwork to the compsable service interface


config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    app.kubernetes.io/name: openstackdataplaneservice
+    app.kubernetes.io/instance: openstackdataplaneservice-sample
+    app.kubernetes.io/part-of: dataplane-operator
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: dataplane-operator
+  name: configure-network
+spec:
+  label: dataplane-deployment-configure-network
+  role:
```
SeanMooney, line 13: this looks similar but not the same as how you specify a play when defining an in line playbook for the AEE operator.

is there a reason you are diverging and using roles here's nice the Hosts and stragey fields make this look more like a playbook then an ansibel role.?

slagle: I can't make sense of what you're asking. Can you phrase it a different way?

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -0,0 +1,82 @@
...
+// OpenStackDataPlaneServiceSpec defines the desired state of OpenStackDataPlaneService
+type OpenStackDataPlaneServiceSpec struct {
+	// Label to use for service
+	// +kubebuilder:validation:Optional
+	Label string `json:"label,omitempty"`
+
+	// Role is the description of an Ansible Role
+	// +kubebuilder:validation:Optional
+	Role *ansibleeev1alpha1.Role `json:"role,omitempty"`
+}
```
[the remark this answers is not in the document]
gibizer: @karelyatin @slawqo I think this discussion are interesting to you too as it involves the design of deploying networking services on edpm node.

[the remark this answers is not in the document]
slagle: The OpenStackDataPlaneService CRD is ansible specific in that it will only equate to OpenStackAnsibleEE CR's. The ansible roles or plays executed by the OpenStackAnsibleEE will be fully configurable with the existing ansibleVars interface we already have.

For things like RabbitMQCluster/TransportURL, the dataplane-operator can look up the associated values given an OpenStackControlPlane name, and set those values within ansibleVars for any ansible role to consume.

I'm not aware of plans by the networking folks to change the way that the networking related services are currently being deployed between dataplane-operator/edpm-ansible. In the EDPM sync meeting, we brought this up and I believe @slawqo was there and indicated there was no planned work to change that. If that is incorrect, we need to draw attention to that, in terms of what issues exist and who is planning on doing the work within the other operators and dataplane-operator.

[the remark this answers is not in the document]
gibizer: > For things like RabbitMQCluster/TransportURL, the dataplane-operator can look up the associated values given an OpenStackControlPlane name, and set those values within ansibleVars for any ansible role to consume.

Unfortunately this introduce a dependency cycle. Similar to what we had when TransportURL CRD was defined by OpenstackOperator. We later moved that CRD to a separate infra-operator to break the cycle.

Any change that result in a new version of openstack-operator will cause a new version of the apis module of the openstack-operator the dataplane-operator depends on. So the dataplane-operator will bump the version of the openstack-operator apis module in its go.mod file (and do the necessary adaptation work if needed). This will result in a new dataplane-operator version (i.e. new git hash on master). This in return will trigger a version bump in the openstack-operator's go.mod file for the dataplane-operator dependency. When that bump is merged in openstack-operator a new version of the opeanstack-operator apis module (a new git hash) is produced and the cycle restarts.

With the nova-operator approach we don't have a dependency cycle as only dataplane-operator depends on nova-operator's api module but nova-operator does not mention dataplane-operator in any of its go.mod files.

> I'm not aware of plans by the networking folks to change the way that the networking related services are currently being deployed between dataplane-operator/edpm-ansible. In the EDPM sync meeting, we brought this up and I believe @slawqo was there and indicated there was no planned work to change that. If that is incorrect, we need to draw attention to that, in terms of what issues exist and who is planning on doing the work within the other operators and dataplane-operator.

I agree we need to settle this sooner than later.

config/samples/dataplane_v1beta1_openstackdataplane.yaml
```diff
@@ -36,6 +36,8 @@ spec:
         deploy: false
   roles:
     edpm-compute:
+      services:
+        - configure-network
```
fultonj, line 40: In this example only the `configure-network` service is on the services list. If other services are migrated the same way (e.g. `InstallOS`, `ConfigureOS`, etc from deployment.go) then it will be easy for the deployer to introduce deployment break points. 

This would be useful to me for the HCI usecase because I could shorten this list to only `configure-network`, and then once the storage network is configured, I could go and deploy ceph manually. I could then come back and extend the list with `configure-os` etc to finish the deployment.

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
fultonj: > > For things like RabbitMQCluster/TransportURL, the dataplane-operator can look up the associated values given an OpenStackControlPlane name, and set those values within ansibleVars for any ansible role to consume.
> 
> Unfortunately this introduces a dependency cycle. Similar to what we had when TransportURL CRD was defined by OpenstackOperator. We later moved that CRD to a separate infra-operator to break the cycle.

This patch doesn't introduce that kind of cycle though. I don't see that kind of import in any of this code.

I just tested this patch in my environment and `oc edit openstackdataplanes.dataplane.openstack.org` reveals that the `edpm_ovn_metadata_agent_DEFAULT_transport_url` ansible var is set the same way it has been set.

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
fultonj: Can we keep this discussion moving so we can make progress?

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+    name: "Deploy EDPM Network"
+    hosts: "all"
+    strategy: "linear"
+    tasks:
+      - name: "Install edpm_bootstrap"
+			  import_role:
```
fultonj, line 20: I can't create this CR in my enviornment:

[fultonj@osp-storage-01 composable]$ oc create -f ~/dataplane-operator/config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
error: error parsing /home/fultonj/dataplane-operator/config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml: error converting YAML to JSON: yaml: line 19: did not find expected key
[fultonj@osp-storage-01 composable]$ 

I see (using whitespace mode in my editor) that there are four tabs and then two spaces before the first `import_role` on line 20. I'll see if I can fix locally and share.

I passed it to https://www.yamllint.com and it said:

```
Tabs are not allowed as indentation at line 20, column 1
Tabs are not allowed as indentation at line 23, column 1
Tabs are not allowed as indentation at line 24, column 1
A block sequence may not be used as an implicit map key at line 26, column 1
Implicit keys need to be on a single line at line 26, column 7
Tabs are not allowed as indentation at line 30, column 1
Implicit map keys need to be followed by map values at line 26, column 7
```

slagle: should be fixed now.

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
```diff
@@ -0,0 +1,59 @@
...
+        import_role:
+          name: "osp.edpm.edpm_bootstrap"
+          tasks_from: "bootstrap.yml"
+          tags:
+            - "edpm_bootstrap"
+      - name: "Grow volumes"
+        import_role:
+          name: "osp.edpm.edpm_growvols"
+          tasks_from: "main.yml"
+        tags: "edpm_growvols"
```
fultonj, line 29: Please change line 29 to the following so it takes line 29 and 30:
```
        tags:
          - "edpm_growvols"
```
Otherwise you'll get 
```
The OpenStackDataPlaneService "configure-network" is invalid: spec.role.tasks[1].tags: Invalid value: "string": spec.role.tasks[1].tags in body must be of type array: "string"
```

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
fultonj: It's my understanding that these issues have been resolved in offline conversations. Please chime in if there's still any objections to merging this.

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
fultonj: It's my understanding that these issues have been resolved in offline conversations. Please chime in if there's still any objections to merging this.

fultonj (commented):
+2, provided the tags list be turned into a list in the next revision in config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml

With the above update, I was able to deploy a working EDPM node where I could boot an instance.

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
slagle: Done.

fultonj (commented):
+2

I think we can merge this now.

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
slagle: To clarify, the discussion about changing ovn deployment from dataplane-operator to ovn-operator seemed to settle around the consensus that what we have now with dataplane-operator is fine to proceed. If the OVN folks want to look at migrating that out of dataplane-operator later, then that is still possible.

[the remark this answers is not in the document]
gibizer: I rest my case here. As @fultonj stated above the current patch does not introduce any dependency cycle. My point about the cycle was a response to @slagle proposal to look up the RabbitMQCluster/TransportURL CR from dataplane-operator.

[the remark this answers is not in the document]
gibizer: I do feel that the current way we fill transport URL configuration via install_yamls is not a nice end user interface though. But that discussion is definitely independent from this PR.

config/samples/dataplane_v1beta1_openstackdataplaneservice.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/name: openstackdataplaneservice
+    app.kubernetes.io/instance: openstackdataplaneservice-sample
+    app.kubernetes.io/part-of: dataplane-operator
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: dataplane-operator
+  name: openstackdataplaneservice-sample
+spec:
+  # TODO(user): Add fields here
```
fao89, line 12: why not use `configure-network` spec here?

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
fao89: what is the idea of having `services` instead of using `samples`?

controllers/openstackdataplanenode_controller.go
```diff
@@ -453,3 +453,9 @@ func (r *OpenStackDataPlaneNodeReconciler) GetAnsibleSSHPrivateKeySecret(instanc
 	}
 	return instanceRole.Spec.NodeTemplate.AnsibleSSHPrivateKeySecret
 }
+
+// GetServices returns the list of services for the node's role
+// Note that these are not inherited from NodeTemplate.
+func (r *OpenStackDataPlaneNodeReconciler) GetServices(instance *dataplanev1beta1.OpenStackDataPlaneNode, instanceRole *dataplanev1beta1.OpenStackDataPlaneRole) []string {
+	return instanceRole.Spec.Services
+}
```
fao89, line 461: shouldn't we get `instance.Spec.Services` when the node is deployed individually?

fao89 (commented):
left a couple of questions, but LGTM

+2

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
slagle: These aren't samples. They will be the actual services CR's that need to be created for the default dataplane services.

config/samples/dataplane_v1beta1_openstackdataplaneservice.yaml
slagle: this one *is* the sample, so I didn't make it specific to configure network.

fao89:
Note: I didn't formally approve because I'm not sure if other people are still reviewing

controllers/openstackdataplanenode_controller.go
slagle: i've just set the services on the role for now. I did not make this an inherited field in NodeTemplate initially. but perhaps we want to?

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
SeanMooney: i was refing to useing the term role here and not play.

roles in ansibel do not have a host selector or execution stragey

    name: "Deploy EDPM Network"
    hosts: "all"
    strategy: "linear"

those are set on a playbook

so the content of the role element below is actually an abstraction fo an ansibel playbook

in the AEE operator we use the key "play" to model inline playbooks
https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/examples/openstack-ansibleee-play.yaml#L8-L14

the host key in particular I'm not sure makes sense to expose or at least require

implicitly, if you are deifing a composable service on a dataplaneNode it would only apply to that one host and similarly on a role it should only apply to that role.

I don't think we should be allowing you to run this on a subset of hosts within a role makes sense or it should not be the norm.
so if the host sub key is supported  then it should default to all and not be required

but that raise the question fo the type of this field.

here you have defined it as a mapping type but in the AEE operator its just a string so its a direct passthrough to ansible.

if you want to support Ansibel in this way accepting this as s string i think makes more sense and just define it inline.

this is not blocking by any means but i think we should be trying to keep this interface and the AEE interface consistent.

api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
```diff
@@ -0,0 +1,113 @@
...
+                    type: boolean
+                  become:
+                    default: false
+                    type: boolean
+                  gather_facts:
+                    default: false
+                    type: boolean
+                  hosts:
+                    default: '{{ primary_role_name | default([]) }}:{{ deploy_target_host
+                      | default(''overcloud'') }}'
```
SeanMooney, line 45: we shoudl not use the term overcloud in operator or other tripleo terminology.

there is not over and under cloud anymore

fultonj: I'm fine with `default(''overcloud'')` being removed. This seems to be the only open issue for this patch so maybe we can make that change and merge so we can move.

config/services/dataplane_v1beta1_openstackdataplaneservice_configurenetwork.yaml
SeanMooney: @fultonj this issue is not adressed

slagle: This is consistent with the Role interface from OpenStackAnsibleEEE. Which supports specifying roles based the Role struct defined in its openstack_ansibleee_types.go, and that is what I'm making use of here:

https://github.com/openstack-k8s-operators/dataplane-operator/pull/158/files#diff-be99dc960fff7094f26203ceac52ad519db728b17568a00d1bc1d9b48b811bacR34

That's also what we used in network.go (and still in other services not migrated to this interface).

fao89 (approved):

PR #163, opened 2023-04-24
Refactor deployment.go to use Interfaces

bshephar:
This change refactors the current implementation of deployment.go to leverage interfaces and structs for each deploy step.

This change standardises the implementation for each module based around a set of standard verbs used in edpm_ansible, and abstracts any of the complexity into the deployments.go file.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -19,6 +19,7 @@ package deployment
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


pkg/deployment/deployment.go
```diff
@@ -31,6 +32,35 @@ import (
...
+type Runner interface {
+	Run(context.Context, *helper.Helper, client.Object, string, string, dataplanev1beta1.AnsibleEESpec) error
+}
+
+type networker struct{}
+
+type OS struct{}
+
+type cephClient struct{}
+
```
bshephar, line 63: So, moving our Slack conversation onto the PR for consistency.

We could leverage these structs to further simplify this but adding name and label fields. Like:

```
type networker struct {
    Name string
    Label string
```

Then move the deployName and deployLabel into the struct to be reference by `ConditionalDeploy()`

kajinamit: Yes and then we can remove `deployLabel` argument from ConditionalDeploy, and also `deployName` should be just a simple action name (eg. `Configure`) without target (eg. `Network`, `OS`).

pkg/deployment/deployment.go
```diff
@@ -366,7 +396,17 @@ func ConditionalDeploy(
...
-		err = deployFunc(ctx, helper, obj, sshKeySecret, inventoryConfigMap, aeeSpec)
+		switch true {
+		case strings.Contains(deployName, "Configure"):
+			err = serviceInterface.Configure(ctx, helper, obj, sshKeySecret, inventoryConfigMap, aeeSpec)
+		case strings.Contains(deployName, "Validate"):
+			err = serviceInterface.Validate(ctx, helper, obj, sshKeySecret, inventoryConfigMap, aeeSpec)
+		case strings.Contains(deployName, "Install"):
+			err = serviceInterface.Install(ctx, helper, obj, sshKeySecret, inventoryConfigMap, aeeSpec)
+		case strings.Contains(deployName, "Run"):
+			err = serviceInterface.Run(ctx, helper, obj, sshKeySecret, inventoryConfigMap, aeeSpec)
```
bshephar, line 407: We can probably change these to just be a `deployVerb` instead and use the `Install` `Configure` `Run` `Validate`. But that will likely change all of the logging, so we might want to address that in a different PR. It would simplify this though, we could just use case statements without looking for substrings.

slagle:
I think this is an improvement to what is there presently, but I would like to consider that most of this code will refactored in a different way in favor of an approach shown in https://github.com/openstack-k8s-operators/dataplane-operator/pull/158/commits/36c378bd7e21085ebb338815b938ada12fb4d75f. That illustrates migrating ConfigureNetwork to a composable interface. The interface works for any service abstraction that is just an ansible execution via OpenStackAnsibleEE. I don't feel we need to use strongly typed go code for what is effectively role name(s) and task file name(s). We started that way in order to make quick progress for the demo, but I don't think that pattern is long lived.

bshephar:
> I think this is an improvement to what is there presently, but I would like to consider that most of this code will be refactored in a different way in favor of the approach shown in [36c378b](https://github.com/openstack-k8s-operators/dataplane-operator/commit/36c378bd7e21085ebb338815b938ada12fb4d75f).
> 
> That commit illustrates migrating ConfigureNetwork to a composable service interface. The interface works for any service abstraction that is just an ansible execution via OpenStackAnsibleEE. I don't feel we need to use strongly typed go code for what is effectively role name(s) and task file name(s). We started that way in order to make quick progress for the demo, but I don't think that pattern is long lived.

No worries. If we have a better solution in the works, I'm happy to abandon this effort in favour of the preferred solution.

PR #164, opened 2023-04-24
Allow nova configurability on the EDPM node

gibizer:
This patch exposes a nova specific template struct within the
DataplaneRole and Node that allows the end user to pass compute
specific configuration to the EDPM node. The template in the Node has
precedence over the template in the Role. If neither is defined then it
indicates that the EDPM node does not need compute services to be
deployed (i.e. it is not a compute node)

Closes: https://issues.redhat.com/browse/OSPRH-93


api/v1beta1/common.go
```diff
@@ -162,3 +169,39 @@ type AnsibleEESpec struct {
 	// Env is a list containing the environment variables to pass to the pod
 	Env []corev1.EnvVar `json:"env,omitempty"`
 }
+
+// NovaTemplate specifies the parameters for the compute service deployment on
+// the EDPM node.
+type NovaTemplate struct {
+	// +kubebuilder:validation:Optional
```
SeanMooney, line 176: defaulting to cell 1 is tempting but i think we are going to see multi cell deployments become more common so long term this should be required not optional.

can address that in a separate pr but we should make this change in Q2.

api/v1beta1/common.go
```diff
@@ -162,3 +169,39 @@ type AnsibleEESpec struct {
...
+	NovaInstance string `json:"novaInstance"`
+
+	// +kubebuilder:validation:Optional
+	// CustomServiceConfig - customize the nova-compute service config using
+	// this parameter to change service defaults, or overwrite rendered
+	// information using raw OpenStack config format. The content gets added to
+	// to /etc/nova/nova.conf.d directory as 02-nova-override.conf file.
+	CustomServiceConfig string `json:"customServiceConfig"`
+
+	// +kubebuilder:validation:Optional
```
SeanMooney, line 194: we proably shoudl remvoe DefaultConfigOverwrite

i think we likely will want to remove it from all the nova CRD sortly so i don't think we need to expose it here since it wont actually work.

gibizer:
/test dataplane-operator-build-deploy

```
+ timeout 240 xargs -I '{}' -d ' ' sh -c 'oc wait --for=condition=Available deployment {} --timeout=-1s'
Error from server (NotFound): deployments.apps "openstack-operator-controller-manager" not found
```
Based on the get_all.log things were still progressing when the wait timed out. So probably a slow node.

controllers/openstackdataplanenode_controller.go
```diff
@@ -233,40 +233,47 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
-			return ctrl.Result{}, err
+			if condition.IsError(instance.Status.Conditions.Get(dataplanev1beta1.NovaComputeReadyCondition)) {
+				r.Log.Info(fmt.Sprintf("%s error", dataplanev1beta1.NovaComputeReadyCondition))
+				err = fmt.Errorf("failed: NovaExternalCompute name %s NovaExternalCompute namespace %s", novaExternalCompute.Name, novaExternalCompute.Namespace)
+				return ctrl.Result{}, err
+			}
 		}
 
-		if instance.Status.Conditions.IsTrue(dataplanev1beta1.NovaComputeReadyCondition) {
+		if instance.Spec.Node.Nova == nil || instance.Status.Conditions.IsTrue(dataplanev1beta1.NovaComputeReadyCondition) {
```
bogdando, line 276: maybe it worth noting that other Spec.Node.Foo as get defined, should all be added to this condition

controllers/openstackdataplanenode_controller.go
```diff
@@ -233,40 +233,47 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
+				instance,
+				ansibleSSHPrivateKeySecret,
+				nodeConfigMap,
+				&instance.Status,
+				instance.Spec.NetworkAttachments,
+				instance.Spec.OpenStackAnsibleEERunnerImage,
+				// There is anther nova template accessible in
+				// instanceRole.Spec.NodeTemplate.Nova
+				// Which one we should use here? Inheritance?
+				instance.Spec.Node.Nova,
```
bogdando, line 251: this operates by the instance instead of its role, so looks correct to me. @slagle WDYT?

docs/openstack_dataplanerole.md
```diff
@@ -78,6 +79,21 @@ NodeSection is a specification of the node attributes
...
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| cellName | CellName is the name of nova cell the compute(s) should be connected to | string | true |
+| novaInstance | NovaInstance is the name of the Nova CR that represents the deployment this compute belongs to | string | true |
+| customServiceConfig | CustomServiceConfig - customize the nova-compute service config using this parameter to change service defaults, or overwrite rendered information using raw OpenStack config format. The content gets added to to /etc/nova/nova.conf.d directory as 02-nova-override.conf file. | string | true |
+| defaultConfigOverwrite | DefaultConfigOverwrite - interface to overwrite default config files like e.g. logging.conf But can also be used to add additional files. Those get added to the service config dir in /etc/nova/ . | map[string]string | true |
+| deploy | Deploy true means the compute service(s) are allowed to be changed on the EDPM node if necessary. If set to false then only the pre-requisite data (e.g. config maps) will be generated but no actual modification on the compute node itself will happen. | bool | true |
```
bogdando, line 96: on the EDPM nodes of this role?

api/v1beta1/common.go
gibizer: I'm OK to make CellName a required field. But I would like to do that in a separate PR to limit the amount of change here.

api/v1beta1/common.go
gibizer: OK. I'm not exposing `DefaultConfigOverwrite` here then.

docs/openstack_dataplanerole.md
gibizer: NovaTemplate is use both in the Node and the Role so I need to word this carefully. I changed this to be more generic.

api/v1beta1/common.go
SeanMooney: thanbks, i have not figured out exactly how i want this to work but i think i want to leverage extra mounts in some way to enable this usecase of adding arbitrary files, but for now i want to defer that until we have time to sit down and think that through without being rushed.

effectively the extra mounts would be added to the AEE container then we would scp them to /var/lib/openstack/nova/<volume name> and modify the container bind mount that volume 

then we need to parse the mount info and copy the content to the correct location in the container using the kolla confing.json and kolla_set_conf script.

if we do that right i think we can make it work for the podman container the same as the openshift container but i need to actually poc that or sit down and draw it out to confrim

gibizer: ack

controllers/openstackdataplanenode_controller.go
```diff
@@ -233,40 +233,46 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
-			nodeConfigMap,
-			&instance.Status,
-			instance.Spec.NetworkAttachments,
-			instance.Spec.OpenStackAnsibleEERunnerImage)
-		if err != nil {
-			return result, err
-		}
+		novaTemplate := deployment.GetNovaTemplate(instance, instanceRole)
+
+		if novaTemplate != nil {
```
fultonj, line 238: I see we do not create the `novaExternalCompute` if the novaTemplate is empty. This is good since we may not alway want the EDPM node to be a compute node.

docs/openstack_dataplanenode.md
```diff
@@ -78,6 +79,20 @@ NodeSection is a specification of the node attributes
 | ansibleVars | AnsibleVars for configuring ansible | string | false |
 | ansibleSSHPrivateKeySecret | AnsibleSSHPrivateKeySecret Private SSH Key secret containing private SSH key for connecting to node. Must be of the form: Secret.data.ssh-privatekey: <base64 encoded private key contents> https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets | string | true |
 | extraMounts | ExtraMounts containing files which can be mounted into an Ansible Execution Pod | []storage.VolMounts | true |
+| nova | NovaTemplate specifies the parameters for the compute service deployment on the EDPM node.If it is specified both in OpenstackDataPlaneRole and the OpenstackDataPlaneNode for the same EDPM node then the configuration in OpenstackDataPlaneNode will be used and the configuration in the OpenstackDataPlaneRole will be ignored. If this is defined in neither then compute service(s) will not be deployed on the EDPM node. | *[NovaTemplate](#novatemplate) | true |
```
fultonj, line 82: nit, space "EDPM node.If"

docs/openstack_dataplanenode.md
```diff
@@ -78,6 +79,20 @@ NodeSection is a specification of the node attributes
...
+
+#### NovaTemplate
+
+NovaTemplate specifies the parameters for the compute service deployment on the EDPM node.
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| cellName | CellName is the name of nova cell the compute(s) should be connected to | string | true |
+| novaInstance | NovaInstance is the name of the Nova CR that represents the deployment the compute(s) belongs to | string | true |
+| customServiceConfig | CustomServiceConfig - customize the nova-compute service config using this parameter to change service defaults, or overwrite rendered information using raw OpenStack config format. The content gets added to to /etc/nova/nova.conf.d directory as 02-nova-override.conf file. | string | true |
```
fultonj, line 94: line 94 is good at explaining how this new feature is supposed to work.

docs/openstack_dataplanenode.md
```diff
@@ -78,6 +79,20 @@ NodeSection is a specification of the node attributes
...
+[Back to Custom Resources](#custom-resources)
+
+#### NovaTemplate
+
+NovaTemplate specifies the parameters for the compute service deployment on the EDPM node.
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| cellName | CellName is the name of nova cell the compute(s) should be connected to | string | true |
+| novaInstance | NovaInstance is the name of the Nova CR that represents the deployment the compute(s) belongs to | string | true |
```
fultonj, line 93: "the name of the Nova CR that represents the deployment the compute(s) belongs to" 
I have some idea of what this means but I'm not certain. If you can provide an example it would make it more clear.

I would add an extra paragraph, outside of this table, which shows a deployment ready to get an EDPM compute node. An example command would then retrieve a NovaCR value so the reader knows for sure what you are referring to.

pkg/deployment/nova.go
```diff
@@ -73,3 +90,16 @@ func DeployNovaExternalCompute(ctx context.Context, helper *helper.Helper, obj c
...
 }
+
+// GetNovaTemplate returns the NovaTemplate instance to be used. The NovaTemplate
+// in the OpenStackDataPlaneNode if defined takes precedence over the NovaTemplate
+// in the OpenStackDataPlaneRole.
+func GetNovaTemplate(
+	node *dataplanev1beta1.OpenStackDataPlaneNode,
+	role *dataplanev1beta1.OpenStackDataPlaneRole,
+) *dataplanev1beta1.NovaTemplate {
+	if node.Spec.Node.Nova == nil {
```
fultonj, line 101: good

fultonj (commented):
This looks good to me. I have a few suggestions for the doc patch to make it more clear.

bshephar:
I think I'm missing some context. We aren't using Ansible variables here to populate the config file? ie. we're not using any of the variables defined here to configure stuff?
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova_compute/defaults/main.yml

I know Nova operator maintains its own playbooks, but does this mean that a user would override any `nova.conf` values by setting `nova.properties.customServiceConfig`, and any values they define in `ansibleVars` will be ignored? 
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane.yaml#L52

pkg/deployment/nova.go
rabi: Hmm.. I thought we've an agreement to do everything for a role including external compute reconciliation in the dataplanerole[1]. This PR seems to be going in the opposite direction to what I'm doing in https://github.com/openstack-k8s-operators/dataplane-operator/pull/165.  We probably need to discuss and agree on this, else it would be lot of wasted effort. 

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/deployment.go#L348

rabi: To clarify a bit more, we should decide whether to implement all provisioning/deployment logic in  dataplanerole or dataplanenode reconciliation but not in both.

gibizer:
> I think I'm missing some context. We aren't using Ansible variables here to populate the config file? ie. we're not using any of the variables defined here to configure stuff? https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova_compute/defaults/main.yml
> 
> I know Nova operator maintains its own playbooks, but does this mean that a user would override any `nova.conf` values by setting `nova.properties.customServiceConfig`, and any values they define in `ansibleVars` will be ignored? https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane.yaml#L52

Even without the currently proposed change nova related ansible variables are ignore today as https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova_compute/defaults/main.yml is not used in dataplane-operator but instead NovaExternalCompute CRD is used to deploy the compute parts of the EDPM node.
The currently proposed `customServiceConfig` can be used to pass configuration options to the compute part of the EDPM node.

@SeanMooney can chime in why https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova_compute/defaults/main.yml is not used by nova-operator. 

I think that having `customServiceConfig` in NovaExternalCompute and therefore in OpenstackDataPlaneNode/Role aligns well with the user interface of the rest of the CRDs in NG. All the control plane services are configured via `customServiceConfig` fields in the various service operator CRDs. Having a single way to pass through service configuration in the whole NG will contribute to a system that is easier to understand and use.

pkg/deployment/nova.go
gibizer: I also missing some context here for sure. :) 

So far I understood that dataplane-operator can deploy individual EDPM nodes via DataplaneNode CR or a set of EDPM nodes via DataplaneRole. Hence my change try to keep both use cases supported with NovaExternalCompute. 

At least the [doc](https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/docs/inheritance.md) still says:
> It's also possible to create a node directly outside of a role CR and define its role. If the following CR is created

rabi: > E.g. Does it mean that the user will not be allowed to create DataplaneNode CR just Role CR? Or we still allow Node CR to be created but only with a valid Role reference?

I think the current user interface is to create a Dataplane CR (neither role/node directly). But there are some discussions I have seen about not having Dataplane CR at all.  I would probably do away with the DataplaneNode CR if all reconciliation is done with DataPlaneRole using one AnsibleEE job (i.e ansible play) per role.

Reg documented agreements, I have not seen one. My understanding is based on baremetal provisioning meeting discussions where we discussed dataplane-operator CRD design.

gibizer: Even if the end user created Dataplane CR I think it currently still possible to only define node in Dataplane CR and not any role.

https://github.com/openstack-k8s-operators/dataplane-operator/blob/c11eadc0c91a90de6b4e67e0e14098ea7fed3333/api/v1beta1/openstackdataplane_types.go#L32-L35 

https://github.com/openstack-k8s-operators/dataplane-operator/blob/c11eadc0c91a90de6b4e67e0e14098ea7fed3333/api/v1beta1/openstackdataplanenode_types.go#L38-L40

So the user can indirectly create Node CRs that are not referring to any Role CRs. I guess then this will change.

fultonj: Just trying to provide some history on past design plans: 
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/docs/architecture.md

controllers/openstackdataplanenode_controller.go
slagle: This needs to be rebased, and it would pick up the change where I've refactored all the deployment code into deployment.go, so that it can be used from both the node and role controllers. This includes the NovaExternalCompute handling.

docs/openstack_dataplanerole.md
slagle: What would the case be for exposing this deploy flag in the dataplane CRD's? We have deployStrategy.deploy on the dataplane CRD's already, can we just set NovaTemplate based on that value?

If there is a use case where we do need these set independently, then we need to account for the handling for when the values are not both true or both false, etc, given that all of the deployment logic is guarded by deployStrategy.deploy.

controllers/openstackdataplanenode_controller.go
slagle: We do need a way to enable/disable it. However, I think we need a common interface for this. What's proposed here wouldn't work for other dataplane services. The ctlplane actually uses an Enabled bool. I'd rather not add a 3rd way.

pkg/deployment/nova.go
slagle: I would use the CRDs as they currently exist as the basis for this patch. We did discuss driving all baremetal provisioning from the role. I don't see how this patch would contradict that? Other that my other inline replies, I'm mostly ok with what is proposed here.

@rabi linked to deployment.go where the common deployment code for NovaExternalComptue now lives, and I think after a rebase, this patch would be more inline with the current deployment patterns. Specifically, this commit:

https://github.com/openstack-k8s-operators/dataplane-operator/commit/e0d8cb1e4aa1b830c26c2236a7568b7fb200c349

The idea is to move any specific deployment logic into deployment.go. The function interfaces there work for a node, a role, or in the special case of NovaExternalCompute either a list of nodes (role deploy) or a list of one node (node deploy).

Once this is rebased, it may alleviate that specific concern.

#### Other comments about the CRD design:

I think what we have now is what we accept as the starting point of the current design. I wouldn't base patches on a different design that is not implemented. If at some point we rework or evolve the design, then of course all impacted code that is already committed would need to be updated.

OpenStackDataPlane allows you to create roles and nodes in a single YAML document, and the created roles/nodes will be owned by the top level DataPlane CR. DataPlane deployment is merely a loop over the roles where we deploy each role. There are no plans to implement a more specialized "DataPlane deployment" (I'm not even sure what that would look like). This is what has been meant when I have said "driving things from the role deployment". It's the role controller that is triggering the role deployment.

A user *can* edit a Role or Node which is created and owned by a DataPlane, however those changes are likely to be overwritten on subsequent reconciliations of DataPlane.

A user *can* create top level Roles and Nodes outside of any DataPlane. This does not necessarily mean they *should* at this point. I think we are still a little ways away from documenting our expected usage paths and best practices.

IMO, and I fully admit I am not k8s API expert, but the patterns in dataplane-operator match other patterns across the k8s API.

E.g., You can create a Deployment, and inline a pod in the YAML document, which creates an owned Pod. You can edit the owned Pod, but your changes would eventually get overwritten. You can also just create top level Pods with their own API and they are not owned by anything. While that's a useful mechanism, it doesn't mean that's how you ought to deploy Pods in production, or that it's a best practice on using k8s.

I tried to model the dataplane-operator CRD design and APIs after these and other patterns I saw in the k8s API (deployStrategy, etc). Of course, that doesn't mean that the k8s API is a perfect API either, and there are a lot of quirks there as well.

rabi: > I don't see how this patch would contradict that?

It would, as we should either do all reconciliation (provisioning/deployment) of nodes in role or the node controller. As per what you're saying provisioning would be role controller managed but deployment should be managed by both does not sound like a good design to me.

> @rabi linked to deployment.go where the common deployment code for NovaExternalComptue now lives, and I think after a rebase, this patch would be more inline with the current deployment patterns.

I think I'm still confused with the current design. I'm not really concerned about where the code lives, but the actual design around object ownership, reconciliation etc. Let me try and explain a few issues I see.

-  I see deployment.Deploy() is called both from the node and role controller. That means both role/node would try to own[1] (and one would fail) the same NovaExternalCompute.

-  Both role and node controller would try to reconcile NovaExternalCompute at the same time for any changes in it[2].

If we do all baremetal provisioning in role controller, with node CR reconciliation we assume the node has already been provisioned/pre-provisioned?  What's the significance of managed:true for node CRD then? Does that mean node controller has to check if a provisioned BMH exists and/or ssh  to the node before starting reconciliation of a node? 

 

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/nova.go#L60

[2] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanenode_controller.go#L267
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanerole_controller.go#L338

docs/openstack_dataplanerole.md
SeanMooney: adoption
we need to be able to have the dataplane operator create the NovaExternalComptue CR with deploy=false so that the generated config can be inspected without the ansibel being executed on the compute nodes.

if i understand correctly the current logic in the dataplane operator will not allow that to be created unless the deployStrategy.deploy=ture

pkg/deployment/nova.go
SeanMooney: there is __NOT__ general agreement on doing all the deployment from a role and I would not want to block this from progressing on that.

i don't belive that driving the deployment of all service at the role level is correct 

i would personally prefer if all reconsiltaiton was driven by the node controller or if we split between the two depending on the where the reosuce is defined.

i.e. where the nova template is defined on the role then it can be created and manage by the role controller and where its defined on the node it should be driven by the node controller.

in the current dataplane design roles are entirely optional.

that is why we have examples of using nodes standalone 
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml

roles are just a grouping mechanism for managing stes of datapalneNodes and propagating common settings.

they in turn allow you to execuat e ansible across multiple host via generating a merged inventory but that could be removed entirely and all execution could be drive at the node level if desired.

i don't think we want to exclusively drive things at the node level but we certenly never agree to only drive things at the role level.

to your questions regarding baremental provisioning.

if managed=false i.e. pre-provisoned hosts (which is expected to be our default for the dev preview) the AEE created by the node reconsiler will simply fail if the node is not provisioned. this would be reflected in the ready status and the admin would have to retriger the reconsile by modifying a filed on the CR or similar. i.e. update the hash in the status to for ce the reconsile to happen again.

we currently don't have any kind fo healtcheck of the node to know if its provisioned but
the expectation is that when using manged=false the OpenstackDataplaneNode CR is not created until the requirements for ansible to be able to connect have been fulfilled.

That is also a prerequisite for managed=true.
in that case the OpenstackDataplenNode CRs should not be created by the role controller until the baremetal operator has finsied provisioning the OS on the physical hosts.

so the precondition for creating an OpenStack dataplane Node is ensuring its provsioned with the required ssh public key added to the authorised keys file on ideally a non root use with passwordless sudo.

SeanMooney: @slagle im not convience we shoudl have OpenStackDataPlane CRD at all and i think we should only be creating the OpenstackDAtaplaneRoles and OpenstackDAtaplaneNode CRs

we add the top level OpenStackDataPlane CRD for not very good reason just to have a top level entity to model the over all datapalne but i don't actually think it add any value and in fact i think its very problematic for larger deployments with 1000 of compute nodes.


The design that gibi has implemented in this PR was the design we agreed to originally in the design doc we created for the interaction between service operator and the dataplane operator. its also what we discussed in brno. i.e. OpenStack services that are deployed on dataplane nodes are modelled by strongly typed service Temaplate the existing of which consoles the deployment of the service. 

This is partly why the AnibleVArs files was intentionally not part of fo the dataplne crd design and why i was so shocked by the fact you added it and called it out in the Q1 demo.
form my perspective groping everything at the role level and exposing the use of ansible for generic service config in the API provide dby the CRD were never agreed to as design goal before they were implemented in the dataplen operator.

i think embdeing 1000s of compute nodes in a single top-level CRD is an anti-pattern that we should not be relying on and i would be very concerned about the debuggability, scalability and user experience to taking that approach.


taking your deployment analagoy the nova template is basically a like a pod spec 
so we are proposing inline a spec at the node/role level to allow the create of the NovaExternalComptue CR. as with k8s pod they can be created in one of several ways

you can create free standing pods, or you can create pods that are part of a a higher level crd i.e. replicaset, deployment, statefulset, deamonset, job
each higher level construct adds additional semantic over the pod wich is the standard and minimum unit of scheduling in k8s.

you can create NovaExternalComptue CR as free-standing resources. for example if you were trying to deploy without the dataplene operator for some reason you just need to provide a valid inventory. i don't really expect that to be used out side of our development and testing but its something we support in the nova operator intentionally to make sure we have no depency on the datapalne operator outside fo the configmap that holds the inventory and the secret that provides the ssh key both of which you can create by hand.

you can also create it as managed by a higher-level resource.
in the current design we to entities that can own it it can either be owned by the OpenStackDataplaneNode (in my original proposal it was always going to be owned by this)
or the OpenstackDataplaneRole.

when we originally discussed the interaction between dataplane role and node i was expecting the role reconsiler to mealy mange creating nodes and the node reconsile would do 99% of the work.

that design evolved to have effectively the same functionality both at the role and node level however not without downsides.
the ability to report errors on a per host per task basis for the openstack dataplen role is much less debugable then when the same ansibel roels are executed purely at the openstackdataplennode level.

in the initial conversation we had on the design i was promoting reporting the status fo the deployment on any given node viat the status fo the openstackDataplaneNode status field.
while the conditions partly do that we don't actually have granular reporting fo the AEE progress there.

we also don't have that for the NovaExternalComptue to be fair but that's partly because the cannot get that info easily formteh AEE CR and i belive its wrong to just use embed the output of the underly in k8s job created by the AEE operator from the nova operator because we should not know about the internal fo how that operator works.
we should only use the public filed on the AEE CRD.

so I think the design that gibi has proposed is consistent with k8s API design in its current form and its also the design pattern we had previously agreed to follow for this integration.

SeanMooney:
> 

the initall feedback we provided on https://github.com/openstack-k8s-operators/dataplane-operator/pull/158
was partly motivated because we thought it was not a suitable interface to mange nova neutron or any OpenStack service  that should be typed and specific not untyped and generic

so no i don't think we should use https://github.com/openstack-k8s-operators/dataplane-operator/pull/158 for this.

[bshephar](https://github.com/bshephar)  yes you are missing context
we are not using https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova_compute/defaults/main.yml and we expiclty had a design go to not use any anisble variable to inclunc the deployment of the nova/libvirt services.

Ansible Vars were  __NOT__ intended to be present in any CRD interface and its why it was originally not included in the design document we created for the interaction.

we explicitly requested that the the edpm_nova_compute role not be imported form tripleo when the edpm ansible repo was created because upon reviewing the content of that role in September we determined that it was not deploying nova correctly and continued many design patterns and legacy functionality that did not align with how we intended to mange nova going forward.

not using any config from the ansible inventory was one fo the design goes we orgainlly had.
that is why i made sure that twe would not have any depency on the inventory bar the host connection info

this was meant to be how we models all services orgaily 

for better or worse the dataplane operator has diverged form that design and made ansible viable at the top level CRD. this is motivated by a desire to allow third-party ansible playbooks to be used.

That was originally not in scope of the Dataplane operator, that change at some point but the original proposal was vendors that need to implement configuration on dataplen nodes would one of the following
1.) execute standalone ansible directly
2.) reues the inventories and secrets form the dataplen operator and use those to directly use the AEE operator to run ther playbooks.
3.) write there own operator to mange the lifecycle of there integration

3 was the preferred approch but required the most investment form them.
1 is what i expected most to do.
2 has mostly been replaced by the functionality that james is building in 
https://github.com/openstack-k8s-operators/dataplane-operator/pull/158 however i strongly believe that should explicitly be used for third-party integration only and not for any code delivered by us. i think that is a far too leaky abstraction that has many fo the issues we had with THT templates in that we have exposed too much of the internals of how we deploy things like the ansible role names in the CR.

pkg/deployment/nova.go
rabi: > roles are just a grouping mechanism for managing stes of datapalneNodes and propagating common settings.

Are we saying role labeled nodes would be reconciled in role controller and the ones not labeled with role would be reconciled in node controller?  NovaExternalCompute is either owned by Role or Node? It can't be both right? What happens when a role is deleted? What happens to the labelled nodes? We would unprovision the nodes as the BaremetalSets are owned by the role. What would happen to the status of dataplane node CRDs that represent those nodes?

> so the precondition for creating an OpenStack dataplane Node is ensuring its provsioned with the required ssh public key added to the authorised keys file on ideally a non root use with passwordless sudo.

IMO, either we reconcile the node completely from nothing to provisioned/deployed node in node controller based on node spec, it can use shared config from the role (config) it's labelled with or we do everything in the role, probably not in both.

controllers/openstackdataplanenode_controller.go
gibizer: Yepp. It is on my TODO list to get this rebased and adapt to the changes on the main.

docs/openstack_dataplanerole.md
gibizer: Yepp. If we need to trigger the config generation of the nova-compute service on the EDPM node without deploying that config (i.e. inspection during adoption) then we need to let the dataplane-operator create NovaExternalCompute CRs with deploy=false.

I'm not sure how the config generation, and config inspection, during adoption will happen to the rest of the EDPM services though. So if there be a separate mechanism for that then we can consider adopting to it.

SeanMooney: the dataplen operator has not been adapted to generating all configs using go templates yet.

I'm not sure when that is planned for but as a result there currently isn't a good way to see what configs it will generate as that is unfortunately still done by Ansible.
ideally it would be adapted to also generate the configs via go templates and just copy them but its a lot of work and I'm not sure we have people to work on it.

pkg/deployment/nova.go
SeanMooney: if we are unprovisioning the physical servers because the baremental set was release when the role was delete i would also expect the OpenStackDataplaneNode CR for that physical server to be deleted.

assuming we want to allow OpenstackDataplaneRole CRs to be deleted if there exist an OpenstackDataplaneNode that refences them.

i kind fo think we should be using a finalise to prevent that.

i.e. the node controller should create a finalise for each node on the roles if a OpenstackDataplaneNode refences a roles.

i think we want to be more intentional especially in the managed=True case to make sure one typo does not nuke your dataplane hosts.

the workflow i think would make more sense would be to scalne the BaremetalSets to 0 before allowing the delete or as i said enforcing no OpenstackDataplaneNodes reference it.
both would have the same effect at the end fo the day.

but yes right now the  NovaExternalCompute should be owned either by the role or node but not both. if it was own by only one i think the  NovaExternalCompute is best aligned with the OpenstackDataplaneNode but we support both right now.

slagle:
Lots to unpack here. I did not agree to a design in Brno, or in a different doc. I recognize there were a couple of proposed designs captured in different docs, but that alone does not indicate agreement or it being the "original". Writing proposals into docs and expecting others to implement what you wrote isn't exactly how things typically work.

There were a couple of specific folks asked to work on dataplane-operator (myself primarily), and I proposed much of the design we have now. I'd be happy to explain those design choices further. This work was merged by others after consensus arrived at through team discussion (calls on Monday), or doc/issue comments.

My primary (only, I suppose) objection to this PR is the enablement of compute services based on the presence of NovaTemplate. I'd like to see a consistent interface for service enablement.

fultonj:
I agree with @slagle that "We did discuss driving all baremetal provisioning from the role. I don't see how this patch would contradict that"

I agree with @SeanMooney that "there is NOT general agreement on doing all the deployment from a role and I would not want to block this from progressing on that." I say this while baring in mind that "baremetal deployment" is a subset of "all deployment" so I'm not opposed to all baremetal provisioning being from a role.

I hear @rabi on his points around object ownership and reconciliation. I think that is something we could all talk about more but not in the context of this patch since that topic is bigger than this patch. If we're going to change that we'd need to change a lot of working code. I will say that the inheritance between roles and nodes does produce working deployments and doesn't fail to reconcile. Anyone can reproduce this by running the operators and creating nodes and roles. Maybe though an undesirable state can also be reached with that we have today. Perhaps it can be created in a lab and opened as a bug. So, I'm happy to meet to talk more about it, but I'd like to not block this review because of that. It's already an established pattern in this operator.

pkg/deployment/nova.go
slagle: > > I don't see how this patch would contradict that?
> 
> It would, as we should either do all reconciliation (provisioning/deployment) of nodes in role or the node controller. As per what you're saying provisioning would be role controller managed but deployment should be managed by both does not sound like a good design to me (with whatever little I know about k8s).

I don't agree it has to be all one way or another. There exist several patterns throughout the API of this exact thing, but I'll go back to Deployment/Pod. You can create a Pod, or you can create a Deployment which creates a Pod (or set of pods). In one case the Pods are owned by the Deployment, in the other they are not. There are best practices and recommendations for using either approach.

> 
> > @rabi linked to deployment.go where the common deployment code for NovaExternalComptue now lives, and I think after a rebase, this patch would be more inline with the current deployment patterns.
> 
> I think I'm still confused with the current design. I'm not really concerned about where the code lives, but the actual design around object ownership, reconciliation etc. Let me try and explain a few issues I see.
> 
>     * I see deployment.Deploy() is called both from the node and role controller. That means both role/node would try to own[1] (and one would fail) the same NovaExternalCompute.

I disagree that's possible with the code we have. Have you seen this happen? I'd consider that a bug if so. When you deploy a role, that's not's going to trigger the reconciliation loop for a node directly. While a node could be reconciled by some other factor, unless deploy=true, it won't call that function and create another NovaExternalCompute.

Of course it is possible create a series of malformed API requests that get things into a messed up state. That exists everywhere throughout the k8s API though. We should add checks to protect where we can (don't deploy a node if the role deploy is in progress, etc).

> 
>     * Both role and node controller would try to reconcile NovaExternalCompute at the same time for any changes in it[2].

No. That's not what that code means in [2]. The controller reconcile would only be triggered for the owner (either node or role). Again, someone could trigger both at the same time unintentionally, and we should protect against that.

> If we do all baremetal provisioning in role controller, with node CR reconciliation we assume the node has already been provisioned/pre-provisioned? What's the significance of managed:true for node CRD then? Does that mean node controller has to check if a provisioned BMH exists and/or ssh to the node before starting reconciliation of a node?

The managed field was added before baremetal support. As you're working on the baremetal support, I'd propose to remove it if you don't think it's needed. I'd agree it's not based on what you've proposed for baremetal support.

slagle:
> I agree with @SeanMooney that "there is NOT general agreement on doing all the deployment from a role and I would not want to block this from progressing on that." I say this while baring in mind that "baremetal deployment" is a subset of "all deployment" so I'm not opposed to all baremetal provisioning being from a role.

Right, I'd also agree there is not complete agreement. It was somewhat of a compromise, but relatively easy to implement, so I think it's a fair compromise to make. You can deploy either a role or node as top level. Again, I point out the numerous other patterns of this in the API.

I would be against removing the role deploy, as I am against more broad usage of the "one ansible execution per node" pattern. I don't think it will scale, and I would rather offload the handling of configuring multiple nodes mostly the same way to ansible. That's what ansible is good at and I'd suggest most users are used to interacting with that type of interface vs one run of ansible per one node.

fultonj:
In an effort to move things forward, I'd like to summarize what I think are the outstanding issues and a  plan to address them.

I don't think we should argue about Node vs Role vs General dataplane in this PR. If we want to talk about that, then it warrants a separate discussion. This PR is consistent with that pattern but so is the rest of the dataplane operator.

My summary from reading all of this is:

1. The PR exposes a nova specific template struct and a mechanism for it to be used.

2. There was some debate about template vs ansible vars, but I think that's now explained. At least if we're going to switch to using Ansible vars that's a major change in direction and perhaps warrants a separate discussion

3. There was a suggestion to change the mechanism by which compute services are enabled.

I think the next step is to address #3 and schedule a separate discussion about Node vs Role vs General dataplane.

Is there anything else we should list as needing to address. It would be nice to have simple list of outstanding issues since there has been a lot written. Sorry if I'm overlooking something so please add it concisely.

SeanMooney:
im not nessisarly against chanign how we enable the the deployment of nova to be something other then the present of the template but i am not sure if enabled is over loaded.

we can try doing that but for adoption we would have to do enabled=true deploy=false on the nova template since the default for enabled on the nova template would be false.

that feels a little more clunky but if that is the preference we could do that.

pkg/deployment/nova.go
rabi: >  You can create a Pod, or you can create a Deployment which creates a Pod (or set of pods)

Are we sure it's an apple-to-apple comparison? I'm not aware of a pod controller in k8s.  There are ReplicaSet and Deployment (an higher level abstraction) controllers.  What I can see is that deployment reconcile calls replicaset update/reconcile[1] and does not do things that replicaset has to do. There are also lots of checks/balances around ownership, claiming objects etc. Not sure if we would be able to go that level with our implementation here.

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/sync.go#L426

>   Have you seen this happen? I'd consider that a bug if so. 

Yes, I've seen that happen when deployStrategy/deploy is set true for node and role at the same time. Though we can have checks to ensure that they're not set to true at the same time.  I would guess those validations would be complex   when you have number of roles and nodes and node roles are toggled/removed. 

> The controller reconcile would only be triggered for the owner (either node or role).

That's right, they would only watch for events for the objects they own. What do we expect to happen when a role label (from spec) is removed from a node? With the current design role reconcile would tear down everything (including unprovisioning the node) as role owns all objects like NovaExternalCompute. Is that what we expect because  there is still a node object with (deployStrategy/deploy: false)?

> Right, I'd also agree there is not complete agreement. It was somewhat of a compromise, but relatively easy to implement, so I think it's a fair compromise to make.

I had the impression (after the baremetal provisioning calls) that we agreed to do everything with role reconciliation, but it seems I misunderstood. Also, I'm not sure we should do compromises because things are easy to implement. I've concerns around the current design and how we want to add baremetal provisioning to role reconcile, but it seems not everyone is bothered the same way. I rest my case.

slagle: > > You can create a Pod, or you can create a Deployment which creates a Pod (or set of pods)
> 
> Are we sure it's an apple-to-apple comparison? I'm not aware of a pod controller in k8s. There are ReplicaSet and Deployment (an higher level abstraction) controllers. What I can see is that deployment reconcile calls replicaset update/reconcile[1] and does not do things that replicaset has to do. There are also lots of checks/balances around ownership, claiming objects etc. Not sure if we would be able to go that level with our implementation here.

Yes, you absolutely can create a Pod on its own. https://kubernetes.io/docs/concepts/workloads/pods/

It's not apples to apples. It's a comparison of an API pattern.

> 
> [1] https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/sync.go#L426
> 
> > Have you seen this happen? I'd consider that a bug if so.
> 
> Yes, I've seen that happen when deployStrategy/deploy is set true for node and role at the same time. Though we can have checks to ensure that they're not set to true at the same time. I would guess those validations would be complex when you have number of roles and nodes and node roles are toggled/removed.

I'll add a check for that.

> 
> > The controller reconcile would only be triggered for the owner (either node or role).
> 
> That's right, they would only watch for events for the objects they own. What do we expect to happen when a role label (from spec) is removed from a node? With the current design role reconcile would tear down everything (including unprovisioning the node) as role owns all objects like NovaExternalCompute. Is that what we expect because there is still a node object with (deployStrategy/deploy: false)?

We don't remove role labels from nodes. If you delete the role, it deletes everything it owns, including the node.

> > Right, I'd also agree there is not complete agreement. It was somewhat of a compromise, but relatively easy to implement, so I think it's a fair compromise to make.
> 
> I had the impression (after the baremetal provisioning calls) that we agreed to do everything with role reconciliation, but it seems I misunderstood. Also, I'm not sure we should do compromises because things are easy to implement. I've concerns around the current design and how we want to add baremetal provisioning to role reconcile, but it seems not everyone is bothered the same way. I rest my case.

All of the deployment is driven from the role reconciliation. It's the role controller that (eventually) creates the OpenStackAnsibleEE CR's. Even for NovaExternalCompute, we deploy the roles, which deploys NovaExternalCompute for the nodes one by one, but the parent code is the role controller. I'm not sure what the conflict is here, if you are just objecting to where the go code lives in what package, or where it's called from, etc.

rabi: >  We don't remove role labels from nodes. If you delete the role, it deletes everything it owns, including the node.

Are you sure we delete the node? I see that the node is owned by dataplane rather than dataplanerole[1].

> if you are just objecting to where the go code lives in what package

As I mentioned earlier I don't care where the code lives and how they're re-used, my concern is with the current design (i.e dependencies, ownership and how and when we reconcile), but let's move on. I've updated the baremetal provisioning PR to just hook baremetal provisioning in role reconcile[2].

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplane_controller.go#L272
[2] https://github.com/openstack-k8s-operators/dataplane-operator/pull/165

slagle: > > We don't remove role labels from nodes. If you delete the role, it deletes everything it owns, including the node.
> 
> Are you sure we delete the node? I see that the node is owned by dataplane rather than dataplanerole[1].

If created by the DataPlane, the node will be deleted when the DataPlane is deleted. If you delete the node on your own, it'd get re-created when the DataPlane is reconciled.

If you create Node and Role CR's outside of any DataPlane, and then delete one that needs the other, then you're going to get an undefined behavior, for which we need to add checks and error handling.

This is the same as if you have a Deployment with Volumes that reference a ConfigMap or Secret. If you then go in and delete that ConfigMap or Secret, you get an error, the next time the Deployment is reconciled you get an error.
 
> > if you are just objecting to where the go code lives in what package
> 
> As I mentioned earlier I don't care where the code lives and how they're re-used, my concern is with the current design (i.e dependencies, ownership and how and when we reconcile), but let's move on. I've updated the baremetal provisioning PR to just hook baremetal provisioning in role reconcile[2].

Please feel open to influence the current design to address your concerns. I'm happy to discuss any proposals or test out any PR's/prototypes.

bogdando: Splitting provision/config/deploy responsibilities over *ExternalCompute CRDs across multiple Node/Role/Datplane controllers could be indeed hard to implement, test, and maintain. We're not as much experienced as Deployment and ReplicaSet controllers authours. Given that, and added the recent "10,000 namespaces" scale testing feedback on etcd and k8s apiserver limitations, it might be architecture wise to ditch dataplane CRD (and even node CRDs) and scale nodes as replica counts of roles and its controllers.

SeanMooney: if we did that i dont think we would have a way to adopt osp 17.1 deployment since we could not model pre provisioned host anymore so i don't thinkthat will work unless we added a way to list pre-provision host in the role directly

basically a list of ansible connection triples (user, host, port)

docs/openstack_dataplanenode.md
gibizer: I added the command to the doc to query the names of the Nova CRs in the system.
If we always deploy via openstack-operator then the default value "nova" is good as openstack-operator hardcodes that as a name of the Nova CR it creates.

controllers/openstackdataplanenode_controller.go
gibizer: It is rebased now.

gibizer:
> > I did manual testing to show that config propagation works.
> > 
> > So compute-0 should get the customService config from the Node while compute-1 should get it from the Role:
> > 
> > ```
> > 2. Define no novaTemplate in the role and only define it for one of the Nodes:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > So the compute-0 is deployed but compute-1 NovaExternalCompute is not created so it is not deployed. I.e. compute-1 is not a compute node.
> 
> The config propagation is the right pattern IMO, but I don't like that the inclusion of compute services is keyed off of the presence of NovaTemplate. What about defaults? Is the user forced to reproduce our default values just to force the deployment of compute services?
> 
> While the backend implementations are different, could we define a common interface for service enablement/disablement? I'm not sure what that would look like, but one idea is that we could list nova-external-compute in the composable services list I proposed in #158. The code could check for that service name and guard the NovaExternalCompute deployment based on its presence.
> 
> Or, we could define a different interface that works for both. I think the user facing interface is the important bit here, and not necessarily that these are implemented differently.

gibizer:
So this is rebased now and I did manual testing locally. This works but I believe we have design issues to settle about how to allow defining an EDPM node without a nova-compute. 

The current patch proposes a pointer NovaTemplate in the NodeSection. If that is nil both in the Node and in the Role then no NovaExternalCompute is created and therefore no nova-compute is deployed.

@slagle had a comment about default:
>  What about defaults? Is the user forced to reproduce our default values just to force the deployment of compute services? 

I don't see issues with the default. The following yaml fragment will enable NovaExternalCompute with the default values defined in the NovaTemplate
```
roles:
      nodeTemplate:
        nova: {}
```
 So the user is not forced to know our default values.

Also from @slagle
> While the backend implementations are different, could we define a common interface for service enablement/disablement? I'm not sure what that would look like, but one idea is that we could list nova-external-compute in the composable services list I proposed in https://github.com/openstack-k8s-operators/dataplane-operator/pull/158. The code could check for that service name and guard the NovaExternalCompute deployment based on its presence.

I can look into additionally guarding NovaExternalCompute creation based on the composable services list. I.e. OpenstackDataPlaneRole.Spec.Services. However I still think that keeping the nullable filed is a good UX as it means the user does not need to define a structure (or see a default value of a structure) about nova-compute service when the deployment of that service is disabled.

gibizer:
@SeanMooney in the last version I implemented the field by field merging off NovaTemplate. So if the Role has
```
role:
    nodeTemplate:
        nova:
            CellName: cell1
```
Then the Node can still have a node specific CustomServiceConfig with:
```
node:
    nova:
        CustomServiceConfig: |
        [DEFAULT]
        debug = True
```

And the resulting NovaExternalCompute will have a merged template with cell1 from the Role and debug=True from the Node.

slagle:
> So this is rebased now and I did manual testing locally. This works but I believe we have design issues to settle about how to allow defining an EDPM node without a nova-compute.
> 
> The current patch proposes a pointer NovaTemplate in the NodeSection. If that is nil both in the Node and in the Role then no NovaExternalCompute is created and therefore no nova-compute is deployed.
> 
> @slagle had a comment about default:
> 
> > What about defaults? Is the user forced to reproduce our default values just to force the deployment of compute services?
> 
> I don't see issues with the default. The following yaml fragment will enable NovaExternalCompute with the default values defined in the NovaTemplate
> 
> ```
> roles:
>       nodeTemplate:
>         nova: {}
> ```
> 
> So the user is not forced to know our default values.

Ok, thanks for confirming.

> 
> Also from @slagle
> 
> > While the backend implementations are different, could we define a common interface for service enablement/disablement? I'm not sure what that would look like, but one idea is that we could list nova-external-compute in the composable services list I proposed in #158. The code could check for that service name and guard the NovaExternalCompute deployment based on its presence.
> 
> I can look into additionally guarding NovaExternalCompute creation based on the composable services list. I.e. OpenstackDataPlaneRole.Spec.Services. However I still think that keeping the nullable filed is a good UX as it means the user does not need to define a structure (or see a default value of a structure) about nova-compute service when the deployment of that service is disabled.

Given the ongoing discussion, we agreed to merge this as-is with the current design in place.

slagle (approved):

PR #165, opened 2023-04-25
Add baremetal provisioning support

rabi:
This adds baremetal provisioning support using openstack-baremetal-operator.


controllers/openstackdataplanerole_controller.go
```diff
@@ -313,6 +325,44 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+func (r *OpenStackDataPlaneRoleReconciler) ReconcileBaremetalSet(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneRole, helper *helper.Helper,
+) error {
+	baremetalSet := &baremetalv1.OpenStackBaremetalSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	helper.GetLogger().Info("Reconciling BaremetalSet")
```
bshephar, line 338: We can use `r.Log.Info()` here. The logger is initialized as part of the `OpenStackDataPlaneRoleReconciler` struct, so we can just use that instead of recalling `helper.GetLogger()`.

controllers/openstackdataplane_controller.go
```diff
@@ -283,18 +308,21 @@ func (r *OpenStackDataPlaneReconciler) CreateDataPlaneNode(ctx context.Context,
...
 
-// CreateDataPlaneRole -
-func (r *OpenStackDataPlaneReconciler) CreateDataPlaneRole(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlane, helper *helper.Helper) error {
-
+// CreateOrPatchDataPlaneRoles Create or Patch DataPlaneRole
+func CreateOrPatchDataPlaneRoles(ctx context.Context,
+	instance *dataplanev1beta1.OpenStackDataPlane, helper *helper.Helper,
+	roleManagedHostMap map[string]map[string]string) error {
+	client := helper.GetClient()
+	logger := helper.GetLogger()
```
bshephar, line 316: Curious to understand why we're changing this from a receiver function? That's what allows us to do `r.Log` and `r.Client` without reinitializing those objects here with `helper.Get*()`.  

There might be some technical reason that I'm not considering, but it could also be that the functions were only recently changed. Maybe the change here is unintentional, so asking just in case.

bshephar (commented):
Seem to be changing all of the functions here from receiver functions. So I just had a query to understand if that was intentional.

controllers/openstackdataplane_controller.go
rabi: Hmm.. I guess you're aware that I've been working on this patch before your changes merged and this was rebased automatically.

Also, I mentioned that we either use the helper or use public properties from reconciler using the receiver. There is no point doing both. You can use r.Scheme rather

controllers/openstackdataplanenode_controller.go
```diff
@@ -114,30 +103,11 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
-	if err != nil {
-		return result, err
-	}
-
 	// Always patch the instance status when exiting this function so we can
 	// persist any changes.
 	defer func() {
 		// update the overall status condition if service is ready
-		if instance.IsReady() || instanceRole.IsReady() {
+		if instance.IsReady() {
```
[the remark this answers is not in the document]
rabi: As we're not reconciling the node separately, we probably don't need to. We can keep each node provisioning/deployment status in role rather than in the respective node CRDs. We can also update the node status from the role controller, but that won't be a good design. 

I'm working on  a patch to simplify the status/conditions  which I'll add to the PR.

fao89 (commented):
LGTM

+2

rabi:
> @rabi could you please teach me how to test it?

There is a make target proposed in https://github.com/openstack-k8s-operators/install_yamls/pull/252 to deploy it from scratch with crc. It will get the nodes to provisioned state. Further deployment can be done by setting deploy:true for the openstackdataplane CR. I guess we should get this PR merged soon to proceed with further work.

hjensas (approved):
Looks ok to me, I think we can merge this and iterate.

PR #171, opened 2023-04-27
Add PreNetwork roles/tasks to ConfigureNetwork

rabi:
We need to do these before nerwork configuration.

### DIFF
--- a/pkg/deployment/network.go
+++ b/pkg/deployment/network.go
@@ -32,6 +32,36 @@ func ConfigureNetwork(ctx context.Context, helper *helper.Helper, obj client.Obj
 
 	tasks := []dataplaneutil.Task{
 		{
+			Name:          "Install edpm_bootstrap",
+			RoleName:      "osp.edpm.edpm_bootstrap",
+			RoleTasksFrom: "bootstrap.yml",
+			Tags:          []string{"edpm_bootstrap"},
+		},
+		{
+			Name:          "Grow volumes",
+			RoleName:      "osp.edpm.edpm_growvols",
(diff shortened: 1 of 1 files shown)


pkg/deployment/network.go
```diff
@@ -31,6 +31,36 @@ import (
 func ConfigureNetwork(ctx context.Context, helper *helper.Helper, obj client.Object, sshKeySecret string, inventoryConfigMap string, aeeSpec dataplanev1beta1.AnsibleEESpec) error {
```
hjensas, line 31: These tasks are not network related - should we rename the file and function or split it out to to a separate (bootstrap.go)? file?

hjensas (commented):
Looks good, but should we mix bootsrap/storage/kernel args into "ConfigureNetwork"?

pkg/deployment/network.go
rabi: Yeah, we should/would do that. However, @slagle told me that there would be major refactoring as part of composable services work.  Therefore, I just added the pre-network[1] edpm tasks here, so that we can have those for the time being.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/deploy_edpm_pre_network.yml

hjensas: OK, makes sense. I will mark this resolved.

fao89 (approved):

rabi:
Sorry there was a copypasta. I noticed during testing. Proposed fix https://github.com/openstack-k8s-operators/dataplane-operator/pull/172

PR #174, opened 2023-05-04
Add `deployIdentifier` field in `deployStrategy`

rebtoor:
Add a `deployIdentifier` field in `deployStrategy` when `deploy` is `true`

`deployIdentifier`, if not defined by the user, will be a random UUID.


controllers/openstackdataplanenode_controller.go
```diff
@@ -218,9 +218,27 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 
 	r.Log.Info("Node", "DeployStrategy", instance.Spec.DeployStrategy.Deploy, "Node.Namespace", instance.Namespace, "Node.Name", instance.Name)
 	if instance.Spec.DeployStrategy.Deploy {
+
+		// create deployIdentifier
+		if len(instance.Spec.DeployStrategy.DeployIdentifier) == 0 {
+			_, err = controllerutil.CreateOrUpdate(ctx, r.Client, instance, func() error {
+				newIdentifier := dataplanev1beta1.GenerateDeployIdentifier()
+				instance.Spec.DeployStrategy.DeployIdentifier = newIdentifier
```
slagle, line 226: This needs to only generate the DeployIdentifier if the Deploy is toggled to true. Meaning, we only want to generate a unique value when the node is requested to be deployed by the user. Not on every future reconcile when Deploy is true. You can compare the old instance.Spec and the new to figure out if the field has been set to true and it was previously false.

This also does not need to be set inside controllerutil.CreateOrUpdate. It will get saved to when the instance is saved in the defer func.

controllers/openstackdataplanerole_controller.go
```diff
@@ -211,6 +211,29 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+		// 	if err != nil {
+		// 		return ctrl.Result{}, err
+		// 	}
+		// 	r.Log.Info("DeployIdentifier updated to: ", "identifier", newIdentifier)
+		// }
+		if len(instance.Spec.DeployStrategy.DeployIdentifier) == 0 {
+			_, err = controllerutil.CreateOrUpdate(ctx, r.Client, instance, func() error {
+				newIdentifier := dataplanev1beta1.GenerateDeployIdentifier()
+				instance.Spec.DeployStrategy.DeployIdentifier = newIdentifier
+				r.Log.Info("DeployIdentifier updated to: ", "identifier", instance.Spec.DeployStrategy.DeployIdentifier)
```
slagle, line 228: My same comments from the node controller apply here as well.

api/v1beta1/common.go
```diff
@@ -91,6 +92,11 @@ type DeployStrategySection struct {
 	// +kubebuilder:validation:Optional
 	// AnsibleSkipTags for ansible execution
 	AnsibleSkipTags string `json:"ansibleSkipTags,omitempty"`
+
+	// DeployIdentifier is a randomly created string created for new deploy
+	// it can be used to override a current deploy on ansibleee objects
```
slagle, line 97: I would phrase this as:

```
// DeployIdentifier is a generated UUID set as input on OpenStackAnsibleEE resources
// so that the OpenStackAnsibleEE controller can determine job input uniqueness.
// It is generated on each new deploy request (when DeployStrategy.Deploy is changed to true).
```

or something along those lines :smile: Feel free to wordsmith.

slagle (commented):
This is looking good. I think it's on the right track. I did not get a chance to try it yet, but I will tomorrow.

api/v1beta1/openstackdataplanenode_types.go
```diff
@@ -144,5 +144,10 @@ func (instance OpenStackDataPlaneNode) GetAnsibleEESpec(role OpenStackDataPlaneR
 	} else {
 		aee.OpenStackAnsibleEERunnerImage = role.Spec.OpenStackAnsibleEERunnerImage
 	}
+	if len(instance.Spec.DeployStrategy.DeployIdentifier) > 0 {
+		aee.DeployIdentifier = instance.Spec.DeployStrategy.DeployIdentifier
+	} else {
+		aee.DeployIdentifier = role.Spec.DeployStrategy.DeployIdentifier
+	}
```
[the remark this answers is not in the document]
rebtoor: ack!

rebtoor:
@slagle i found a workaround for the `deployStrategy.deployIdentifier` definition and now is correctly defined in the CR when `deployStrategy.deploy = true`.
BTW, as pointed in chat, when we try to force `deployStrategy.deploy = false` and `deployStrategy.deployIdentifier = ""` at the end of the deploy, it won't work since the `PatchInstance` method in the defer func, doesn't correctly patch the CR because it evaulates only changes in `metadata` and `status` field. Adopting the same approach as the `deployIdentifier` definition doesn't work as well because it cause an endless loop in the reconcile since the values continue to flap.

Any suggestion is very welcome :sweat_smile:

controllers/openstackdataplanenode_controller.go
```diff
@@ -225,12 +243,15 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
-
+		// Explicitly set instance.Status.Deploy = false
+		// Explicitly set instance.Status.DeployIdentifier = ""
+		// We don't want another deploy triggered by any reconcile request, it should
+		// only be triggered when the user (or another controller) specifically
+		// sets it to true.
+		r.Log.Info("Set Deploy to false")
+		instance.Status.Deploy = false
+		r.Log.Info("Set DeployIdentifier to empty value")
+		instance.Status.DeployIdentifier = ""
```
slagle, line 254: is the switch to setting these on Status since we couldn't figure out how to update the Spec? I guess my concern with this is that Spec.DeployStrategy.Deploy would still be true after a deploy.

Then the status fields are set to false. So on the next reconcile wouldn't it start a new deploy?

To edit the fields on the spec can we use Update as seen here?
https://cloud.redhat.com/blog/kubernetes-operators-best-practices
(search for r.GetClient().Update)

rebtoor: You're right, it seems that it was working but it keep updating ansibleee CRs even if their done. I'll try again with `r.GetClient().Update` approach.

bshephar:
> @slagle i found a workaround for the `deployStrategy.deployIdentifier` definition and now is correctly defined in the CR when `deployStrategy.deploy = true`. BTW, as pointed in chat, when we try to force `deployStrategy.deploy = false` and `deployStrategy.deployIdentifier = ""` at the end of the deploy, it won't work since the `PatchInstance` method in the defer func, doesn't correctly patch the CR because it evaulates only changes in `metadata` and `status` field. Adopting the same approach as the `deployIdentifier` definition doesn't work as well because it cause an endless loop in the reconcile since the values continue to flap.
> 
> Any suggestion is very welcome


I think we discussed already. But to refresh my memory, can we just use `deployIdentifier` in the `status`? If the user says `deployStrategy.deploy = False`, then we don't update the deploy identifier in the status. What was the use case for having a user manually define a deployIdentifier in the spec?

rebtoor:
> I think we discussed already. But to refresh my memory, can we just use `deployIdentifier` in the `status`? If the user says `deployStrategy.deploy = False`, then we don't update the deploy identifier in the status. What was the use case for having a user manually define a deployIdentifier in the spec?

The user wants to define the scope of a deploy to force a new one.

bshephar:
> > I think we discussed already. But to refresh my memory, can we just use `deployIdentifier` in the `status`? If the user says `deployStrategy.deploy = False`, then we don't update the deploy identifier in the status. What was the use case for having a user manually define a deployIdentifier in the spec?
> 
> The user wants to define the scope of a deploy to force a new one.

Ok, so in that case I wonder if just setting the deployIdentifier in the spec back to "" would be sufficient to kick it off again. We could check the status during the reconcile loop, if the `deployIdentifier == nil` then we just recreate the jobs and run it again?

Save putting the logic in the Spec which seems to be the main source of conflict for implementing this?

controllers/openstackdataplanenode_controller.go
```diff
@@ -199,12 +199,27 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 
 	r.Log.Info("Node", "DeployStrategy", instance.Spec.DeployStrategy.Deploy, "Node.Namespace", instance.Namespace, "Node.Name", instance.Name)
 	if instance.Spec.DeployStrategy.Deploy {
+
+		// create deployIdentifier
+		if len(instance.Spec.DeployStrategy.DeployIdentifier) == 0 {
+			// patch := client.MergeFrom(instance.DeepCopy())
+			newIdentifier := dataplanev1beta1.GenerateDeployIdentifier()
+			instance.Spec.DeployStrategy.DeployIdentifier = newIdentifier
```
bshephar, line 207: This is why it keeps reconciling, because we update the spec here. Each time we reach here, we call `GenerateDeployIdentifier()` which gives us a new unique string, and causes us to reconcile again. Let's update the status instead with the current identifier. I think we need some additional logic here to protect against generating a new string each reconcile loop, since the operator will reconcile by default every 10 minutes, the current implementation would result in a new identifier every 10 minutes, potentially triggering an action.

It probably makes sense to have a number that increments rather than a randomly generated string? So if we update this to first check the current deploy identifier, if it's "", then we need to generate one. If it ISN'T "", then we should increment it if required. So, if the instance Spec has changed, then we need to increment it for example.

bshephar: In this if statement here, we're checking to see IF the user has provided us a custom DeployIdentifier, if they have NOT, then we are generating a new one. Expanding on my above point, I think we should only do this if one doesn't exist in the Status as well. So the options here are,

1. Instead of a string, we could return a int64 from the `GenerateDeployIdentifier()` method. Then we could increment that integer for each subsequent deploy. (But maybe we don't care to be that granular, and all we really want is uniqueness? If so, see option 2)

2. We simply change this to: `instance.Status.DeployIdentifier = newIdentifier`. Although, it would probably be more efficient to just do:

```
instance.Status.DeployIdentifier = dataplanev1beta1.GenerateDeployIdentifier()
```

Then we save space on the variable allocation. 

In cases where a use HAS provided a DeployIdentifier, then we want to introduce a new method with logic that will determine what action we should take in that case. So, I think what we would want to do is to set the `instance.Status.DeployIdentifier == instance.Spec.DeployStrategy.DeployIdentifier`. 

Since the deploy Identifier has changed, we then need to take action based upon that outcome and recreate the jobs.

controllers/openstackdataplanenode_controller.go
```diff
@@ -226,10 +241,18 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
 
 		// Explicitly set instance.Spec.Deploy = false
+		// Explicitly set instance.Spec.DeployStrategy.DeployIdentifier = ""
 		// We don't want another deploy triggered by any reconcile request, it
 		// should only be triggered when the user (or another controller)
 		// specifically sets it to true.
+		// patch := client.MergeFrom(instance.DeepCopy())
+		r.Log.Info("Set DeployStrategy.Deploy to false")
 		instance.Spec.DeployStrategy.Deploy = false
+		instance.Spec.DeployStrategy.DeployIdentifier = ""
```
bshephar, line 251: This ties into my comments above. We end up setting a new string for the deployIdentifier above, then reset it to "" here?. So when it gets back to the above if statement, it's `== 0` and off we go again.

We will need to make sure we maintain logic that will prevent another rollout when we update this. The controller will still reconcile when we do this, so we'll need to have the logic in place to handle it.

controllers/openstackdataplanerole_controller.go
```diff
@@ -209,6 +209,19 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 
 	r.Log.Info("Role", "DeployStrategy", instance.Spec.DeployStrategy.Deploy, "Role.Namespace", instance.Namespace, "Role.Name", instance.Name)
 	if instance.Spec.DeployStrategy.Deploy {
+
+		// create deployIdentifier
+		if len(instance.Spec.DeployStrategy.DeployIdentifier) == 0 {
+			// patch := client.MergeFrom(instance.DeepCopy())
+			newIdentifier := dataplanev1beta1.GenerateDeployIdentifier()
+			instance.Spec.DeployStrategy.DeployIdentifier = newIdentifier
```
bshephar, line 217: Same as above, let's update the Status here rather than the Spec. We can a different path in situations where the user provides a string.

kajinamit:
Re-reading the proposal I'm confused with what we want to achieve by this change. The aim of deploymentIdentifier proposed here is different from the one in TripleO. In TripleO the identifier was generated in every deployment run to ensure specific tasks are executed every time. We probably have to step back and understand use case of this feature more in detail.

So currently we have `instance.Spec.DeployStrategy.Deploy` to control whether the task should be executed or not. This proposed adding a different field `instance.Spec.DeployStrategy.DeployIdentifier` to control a similar behavior.

With this change the client should set the identifier field properly to determine whether the job should be executed or not. Then that means the client can control whether the Deploy field is updated and I don't see much benefit of this feature.

If we want the dataplane operator to decide whether the task should be executed based on input then we probably should generate the hash from input instead of using a ramdom string.

controllers/openstackdataplanenode_controller.go
slagle: > This is why it keeps reconciling, because we update the spec here. Each time we reach here, we call `GenerateDeployIdentifier()` which gives us a new unique string, and causes us to reconcile again. Let's update the status instead with the current identifier. I think we need some additional logic here to protect against generating a new string each reconcile loop, since the operator will reconcile by default every 10 minutes, the current implementation would result in a new identifier every 10 minutes, potentially triggering an action.

The GenerateDeployIdentifier is only reached if the current length on the spec is 0, so I don't think it would continually reconcile past the first time.
 
> It probably makes sense to have a number that increments rather than a randomly generated string? So if we update this to first check the current deploy identifier, if it's "", then we need to generate one. If it ISN'T "", then we should increment it if required. So, if the instance Spec has changed, then we need to increment it for example.

What would be the difference between a random/unique value and an incrementing value?

I agree that setting instance.Status.DeployIdentifier would be useful, and we could use that to track if the deployment "in progress" is the one started for the requested deploy.

controllers/openstackdataplanenode_controller.go
slagle: This line is never reached until the deployment is done. If a deployment is in progress, we've returned earlier. By the time we reach here, the deployment is done and we reset Deploy and DeployIdentifier back to the non-deploying values. On the next (auto)reconcile a deploy is not started since Deploy=false.

controllers/openstackdataplanerole_controller.go
slagle: I think we would want the value on the Spec to equal the value in the Status to indicate that's the deployment in progress wouldn't we? Otherwise, we would keep hitting these lines since the Spec value would always length of 0.

controllers/openstackdataplanenode_controller.go
rebtoor: As I said earlier, what @slagle says here is correct. Those values are set to their default only when the deploy is done.

Again, the fact that it starts from scratch after some minutes (?) it doesn't caused by this (IMHO).

controllers/openstackdataplanerole_controller.go
rebtoor: `Spec` shouldn't reflect the value of the `Status` field in my opinion but vice versa.

`Spec.deployIdentifier` should be what we want -> if it's defined in the CR, we use it, otherwise it will be created as a random UUID.

`Status.deployIdentifier` should be valued with the value of `Spec.deployIdentifier` currently running.

My doubts here are:

* If the deploy is completed the `Status.deployIdentifier` field should be resetted to `""` just like `Spec.deployIdentifier`?
* If there's a mismatch between those 2 fields, what should happen? At the moment `.Spec.deployIdentifier` is passed down to AnsibleEE CR as label, and due to https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/145 , if it changes in the AnsibleEE CR, it will force a job recreation (after its/their completion) because the hash is changed. So we could:
  * overwrite the `Status.deployIdentifier` field with the content of `Spec.deployIdentifier` logging to the user that there's an existing deploy that will be overwritten
  * ignore(?) the new CR throwing an error that warns the user that there is a deployment in progress

What do you think folks?

rebtoor:
> Re-reading the proposal I'm confused with what we want to achieve by this change. The aim of deploymentIdentifier proposed here is different from the one in TripleO. In TripleO the identifier was generated in every deployment run to ensure specific tasks are executed every time. We probably have to step back and understand use case of this feature more in detail.

I don't think we aim to replicate the TripleO behavior but I'll left the final word to @slagle :)

> So currently we have `instance.Spec.DeployStrategy.Deploy` to control whether the task should be executed or not. This proposed adding a different field `instance.Spec.DeployStrategy.DeployIdentifier` to control a similar behavior.
> 
> With this change the client should set the identifier field properly to determine whether the deployment should be executed or not. Then that means the client can control whether the Deploy field should be updated and I don't see much benefit of this feature.

`instance.Spec.DeployStrategy.Deploy` will control the deploy execution as a boolean.

`instance.Spec.DeployStrategy.DeployIdentifier` will not control the deploy execution, it will help the user to identify a deployment. If it's provided in the CR by the user, that value will be used, otherwise a UUID will be generated. That value will be passed to the AnsibleEE CR as a label (and the ansibleee-operator will pass the value to the job)

> If we want the dataplane operator to decide whether the task should be executed based on input then we probably should generate the hash from input instead of using a ramdom string. But this can be implemented adding the status field which stores the hash generated from the content and compare the generated hash at the beginning of reconciliation with the recorded value.

The problem that we want to solve here is explained by James on Jira issue, I'll quote him here:

> [...] This gives the operator a way to tell if a found instance of the CR is actually equivalent to the deployment being requested. For example, an older NovaExternalCompute may still be in progress (stale), and the dataplane-operator has no way of knowing if that was triggered by the current deployment request or an older one. Using a hash, the operator could make this distinction and know if it needs to create a new NovaExternalCompute with the new input values or wait on the existing one to finish. 

Part of this (ansibleee crs deploymentIdentifier handling, hash of input fields hashes, job recreation when hash change) is solved by https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/145 so here we'd like to solve the dataplane cr part.
If a Dataplane CR already exists with same parameters, how we "force" the the creation of a new AnsibleEE CRs (and, consenquently, new jobs)? Anyway not everything is clear to me too, so I asked a couple of questions above.

I hope I added a little more context. :sweat_smile:

controllers/openstackdataplanenode_controller.go
bshephar: > This line is never reached until the deployment is done. If a deployment is in progress, we've returned earlier. By the time we reach here, the deployment is done and we reset Deploy and DeployIdentifier back to the non-deploying values. On the next (auto)reconcile a deploy is not started since Deploy=false.

It definitely is:
```
{"level":"info","ts":"2023-05-31T20:52:59+10:00","msg":"Adding label openstackdataplanerole=edpm-compute","controller":"openstackdataplanenode","controllerGroup":"dataplane.openstack.org","controllerKind":"OpenStackDataPlaneNode","OpenStackDataPlaneNode":{"name":"edpm-compute-0","namespace":"openstack"},"namespace":"openstack","name":"edpm-compute-0","reconcileID":"9dc6c525-9c9c-4df8-a424-cb705cb39c1e"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","msg":"Node","controller":"openstackdataplanenode","controllerGroup":"dataplane.openstack.org","controllerKind":"OpenStackDataPlaneNode","OpenStackDataPlaneNode":{"name":"edpm-compute-0","namespace":"openstack"},"namespace":"openstack","name":"edpm-compute-0","reconcileID":"9dc6c525-9c9c-4df8-a424-cb705cb39c1e","DeployStrategy":true,"Node.Namespace":"openstack","Node.Name":"edpm-compute-0"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","logger":"controllers.OpenStackDataPlane","msg":"Create Or Patch DataPlaneRole","roleName":"edpm-compute"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","logger":"controllers.OpenStackDataPlane","msg":"CreateDataPlaneNode","nodeName":"edpm-compute-0"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","logger":"controllers.OpenStackDataPlane","msg":"Create Or Patch DataPlaneRole","roleName":"edpm-compute"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","logger":"controllers.OpenStackDataPlane","msg":"Set DeployStrategy.Deploy to false"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","msg":"DeployIdentifier updated to: ","controller":"openstackdataplanenode","controllerGroup":"dataplane.openstack.org","controllerKind":"OpenStackDataPlaneNode","OpenStackDataPlaneNode":{"name":"edpm-compute-0","namespace":"openstack"},"namespace":"openstack","name":"edpm-compute-0","reconcileID":"9dc6c525-9c9c-4df8-a424-cb705cb39c1e","identifier":"672c04ae-9681-48fa-b7de-96bfb939b03b"}
{"level":"info","ts":"2023-05-31T20:52:59+10:00","msg":"Starting DataPlaneNode deploy","controller":"openstackdataplanenode","controllerGroup":"dataplane.openstack.org","controllerKind":"OpenStackDataPlaneNode","OpenStackDataPlaneNode":{"name":"edpm-compute-0","namespace":"openstack"},"namespace":"openstack","name":"edpm-compute-0","reconcileID":"9dc6c525-9c9c-4df8-a424-cb705cb39c1e"}
```

We can see there that it clears the `deployIdentifer` and starts a new reconcile since the reconcileID has changed. This loops on repeat indefinitely. So we start reconciling, then we update the Spec with the `deployIdetifier` UUID, we update the Spec with the empty value for `deployIdentifier`, then we start again.

If you `make run`, create a dataplanenode and then just run this a few times in succession:
```
❯ oc get openstackdataplanenode edpm-compute-0 -o jsonpath='{.spec.deployStrategy}' | jq .
```

You can see that it does exactly there, where it's adding, clearing, re-adding that `deployIdentifer`.

slagle: That would be the issue then if deploy is not getting set back to false.

slagle: I don't see the behavior you describe. How exactly did you reproduce it?

This is what I did:

I have a diff to disable the actual deployment execution for a node, and it adds a 5 sec sleep. This gives me time to query the node and check the deployStrategy values:

```
diff --git a/controllers/openstackdataplanenode_controller.go b/controllers/openstackdataplanenode_controller.go
index 40c383c..425ee30 100644
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -191,7 +191,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		}
 	}
 
-	nodeConfigMap, err := r.GenerateInventory(ctx, instance, instanceRole)
+	// nodeConfigMap, err := r.GenerateInventory(ctx, instance, instanceRole)
 	if err != nil {
 		util.LogErrorForObject(helper, err, fmt.Sprintf("Unable to generate inventory for %s", instance.Name), instance)
 		return ctrl.Result{}, err
@@ -204,6 +204,7 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		if len(instance.Spec.DeployStrategy.DeployIdentifier) == 0 {
 			// patch := client.MergeFrom(instance.DeepCopy())
 			newIdentifier := dataplanev1beta1.GenerateDeployIdentifier()
+			r.Log.Info("#############GENERATING DEPLOYIDENTIFIER", "newIdentifier", newIdentifier)
 			instance.Spec.DeployStrategy.DeployIdentifier = newIdentifier
 			err = r.Client.Update(ctx, instance)
 			if err != nil {
@@ -216,11 +217,13 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Info("Set DeploymentReadyCondition false", "instance", instance)
 		instance.Status.Conditions.Set(condition.FalseCondition(condition.DeploymentReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.DeploymentReadyRunningMessage))
 
-		nodes := &dataplanev1beta1.OpenStackDataPlaneNodeList{
-			Items: []dataplanev1beta1.OpenStackDataPlaneNode{*instance},
-		}
+		// nodes := &dataplanev1beta1.OpenStackDataPlaneNodeList{
+		// 	Items: []dataplanev1beta1.OpenStackDataPlaneNode{*instance},
+		// }
 
-		deployResult, err := deployment.Deploy(ctx, helper, instance, nodes, ansibleSSHPrivateKeySecret, nodeConfigMap, &instance.Status, instance.GetAnsibleEESpec(*instanceRole), r.GetServices(instance, instanceRole), instanceRole)
+		// deployResult, err := deployment.Deploy(ctx, helper, instance, nodes, ansibleSSHPrivateKeySecret, nodeConfigMap, &instance.Status, instance.GetAnsibleEESpec(*instanceRole), r.GetServices(instance, instanceRole), instanceRole)
+		r.Log.Info("##############SLEEPING FOR 5 SECS")
+		time.Sleep(5 * time.Second)
 		if err != nil {
 			util.LogErrorForObject(helper, err, fmt.Sprintf("Unable to deploy %s", instance.Name), instance)
 			instance.Status.Conditions.Set(condition.FalseCondition(
@@ -231,10 +234,10 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 				err.Error()))
 			return ctrl.Result{}, err
 		}
-		if deployResult != nil {
-			result = *deployResult
-			return result, nil
-		}
+		// if deployResult != nil {
+		// 	result = *deployResult
+		// 	return result, nil
+		// }
 
 		r.Log.Info("Set DeploymentReadyCondition true", "instance", instance)
```

I make sure CRD's are updated in the cluster, start the operator (go run main.go), create the sample edpm-compute-0 (with the role line removed), patch the node to set deployStrategy.deploy to true, then check the deployStrategy values.

```
oc create -f config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0.yaml
oc patch openstackdataplanenode edpm-compute-0  -p='[{"op": "replace", "path": "/spec/deployStrategy/deploy", "value":true}]' --type json
oc get osdpnode edpm-compute-0 -o json | jq .spec.deployStrategy
```

Once I hit the sleep, I get the following:

```
{
  "deploy": true,
  "deployIdentifier": "406efbab-6fda-4b3b-8e75-ab41ab4a7979"
}
```

Once past the sleep, deploy is toggled back to false and deployIdentifier is dropped (I suppose due to it being set back to ""). I never see a new value generated.

```
{
  "deploy": false
}
```

To simulate an actual deployment (with requeues as if it was watching for changes to an OpenStackAnsibleEE), I add the following line after where I've commented out the check for deployResult:

```
// if deployResult != nil {                           
//  result = *deployResult                            
//  return result, nil                                
// }                                                  
return ctrl.Result{RequeueAfter: time.Second * 2}, err
```

In that case, I see a single unique value for deployIdentifier generated and it stays consistent with that value (not changing).

However, if the node is owned by the dataplane, then setting deploy=true on the node and then seeing deployIdentifier constantly getting reset, I could see that happening due to the dataplane reconcile overwriting changes done to its owned node resource. Could that explain the behavior you saw?

slagle: I can see a situation where if deploying from the DataPlane, the role could get suck in a loop. I'll comment in the code where I think this is happening.

controllers/openstackdataplanerole_controller.go
slagle: I'm not opposed to tracking this in the Status. If it makes the implementation easier to follow, I'm all for it.

controllers/openstackdataplane_controller.go
```diff
@@ -133,6 +135,7 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	var deployErrors []string
 	shouldRequeue := false
 	if instance.Spec.DeployStrategy.Deploy {
+
```
slagle, line 138: Add the logic here as well to set the DeployIdentifier if not set.

slagle: At line 170, instead of just setting DeployStrategy.Deploy on the role, set DeployStrategy. That way the role gets the same DeployIdentifier as the DataPlane, and the role won't generate a new one.

slagle: Before line 166, we need to add a check if the role IsReady, and if the DeployIdentifier on the role matches that of the DataPlane. If both are true, then that means the role completed the deploy associated with this DataPlane deploy request. If that's the case, don't set DeployStrategy on the role again.

I think that's what's causing the loop....the role finishes and sets it's deploy=false, then from here it's get set back to true.

rebtoor: Hey @slagle, i tried this approach but it didn't work. For what I saw the role deployIdentifer and the dataplane deployIdentifier are always different...i guess because we never pass "deployStrategy" from dataplane to dataplanerole (in the CreateOrPatch).

slagle: It looks like it's doing the right thing to me. I used some of the kuttl test files to test it:


```shell
Mon Jun 05 02:36:23 (jslagle@teletran-1) [~/code/osp-ng/openstack-k8s-operators/dataplane-operator] (deployIdentifier)$ oc create -f tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml 
openstackdataplane.dataplane.openstack.org/openstack-edpm-no-nodes created

Mon Jun 05 03:51:31 (jslagle@teletran-1) [~/code/osp-ng/openstack-k8s-operators/dataplane-operator] (deployIdentifier)$ oc apply -f tests/kuttl/tests/dataplane-deploy-no-nodes-test/01-dataplane-deploy.yaml 
openstackdataplane.dataplane.openstack.org/openstack-edpm-no-nodes configured


Mon Jun 05 03:51:59 (jslagle@teletran-1) [~/code/osp-ng/osp-ng] (main)$ oc get osdp openstack-edpm-no-nodes -o json | jq .spec.deployStrategy
{
  "deploy": true,
  "deployIdentifier": "31822b62-c984-4ee7-9d17-73feb48a66ea"
}

Mon Jun 05 03:52:40 (jslagle@teletran-1) [~/code/osp-ng/osp-ng] (main)$ oc get osdprole edpm-compute-no-nodes -o json | jq .spec.deployStrategy
{
  "deploy": true,
  "deployIdentifier": "31822b62-c984-4ee7-9d17-73feb48a66ea"
}

```

The other thing I am seeing though is an infinite reconcile loop between DataPlane and the role. Looks like the DataPlane keeps trying to toggle deploy back to true on the role.

controllers/openstackdataplane_controller.go
```diff
@@ -163,31 +184,36 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-				_, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), &role, func() error {
-					r.Log.Info("Reconciling Role", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-					helper.GetLogger().Info("CreateOrPatch Role.DeployStrategy.Deploy", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-					role.Spec.DeployStrategy.Deploy = instance.Spec.DeployStrategy.Deploy
+			// fmt.Printf("role deployidentifier: %s\n", role.Spec.DeployStrategy.DeployIdentifier)
+			// fmt.Printf("instance deployidentifier: %s\n", instance.Spec.DeployStrategy.DeployIdentifier)
+			if role.IsReady() && role.Spec.DeployStrategy.DeployIdentifier == instance.Spec.DeployStrategy.DeployIdentifier {
+				r.Log.Info("Role successfully deployed", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
+			} else {
+				if !role.Spec.DeployStrategy.Deploy {
```
slagle, line 192: this line needs to check DeployIdentifier on the role. something like:

```
if !role.Spec.DeployStrategy.Deploy && role.Status.DeployIdentifier != instance.Spec.DeployStrategy.DeployIdentifier {
```

we don't want this to keep setting deploy=true on the role after the role has finished, that creates a loop.

slagle:
This will not be needed with https://github.com/openstack-k8s-operators/dataplane-operator/pull/378

PR #176, opened 2023-05-05
Add kuttl tests for conditions and refactor conditions

slagle:
- Use DeploymentReadyCondition and separate from ReadyCondition
- Add SetupReadyCondition

Depends On https://github.com/openstack-k8s-operators/lib-common/pull/259


controllers/openstackdataplane_controller.go
```diff
@@ -98,8 +98,6 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 			// something is not ready so reset the Ready condition
 			instance.Status.Conditions.MarkUnknown(
 				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
-			// and recalculate it based on the state of the rest of the conditions
-			instance.Status.Conditions.Set(instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
```
gibizer, line 101: I think the source of the issue is that IsReady() was depending on the ReadyCondition itself. 

In nova-operator we have a bit different logic where we don't use IsReady() to decide if we are ready to avoid the logical loop between ReadyConditon and IsReady(). Instead set ReadyCondition to True if all the conditions are true execpt Ready. Note that allSubConditionIsTrue now available in lib-common  now as [`condition.AllSubConditionIsTrue()`](https://github.com/openstack-k8s-operators/lib-common/blob/1db7d191885e82ed25e51a34536c601a9082275f/modules/common/condition/funcs.go#L164).
```golang
	defer func() {
		// update the Ready condition based on the sub conditions
		if allSubConditionIsTrue(instance.Status) {
			instance.Status.Conditions.MarkTrue(
				condition.ReadyCondition, condition.ReadyMessage)
		} else {
			// something is not ready so reset the Ready condition
			instance.Status.Conditions.MarkUnknown(
				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
			// and recalculate it based on the state of the rest of the conditions
			instance.Status.Conditions.Set(
				instance.Status.Conditions.Mirror(condition.ReadyCondition))
		}
		err := h.PatchInstance(ctx, instance)
		if err != nil {
			_err = err
			return
		}
	}()
```
https://github.com/openstack-k8s-operators/nova-operator/blob/94103fbbf2b936349b8554e11f9235db094ab2e0/controllers/nova_controller.go#L125-L143

So I'm not sure you need to remove `instance.Status.Conditions.Set(instance.Status.Conditions.Mirror(condition.ReadyCondition))` here. Removing it will cause that ReadyCondition remains in Unknown state even if the Reconciliation failed and some of the other conditions already went to False with an error message.

controllers/openstackdataplanenode_controller.go
```diff
@@ -146,6 +133,10 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 			// and recalculate it based on the state of the rest of the conditions
 			instance.Status.Conditions.Set(instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
+
+		// Ensure conditions are always sorted by type
+		instance.Status.Conditions.Sort()
```
gibizer, line 138: If this is really needed then I think we have a bug in lib-common. I think all the condition modifying function in lib-common ensure that the list is sorted.

controllers/openstackdataplane_controller.go
slagle: You are correct about IsReady(). I do have that in the commit message, but that does not show up in the PR body.

You are also correct about removing this line not being necessary. I believe I had started to refactor all 3 controllers, and then I missed adding this one back when I settled on an approach. I have added it back in the next commit, added similar conditions and handling to DataPlane, and updated the kuttl tests.

controllers/openstackdataplanenode_controller.go
slagle: Thanks, I've removed these lines. They are indeed sorted in the call to Set().

gibizer (commented):
my comments were resolved so this looks good to me

slagle:
The kuttl tests are being enabled in CI in https://github.com/openshift/release/pull/39222

PR #183, opened 2023-05-13
Don't return error  when baremetalset not ready

rabi:
Also ensures that we don't patch role object multiple times to create an infinite reconciliation loop.


bshephar (approved):
lgtm. Tested this locally, works as described for me. Although, I didn't get any Nova jobs launched, but that issue appears to be local and unrelated to this PR.

PR #186, opened 2023-05-15
Don't requeue when OpenStackBaremetalSet is not ready

slagle:
Since the OpenStackBaremetalSet is owned by the role controller, there
is no need to requeue if it's not ready. The role controller reconcile
loop will be automatically triggered on any change to any owned
resource. This removes needless reconcile loops and relies on the
builtin event watching.


controllers/openstackdataplanerole_controller.go
```diff
@@ -297,18 +293,18 @@ func (r *OpenStackDataPlaneRoleReconciler) ReconcileBaremetalSet(ctx context.Con
...
 			dataplanev1beta1.RoleBaremetalProvisionErrorMessage)
-		return ctrl.Result{}, err
+		return &ctrl.Result{}, err
 	}
 
 	// Check if baremetalSet is ready
 	if !baremetalSet.IsReady() {
 		util.LogForObject(helper, "BaremetalSet not ready, Requeueing", instance)
-		return ctrl.Result{RequeueAfter: bmProvisionRetryDelay}, nil
+		return nil, nil
```
rabi, line 302: Would not returning `nil` here would  result it moving ahead (ignored in L155-158 check) and try deploy on not yet provisioned nodes, if the deploy flag is changed to true?

controllers/openstackdataplanerole_controller.go
```diff
@@ -275,7 +271,7 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 
 // ReconcileBaremetalSet Reconcile OpenStackBaremetalSet
 func (r *OpenStackDataPlaneRoleReconciler) ReconcileBaremetalSet(ctx context.Context, instance *dataplanev1beta1.OpenStackDataPlaneRole, helper *helper.Helper,
-) (ctrl.Result, error) {
+) (*ctrl.Result, error) {
```
rabi, line 274: If we think moving ahead with deployment and then retrying (for failed  AnsibleExecutionJobs is fine) till baremetalset is ready, we can remove ctrl.Result return value from this function.

controllers/openstackdataplanerole_controller.go
slagle: You are right, I accidentally reversed the logic between this line and line 307. I'll fix that.

controllers/openstackdataplanerole_controller.go
slagle: When I fix the previous comment, this will return &ctrl.Result{} when the baremetalset is not ready, which will cause the main reconcile to stop and return as well.

controllers/openstackdataplanerole_controller.go
```diff
@@ -157,9 +153,9 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 	if len(instance.Spec.BaremetalSetTemplate.BaremetalHosts) > 0 {
 		ctrlResult, err := r.ReconcileBaremetalSet(ctx, instance, helper)
 		if err != nil {
-			return ctrlResult, err
-		} else if (ctrlResult != ctrl.Result{}) {
-			return ctrlResult, nil
```
rabi, line 162: returning and checking ctrlResult as `nil` is not standard and confusing.  We could return just  a 'isReady' boolean from ReconcileBaremetalSet() rather than *ctrl.Result I guess. Also, below check can be simplified.
if (err != nil || ctrlResult != nil) {
    return ctrlResult, err}

slagle: Done.

slagle:
recheck

PR #189, opened 2023-05-17
Enable service play

fao89:
I'm working on https://github.com/openstack-k8s-operators/install_yamls/pull/277
and it would be interesting to have a service for installing tripleo repos instead of calling `make edpm_compute_repos` N times

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
@@ -29,6 +29,8 @@ spec:
             properties:
               label:
                 type: string
+              play:
+                type: string
               role:
                 properties:
                   any_errors_fatal:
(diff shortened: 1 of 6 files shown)


config/services/dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml
```diff
@@ -0,0 +1,24 @@
...
+      tasks:
+        - name: Enable podified-repos
+          shell: |
+            rpm -q git || sudo yum -y install git
+            sudo yum -y install python-setuptools python-requests python3-pip
+            git clone https://github.com/openstack-k8s-operators/repo-setup
+            pushd repo-setup
+            sudo python3 setup.py install
+            popd
+            sudo /usr/local/bin/repo-setup current-podified-dev
```
slagle, line 24: I don't think we want a service here that enables our developement repos. That should stay in install_yamls. The rest of the change looks ok to me.

slagle (changes requested):

slagle (approved):

PR #195, opened 2023-05-22
Remove openstackdataplaneservice controller

rabi:
We're not doing anything in the openstackdataplaneservice reconciliation. The controller is kind of redundant atm. I suspect this would be some kind of static configuration of the play to run. We could have used a configmap instead, but I guess CRD will give us the opportunity to vlaidate/default the spec attributes.

This patch proposes to remove the controller unless it's expected to do something meaningful in the future.


slagle (commented):
+2, LGTM.

There's no need for the controller atm. The DataPlaneService is really just providing us an interface, schema, validation, etc.

PR #201, opened 2023-05-23
Profile dataplane tasks

fao89:


config/samples/dataplane_v1beta1_openstackdataplane.yaml
```diff
@@ -45,6 +45,8 @@ spec:
           value: "True"
         - name: ANSIBLE_VERBOSITY
           value: "2"
+        - name: "ANSIBLE_CALLBACKS_ENABLED"
+          value: "profile_tasks"
```
fao89, line 49: @slagle @fultonj @gibizer @SeanMooney 
if I could set/unset `ANSIBLE_CALLBACKS_ENABLED` in one place, it would be great for debugging, 
so instead of setting them on dataplane,nova,... we could set in one place and it propagates.
I'm wondering how can we have global default config for AEE

fultonj: @fao89 I take it this only sets it for dataplane but that it does so correctly. In that case I'm +2 to merge it though I agree it would be nice to easily set this option for the nova and libvirt ansible runs too.

fao89: I've started this: https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/158

slagle: I like the approach over the ansibleee PR. I left a review there.

fao89: I'll close this PR and move forward with the ansibleee PR

PR #203, opened 2023-05-24
Migrate composable services

slagle:
- Move services under NodeSection.
- Migrate ValidateNetwork to validate-network composable service
- Migrate OS services to composable service interface
- Migrate OpenStack tasks into OS


config/samples/dataplane_v1beta1_openstackdataplane.yaml
```diff
@@ -55,6 +53,12 @@ spec:
         domainName: osptest.openstack.org
       preProvisioned: true
       nodeTemplate:
+        services:
+          - configure-network
+          - validate-network
+          - configure-os
+          - install-os
+          - run-os
```
fao89, line 61: we will need to update: https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L74-L78

slagle: https://github.com/openstack-k8s-operators/install_yamls/pull/290 should fix that

slagle:
recheck

slagle:
recheck

slagle:
recheck

pkg/deployment/deployment.go
```diff
@@ -99,127 +99,6 @@ func Deploy(
...
-		deployLabel,
-		aeeSpec,
-		foundService)
-
-	if err != nil || !status.Conditions.IsTrue(readyCondition) {
-		return &ctrl.Result{}, err
-	}
-	log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
-
 	// ConfigureCephClient
```
fao89, line 102: @fultonj does ceph need to be installed specifically before/after some service?

fultonj: @fao89 `ConfigureCephClient` needs to be done before we create the NovaExternal CR.

tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml
```diff
@@ -9,5 +9,9 @@ spec:
     edpm-compute-no-nodes:
       nodeTemplate:
         ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
-      services:
-        - configure-network
+        services:
```
fultonj, line 12: I wonder if ceph-client should be added as a `DataPlaneService` here in a follow up patch between `validate-network` and `configure-os`. I don't think not having it here needs to block this patch. It's execution is conditional on the data-plane CR having an `ExtraMounts` of type `ceph`.

fultonj (commented):
+2

fao89 (approved):
recheck

fao89:
recheck

rebtoor (approved):

SeanMooney:
it would be nice to update https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml

to work with this.

i just got to the point of using that to deploy a edpm node with nova which works but because 
you canot list openstackDataplaneService on a openstackDatapalneNode i cannot complete the deployment of nova properly since the networks are not configured

https://github.com/openstack-k8s-operators/dataplane-operator/commit/a33ea06b897bb7b58f4716e1bdd11e88712657e2 moved that to only work with the service interface and broke the standalone dataplane node workflow as a result

SeanMooney:
check-rdo

tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml
slagle: Yes, I think it needs to be a composable service. I can follow up with that.

rebtoor (approved):

fultonj (approved):
+2

slagle:
/test dataplane-operator-build-deploy-kuttl

 INFO[2023-05-30T12:37:16Z] Waiting for cluster claim openstack-k8s-operators-cluster-pool/6f14d3bf-04c4-48e5-af7e-2ccad4e60103 to be fulfilled. 
WARN[2023-05-30T13:37:16Z] Found 1 conditions for ClusterClaim:
  *[2023-05-30T12:37:16Z]NoClusters: No clusters in pool are ready to be claimed 
INFO[2023-05-30T13:37:16Z] Releasing cluster claims for test dataplane-operator-build-deploy-kuttl 
INFO[2023-05-30T13:37:16Z] Ran for 1h5m46s                              
ERRO[2023-05-30T13:37:16Z] Some steps failed:                           
ERRO[2023-05-30T13:37:16Z] 
  * could not run steps: step dataplane-operator-build-deploy-kuttl failed: failed to wait for the created cluster claim to become ready: timed out waiting for the condition

slagle:
recheck

config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0_full.yaml
```diff
@@ -7,6 +7,12 @@ spec:
   ansibleHost: 192.168.122.100
   ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
   node:
+    services:
+      - configure-network
+      - validate-network
```
SeanMooney, line 12: this si a nit or question more then a request for change but should we be validating before or after?

is there a general pattern or nameing convention we should set here going forward

i.e. is validate-network ensuring the input is correct or is it ensuring the network is not configured as expected.

i would assume the former in which case i think the order is reversed.

configure-network should fail if it was not able to configure the network requested so if its the latter (validating that configure-network did the right thing) then i think we should combine the two

slagle: configure-network applies the user requested network configuration. validate-network then validates it. configuring something does not imply validation, just in the same way that if you set an option in a conf file, it is not validated until later when used (in most cases anyway).

slagle:
recheck

slagle:
recheck

raukadah:
recheck

api/v1beta1/common.go
```diff
@@ -74,8 +74,9 @@ type NodeSection struct {
 	NetworkData *corev1.SecretReference `json:"networkData,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={configure-network,validate-network,install-os,configure-os,run-os}
```
fao89, line 77: 👍🏽

fao89 (approved):

raukadah:
recheck

raukadah:
recheck

raukadah:
recheck

raukadah:
@slagle hitting this currently in [ci](https://logserver.rdoproject.org/03/203/464a4bd38c9d7300ed8b026e2c444ac3e31f29cd/github-check/dataplane-operator-crc-podified-edpm-deployment/5113879/ci-framework-data/logs/ci_make_005_run_edpm_deploy.log) during edpm_deploy.
```
+ '[' '!' -f ansibleee-ssh-key-id_rsa ']'
+ cat
+ oc apply -f namespace.yaml
+ oc create secret generic dataplane-ansible-ssh-private-key-secret --save-config --dry-run=client --from-file=authorized_keys=ansibleee-ssh-key-id_rsa.pub --from-file=ssh-privatekey=ansibleee-ssh-key-id_rsa --from-file=ssh-publickey=ansibleee-ssh-key-id_rsa.pub -o yaml
+ oc apply -f -
+ popd
+ popd
error: add operation does not apply: doc is missing path: "/spec/roles/edpm-compute/services/0": missing value
error: no objects passed to apply
gmake: *** [Makefile:493: edpm_deploy] Error 1
```

fao89:
@raukadah this PR changes the path, install_yamls need to be updated, I'll start a PR

fao89:
@raukadah https://github.com/openstack-k8s-operators/install_yamls/pull/305

raukadah:
recheck

raukadah (approved):

PR #204, opened 2023-05-24
[envtest]Enhance test make target

gibizer:
I switched to ginkgo cli to execute the env test. This allows better configurability of the execution. To still run the existing unit test there is a separate go test call and a ginkgo call in make test target.

Added GINKGO_ARGS so that we can pass parameters to the test runner, e.g.
```
  make test GINKGO_ARGS="-v --focus 'Should have created a OpenStackDataplaneRole'"
```
will run a single test case with verbose logging.

Also added parallel test execution and for that disabled the metrics service as that wants to bind to the same port (8080) from each parallel tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,9 +108,14 @@ vet: gowork ## Run go vet against code.
 	go vet ./...
 	go vet ./api/...
 
+PROCS?=$(shell expr $(shell nproc --ignore 2) / 2)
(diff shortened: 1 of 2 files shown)


slagle:
recheck

raukadah:
recheck

rebtoor (approved):

slagle (approved):

PR #205, opened 2023-05-25
Make dataplanenode controller watch dataplanerole

rabi:
Make the dataplanenodes reconcile for dataplanerole. This would ensure the following:

- Role changes would result in node recocnile.
- This would mark the node as Ready when the role is Ready.
- We don't have to update dataplane node status from dataplanerole controller which is an anti-pattern.


rabi:
recheck

rabi:
recheck

rabi:
recheck

rabi:
recheck

rebtoor:
recheck

slagle (approved):

PR #216, opened 2023-06-01
OpenStackDataPlane sample CR for Nova Ceph Config

fultonj:
Add dataplane_v1beta1_openstackdataplane_ceph.yaml based on dataplane_v1beta1_openstackdataplane.yaml but also has extraMounts and a customServiceConfig for nova. These parameters have been confirmed to configure an EDPM compute node which can store its ephemeral data in a Ceph RBD pool. This will also be an example referenced from the documentation [1].

[1] https://github.com/openstack-k8s-operators/docs/pull/36

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplane_ceph.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane_ceph.yaml
@@ -1,0 +1,184 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlane
+metadata:
+  name: openstack-edpm
+spec:
+  deployStrategy:
+      deploy: false
+  nodes:
+    edpm-compute-0:
+      role: edpm-compute
+      hostName: edpm-compute-0
+      ansibleHost: 192.168.122.100
+      node:
+        ansibleVars: |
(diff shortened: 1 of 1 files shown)


config/samples/dataplane_v1beta1_openstackdataplane_ceph.yaml
```diff
@@ -0,0 +1,180 @@
...
+          storage_ip: 172.18.0.101
+          tenant_ip: 172.19.0.101
+          fqdn_internal_api: edpm-compute-1.example.com
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+      deployStrategy:
+        deploy: false
+  roles:
+    edpm-compute:
+      services:
+        - configure-network
```
fao89, line 40: it needs to be updated

fao89 (approved):
left one minor comment, but LGTM

config/samples/dataplane_v1beta1_openstackdataplane_ceph.yaml
fultonj: Thank you. I have updated that in my most recent PS.

PR #217, opened 2023-06-02
Add deployProtection annotation

bshephar:
When a node is already deployed, we need to protect against eroneous Ansible executions. This is essential to ensure predictability in our deployments and ensure nodes are only being touched when there is explicit intent to do so.

This change annotates the OpenStackDataPlaneNode at the end of the deployment.

### DIFF
--- a/controllers/openstackdataplane_controller.go
+++ b/controllers/openstackdataplane_controller.go
@@ -24,6 +24,7 @@ import (
 
 	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/dataplane-operator/pkg/deployment"
+	dataplaneutil "github.com/openstack-k8s-operators/dataplane-operator/pkg/util"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
@@ -47,6 +48,15 @@ type OpenStackDataPlaneReconciler struct {
 	Kclient kubernetes.Interface
 	Scheme  *runtime.Scheme
 	Log     logr.Logger
+}
+
+const redeployProtectionAnnotationName string = "edpm.openstack.org/deploy-protection"
+
+// redeployProtectionAnnoation provides protection against eroneous Ansible exections. This annotation
+// should be explicitly cleared by any controller or user who wishes to re-run a Ansible execution against
+// a node that has already been deployed.
+var redeployProtectionAnnotation = map[string]string{
+	redeployProtectionAnnotationName: "true",
 }
 
 //+kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplanes,verbs=get;list;watch;create;update;patch;delete
@@ -128,6 +138,13 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	// all setup tasks complete, mark SetupReadyCondition True
 	instance.Status.Conditions.MarkTrue(dataplanev1.SetupReadyCondition, condition.ReadyMessage)
 
+	// Check to see if the OpenStackDataPlane has the redeployProtectionAnnotation. If it does, we will exit the Reconcile here.
+	_, unsafeToExecute := dataplaneutil.CheckDeployProtectionAnnotation(instance, redeployProtectionAnnotation)
+	if unsafeToExecute {
+		r.Log.Info("This OpenStackDataPlane object currently has the deployProtection annotation, which will prevent further task execution")
+		return ctrl.Result{Requeue: false}, nil
+	}
+
 	ctrlResult, err := createOrPatchDataPlaneResources(ctx, instance, helper)
 	if err != nil {
 		return ctrl.Result{}, err
@@ -137,60 +154,56 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	var deployErrors []string
 	shouldRequeue := false
-	if instance.Spec.DeployStrategy.Deploy {
-		r.Log.Info("Starting DataPlane deploy")
-		r.Log.Info("Set DeploymentReadyCondition false", "instance", instance)
-		instance.Status.Conditions.Set(condition.FalseCondition(condition.DeploymentReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.DeploymentReadyRunningMessage))
-		roles := &dataplanev1.OpenStackDataPlaneRoleList{}
 
-		listOpts := []client.ListOption{
-			client.InNamespace(instance.GetNamespace()),
-		}
-		labelSelector := map[string]string{
-			"openstackdataplane": instance.Name,
-		}
-		if len(labelSelector) > 0 {
-			labels := client.MatchingLabels(labelSelector)
-			listOpts = append(listOpts, labels)
-		}
-		err = r.Client.List(ctx, roles, listOpts...)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
+	r.Log.Info("Starting DataPlane deploy")
+	r.Log.Info("Set DeploymentReadyCondition false", "instance", instance)
+	instance.Status.Conditions.Set(condition.FalseCondition(condition.DeploymentReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.DeploymentReadyRunningMessage))
+	roles := &dataplanev1.OpenStackDataPlaneRoleList{}
 
-		for _, role := range roles.Items {
-			logger.Info("DataPlane deploy", "role.Name", role.Name)
-			if role.Spec.DataPlane != instance.Name {
-				err = fmt.Errorf("role %s: role.DataPlane does not match with role.Label", role.Name)
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.GetNamespace()),
+	}
+	labelSelector := map[string]string{
+		"openstackdataplane": instance.Name,
+	}
+	if len(labelSelector) > 0 {
+		labels := client.MatchingLabels(labelSelector)
+		listOpts = append(listOpts, labels)
+	}
+	err = r.Client.List(ctx, roles, listOpts...)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	for _, role := range roles.Items {
+		logger.Info("DataPlane deploy", "role.Name", role.Name)
+		if role.Spec.DataPlane != instance.Name {
+			err = fmt.Errorf("role %s: role.DataPlane does not match with role.Label", role.Name)
+			deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+err.Error())
+		}
+		r.Log.Info("Role", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
+		_, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), &role, func() error {
+			r.Log.Info("Reconciling Role", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
+			helper.GetLogger().Info("CreateOrPatch", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
+			if err != nil {
 				deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+err.Error())
 			}
-			r.Log.Info("Role", "DeployStrategy.Deploy", role.Spec.DeployStrategy.Deploy, "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-			if !role.Spec.DeployStrategy.Deploy {
-				_, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), &role, func() error {
-					r.Log.Info("Reconciling Role", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-					helper.GetLogger().Info("CreateOrPatch Role.DeployStrategy.Deploy", "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-					role.Spec.DeployStrategy.Deploy = instance.Spec.DeployStrategy.Deploy
-					if err != nil {
-						deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+err.Error())
-					}
-					return nil
-				})
-				if err != nil {
-					deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+err.Error())
+			return nil
+		})
+		if err != nil {
+			deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+err.Error())
+		}
+		if !role.IsReady() {
+			r.Log.Info("Role", "IsReady", role.IsReady(), "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
+			shouldRequeue = true
+			mirroredCondition := role.Status.Conditions.Mirror(condition.ReadyCondition)
+			if mirroredCondition != nil {
+				r.Log.Info("Role", "Status", mirroredCondition.Message, "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
+				if condition.IsError(mirroredCondition) {
+					deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+mirroredCondition.Message)
 				}
 			}
-			if !role.IsReady() {
-				r.Log.Info("Role", "IsReady", role.IsReady(), "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-				shouldRequeue = true
-				mirroredCondition := role.Status.Conditions.Mirror(condition.ReadyCondition)
-				if mirroredCondition != nil {
-					r.Log.Info("Role", "Status", mirroredCondition.Message, "Role.Namespace", instance.Namespace, "Role.Name", role.Name)
-					if condition.IsError(mirroredCondition) {
-						deployErrors = append(deployErrors, "role.Name: "+role.Name+" error: "+mirroredCondition.Message)
-					}
-				}
 
-			}
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


controllers/openstackdataplane_controller.go
```diff
@@ -135,7 +142,15 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-	if instance.Spec.DeployStrategy.Deploy {
+
+	unsafeToExecute := r.checkDeployProtectionAnnotation(instance, redeployProtectionAnnotation)
+
+	if unsafeToExecute {
+		r.Log.Info("This OpenStackDataPlane object currently has the deployProtection annotation, which will prevent further task execution")
+		return ctrl.Result{Requeue: false}, nil
+	}
+
+	if !unsafeToExecute {
```
gibizer, line 153: Because there is always a return if unsafeToExecute == true, you can remove this condition as you can only reach this point if unsafeToExecute == false

controllers/openstackdataplane_controller.go
```diff
@@ -412,3 +419,13 @@ func buildBMHHostMap(instance *dataplanev1beta1.OpenStackDataPlane,
 		}
 	}
 }
+
+// checkDeployProtectionAnnotation checks to see if the node has the deploy protection annotation set. If this is set,
+// we will block the execution of Ansible tasks until it is cleared.
+func (r *OpenStackDataPlaneReconciler) checkDeployProtectionAnnotation(instance *dataplanev1beta1.OpenStackDataPlane, annotation map[string]string) bool {
+	for k := range instance.ObjectMeta.GetAnnotations() {
+		_, keyExists := annotation[k]
+		return keyExists
```
gibizer, line 428: This now does not check the value of the annotation. So we have to carefully document that fact. E.g `"openstack.org/deploy-protection": "false"`, does not mean the protection is disabled.

bogdando: I think this also needs to check if keyExists and its value is true'ish or an empty string.

controllers/openstackdataplane_controller.go
```diff
@@ -135,7 +142,15 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	var deployErrors []string
 	shouldRequeue := false
-	if instance.Spec.DeployStrategy.Deploy {
```
bogdando, line 138: we might want to keep this guard as well, otherwise the whole DeployStrategy is a subject of removal within this change

controllers/openstackdataplanenode_controller.go
```diff
@@ -204,8 +204,15 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 	// all setup tasks complete, mark SetupReadyCondition True
 	instance.Status.Conditions.MarkTrue(dataplanev1beta1.SetupReadyCondition, condition.ReadyMessage)
 
-	r.Log.Info("Node", "DeployStrategy", instance.Spec.DeployStrategy.Deploy, "Node.Namespace", instance.Namespace, "Node.Name", instance.Name)
-	if instance.Spec.DeployStrategy.Deploy {
```
bogdando, line 208: ditto, this triggering condition needs to be preserved

controllers/openstackdataplanenode_controller.go
```diff
@@ -204,8 +204,15 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
+	r.Log.Info("Node", "Node.Namespace", instance.Namespace, "Node.Name", instance.Name)
+	unsafeToExecute := r.checkDeployProtectionAnnotation(instance, redeployProtectionAnnotation)
+
+	if unsafeToExecute {
+		r.Log.Info("This node currently has the deployProtection annotation, which will prevent further task execution")
+		return ctrl.Result{Requeue: false}, nil
+	}
+
+	if !unsafeToExecute {
```
bogdando, line 215: ditto, is redundant as well

controllers/openstackdataplanenode_controller.go
```diff
@@ -471,3 +478,13 @@ func (r *OpenStackDataPlaneNodeReconciler) GetServices(instance *dataplanev1beta
 	}
 	return instanceRole.Spec.NodeTemplate.Services
 }
+
+// checkDeployProtectionAnnotation checks to see if the node has the deploy protection annotation set. If this is set,
+// we will block the execution of Ansible tasks until it is cleared.
+func (r *OpenStackDataPlaneNodeReconciler) checkDeployProtectionAnnotation(instance *dataplanev1beta1.OpenStackDataPlaneNode, annotation map[string]string) bool {
+	for k := range instance.ObjectMeta.GetAnnotations() {
+		_, keyExists := annotation[k]
+		return keyExists
```
bogdando, line 487: ditto, an extended values check

controllers/openstackdataplanerole_controller.go
```diff
@@ -210,8 +210,14 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
 
+	unsafeToExecute := r.checkDeployProtectionAnnotation(instance, redeployProtectionAnnotation)
+	if unsafeToExecute {
+		r.Log.Info("This OpenStackDataPlaneRole object currently has the deployProtection annotation, which will prevent further task execution")
+		return ctrl.Result{Requeue: false}, nil
+	}
+
 	r.Log.Info("Role", "DeployStrategy", instance.Spec.DeployStrategy.Deploy, "Role.Namespace", instance.Namespace, "Role.Name", instance.Name)
-	if instance.Spec.DeployStrategy.Deploy {
+	if !unsafeToExecute {
```
bogdando, line 220: redundant condition

controllers/openstackdataplanerole_controller.go
```diff
@@ -407,3 +411,13 @@ func resolveAnsibleVars(node *dataplanev1beta1.NodeSection, host *ansible.Host,
 
 	return nil
 }
+
+// checkDeployProtectionAnnotation checks to see if the node has the deploy protection annotation set. If this is set,
+// we will block the execution of Ansible tasks until it is cleared.
+func (r *OpenStackDataPlaneRoleReconciler) checkDeployProtectionAnnotation(instance *dataplanev1beta1.OpenStackDataPlaneRole, annotation map[string]string) bool {
+	for k := range instance.ObjectMeta.GetAnnotations() {
+		_, keyExists := annotation[k]
+		return keyExists
```
bogdando, line 420: ditto (makes sense to move it to some shared place?)

controllers/openstackdataplane_controller.go
bshephar: It probably doesn't make sense to keep it tbh. If a user creates or updates the Dataplane spec, and removes the annotation. They would expect it to deploy like any other Kubernetes resource that has had it's spec updated. Having this as part of the spec doesn't align well with Kubernetes constructs and duplicates the intention behind the annotation I guess.

I think we're better off just removing it. But there are a lot of places I need to do that, and they each require some careful consideration around the original intention, and how that intention can be preserved as part of this change. Leave it with me, I'll send another revision of this PR early next week when I'm back from leave and remove these. I'll be interested to hear your inputs on my implementation to see what you think about it.

controllers/openstackdataplane_controller.go
```diff
@@ -47,6 +47,13 @@ type OpenStackDataPlaneReconciler struct {
 	Log     logr.Logger
 }
 
+// redeployProtectionAnnoation provides protection against eroneous Ansible exections. This annotation
+// should be explicitly cleared by any controller or user who wishes to re-run a Ansible execution against
+// a node that has already been deployed.
+var redeployProtectionAnnotation = map[string]string{
+	"openstack.org/deploy-protection": "true",
```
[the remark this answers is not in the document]
bshephar: Agreed, I'll change this too.

controllers/openstackdataplane_controller.go
bshephar: done

controllers/openstackdataplanenode_controller.go
bshephar: I did this intentionally because I thought it might be better to have users completely remove the annotation, rather than allowing true and false values. I just thought it might be less complex from a user experience perspective. But happy to get a few more inputs on this one and we can implement the agreed upon handling.

controllers/openstackdataplanerole_controller.go
bshephar: Removed, thanks.

controllers/openstackdataplane_controller.go
bshephar: This was intentional. I thought it would ensure there is definitely explicit intent if the user has to completely remove the annotation rather than just changing it to false. But if we think it should accept both true and false, we can change it also compare values.

controllers/openstackdataplanerole_controller.go
bshephar: It will always need to accept a different type for the instance, ie `*dataplanev1beta1.OpenStackDataPlaneRole` `*dataplanev1beta1.OpenStackDataPlaneNode`, `*dataplanev1beta1.OpenStackDataPlane`. Not sure I can generic out those Types to put a single function in a common place.

bshephar: These have now been moved @bogdando into common.go. I was trying to use Go generics, but @gibizer pointed out that I can just use `client.Object` since all of the OSDP types will satisfy that interface. So that allowed for easier common functions:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/217/files#diff-482605cf1d0af2d53f36083c5b246016c84e00e3c34dedd05ebad6c8cec51dc8R10-R36

controllers/openstackdataplanerole_controller.go
```diff
@@ -210,8 +210,14 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 	// all setup tasks complete, mark SetupReadyCondition True
 	instance.Status.Conditions.MarkTrue(dataplanev1beta1.SetupReadyCondition, condition.ReadyMessage)
 
+	unsafeToExecute := r.checkDeployProtectionAnnotation(instance, redeployProtectionAnnotation)
+	if unsafeToExecute {
+		r.Log.Info("This OpenStackDataPlaneRole object currently has the deployProtection annotation, which will prevent further task execution")
+		return ctrl.Result{Requeue: false}, nil
+	}
+
 	r.Log.Info("Role", "DeployStrategy", instance.Spec.DeployStrategy.Deploy, "Role.Namespace", instance.Namespace, "Role.Name", instance.Name)
```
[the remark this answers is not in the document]
bshephar: Haven't forgotten about your comment here. Just addressing a bunch of other issues first.

So, I still need some interface that will actually trigger a redeploy. I was thinking maybe this is where we implement the `deployIdentifier` rather than a simple `bool` condition? Ultimately, I want to avoid having any controllers updating user owned resource specs. It's fine for a controller to update a Controller managed object's spec, it's the only thing that should be doing that. But they shouldn't do it to objects that aren't controller managed.

So, I was thinking we can take the removal of the Annotation as the user intention to allow re-runs. If the user has removed the annotation, then we can generate a new unique deployIdentifier, update the Status of the user owned object to indicate a change has been acknowledged, then pass the new unique `deployIdentifer` to the controller managed CR's in their spec. Which should trigger a re-run of the Ansible jobs.

I'll work on that next week as a seperate commit on this PR and we can debate what that implementation ultimately looks like.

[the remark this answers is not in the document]
bshephar: Well, I'll collab with @rebtoor on it, since he has already done all of the work for the `deployIdentifier`. We'll see how we can fit that in here and make some magic happen.

controllers/openstackdataplane_controller.go
```diff
@@ -207,34 +220,25 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	if instance.Status.Conditions.IsUnknown(condition.DeploymentReadyCondition) {
 		r.Log.Info("Set DeploymentReadyCondition false")
 		instance.Status.Conditions.Set(condition.FalseCondition(condition.DeploymentReadyCondition, condition.NotRequestedReason, condition.SeverityInfo, condition.DeploymentReadyInitMessage))
 	}
 
+	// We don't want another deploy triggered by any reconcile request, it
+	// should only be triggered when the user (or another controller)
+	// specifically clears this annotation and requests it.
+	r.Log.Info("Setting deploy protection annotation")
+	instance.ObjectMeta.SetAnnotations(redeployProtectionAnnotation)
```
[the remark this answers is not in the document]
bogdando: > the reconcile is requeued due to the failure and it will not see the annotation so ti will re-run the ansible tasks

Do we link the "successful deployment" from the step 1 with the object containing that deploy-protection annotation? If we do, the stored hash in its status should include its linked AEE job status, if there is a link to an existing job. Then the re-queued reconciler will notice that the re-calculated hash matches the linked AEE status and will re-add the lost deploy-protection annotation.  But if we delete the linked job, then ansible will become re-executed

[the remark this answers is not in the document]
gibizer: > > the reconcile is requeued due to the failure and it will not see the annotation so ti will re-run the ansible tasks
> 
> Do we link the "successful deployment" from the step 1 with the object containing that deploy-protection annotation? If we do, the stored hash in its status should include its linked AEE/job status, if there is a link to an existing one. Then the re-queued reconciler will notice that the re-calculated hash matches the linked AEE status and will re-add the lost deploy-protection annotation, while skipping ansible execution. But if we delete the linked job, then ansible would be executed again

This is what we do with the Jobs. i) we store the hash and ii) we don't delete the job long enough that we can make sure that i) is succeeded.

bshephar:
Looks like the annotation is applied correctly at the end. But we're failing to `configure-network`:
```
2023-07-03T13:48:58.476212574+00:00 stdout F [1;31mfatal: [edpm-compute-0]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: Could not resolve hostname edpm-compute-0: Name or service not known", "unreachable": true}
```

Looks like it was provisioned, although the name in BMO != edpm-compute-0:
```
2023-07-03T13:55:32.127176196+00:00 stderr F {"level":"info","ts":"2023-07-03T13:55:32Z","logger":"controllers.BareMetalHost","msg":"using hardware profile","baremetalhost":{"name":"crc-bmaas-0","namespace":"openstack"},"provisioningState":"provisioned","profile":"unknown"}
--
2023-07-03T13:55:32.127263348+00:00 stderr F {"level":"info","ts":"2023-07-03T13:55:32Z","logger":"controllers.BareMetalHost","msg":"verified access to the BMC","baremetalhost":{"name":"crc-bmaas-0","namespace":"openstack"},"provisioningState":"provisioned"}
2023-07-03T13:55:32.162919686+00:00 stderr F {"level":"info","ts":"2023-07-03T13:55:32Z","logger":"controllers.BareMetalHost","msg":"done","baremetalhost":{"name":"crc-bmaas-0","namespace":"openstack"},"provisioningState":"provisioned","requeue":true,"after":60}
```

controllers/openstackdataplane_controller.go
```diff
@@ -209,34 +222,28 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-
 	// Set DeploymentReadyCondition to False if it was unknown.
 	// Handles the case where the Node is created with
-	// DeployStrategy.Deploy=false.
+	// If the status is Unknown at this point. We will return to avoid setting the
+	// redeployProtectionAnnotaton
 	if instance.Status.Conditions.IsUnknown(condition.DeploymentReadyCondition) {
 		r.Log.Info("Set DeploymentReadyCondition false")
 		instance.Status.Conditions.Set(condition.FalseCondition(condition.DeploymentReadyCondition, condition.NotRequestedReason, condition.SeverityInfo, condition.DeploymentReadyInitMessage))
+		return ctrl.Result{}, nil
```
bshephar, line 238: We might need to implement a more comprehensive check here. We want to avoid setting the annotation unless the deployment has actually completed successfully. So I'll try returning here and see if that achieves the desired result, otherwise I might need to implement some additional logic to ensure we only set the annotation once everything has completed.

bshephar:
/recheck

```
Warning: resource namespaces/openstack-operators is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by oc apply. oc apply should only be used on resources created declaratively by either oc create --save-config or oc apply. The missing annotation will be patched automatically.
--
Warning: would violate PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (container "cert-manager" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "cert-manager" must set securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container "cert-manager" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
error: no matching resources found
```

controllers/openstackdataplanenode_controller.go
bshephar: Hmm, na, I don't like it. It's just very unnatural, it has no equivalence in Kubernetes - where you create the object and it doesn't rollout. So from a user perspective, this introduces an interface that increases the perceived complexity of our solution since we need to block normal Kubernetes workflows.

I think the challenge we have is that we have all of these nested CR's that can each be individually created. But then we can create them as part of the OpenStackDataplane CR as well. From a larger, outside of the scope of this change, I would challenge that we should even do that at all.

I'll have a think about how we can avoid triggering here, since I can see it creates two jobs for the node:
```
❯ oc get openstackansibleee
NAME                                                    NETWORKATTACHMENTS   STATUS   MESSAGE
dataplane-deployment-configure-network-edpm-compute                          True     AnsibleExecutionJob complete
dataplane-deployment-configure-network-edpm-compute-0                        True     AnsibleExecutionJob complete
dataplane-deployment-configure-os-edpm-compute                               False    AnsibleExecutionJob is running
dataplane-deployment-configure-os-edpm-compute-0                             False    AnsibleExecutionJob is running
dataplane-deployment-install-os-edpm-compute                                 True     AnsibleExecutionJob complete
dataplane-deployment-install-os-edpm-compute-0                               True     AnsibleExecutionJob complete
dataplane-deployment-validate-network-edpm-compute                           True     AnsibleExecutionJob complete
dataplane-deployment-validate-network-edpm-compute-0                         True     AnsibleExecutionJob complete
```

One created by the role:
```
❯ oc get openstackansibleee dataplane-deployment-configure-network-edpm-compute -o jsonpath='{.metadata.ownerReferences}' | jq .
[
  {
    "apiVersion": "dataplane.openstack.org/v1beta1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "OpenStackDataPlaneRole",
    "name": "edpm-compute",
    "uid": "9a8d6b55-fe97-4916-94eb-dbe88a3a5a33"
  }
]
```

And one by the node:
```
❯ oc get openstackansibleee dataplane-deployment-configure-network-edpm-compute-0 -o jsonpath='{.metadata.ownerReferences}' | jq .
[
  {
    "apiVersion": "dataplane.openstack.org/v1beta1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "OpenStackDataPlaneNode",
    "name": "edpm-compute-0",
    "uid": "a282f7a7-7ce5-48dc-87dd-120936de1a3f"
  }
]
```

Seems to do the same with `NovaExternalCompute` too, and complains about the resource being owned by another controller.

bshephar: `NovaExternalCompute` issue caused by this:

So the `OpenStackDataplaneRole` controller is trying to update it:
```
{"level":"error","ts":"2023-07-07T16:52:22+10:00","msg":"Reconciler error","controller":"openstackdataplanerole","controllerGroup":"dataplane.openstack.org","controllerKind":"OpenStackDataPlaneRole","OpenStackDataPlaneRole":{"name":"edpm-compute","namespace":"openstack-operators"},"namespace":"openstack-operators","name":"edpm-compute","reconcileID":"b64c7a8b-5bc2-4d43-b842-7eb01ac7a58c","error":"DeployNovaExternalCompute error:: Object openstack-operators/edpm-compute-0 is already owned by another OpenStackDataPlaneNode controller edpm-compute-0","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/Users/brendanshephard/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/brendanshephard/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/Users/brendanshephard/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:235"}
```

But it's owned by the `OpenStackDataplaneNode` controller:
```
❯ oc get novaexternalcompute -o yaml | yq '.items[].metadata.ownerReferences'
- apiVersion: dataplane.openstack.org/v1beta1
  blockOwnerDeletion: true
  controller: true
  kind: OpenStackDataPlaneNode
  name: edpm-compute-0
  uid: 7413790d-0453-462a-8749-b296d205fd1a
```

This is likely because all objects are essentially setting the equivalent of `deployStrategy.Deploy` now and we don't stop the execution of nodes while allowing the role. 

This is the summation of the issue caused by this removal, and something I will continue to work on.

rabi:
recheck

rebtoor:
recheck

rebtoor:
recheck

slagle:
Closing in favor of https://github.com/openstack-k8s-operators/dataplane-operator/pull/378

PR #223, opened 2023-06-08
Revert "Move services under NodeSection."

slagle:
This reverts commit df9df7c8ee524c7c149ab9ec63186d5ed36548ca.

There is not a straightforward way to implement the services list with
role inheritance given the current deployment design. The issue is that
a role deployment does not take a nodes customized list of services into
account.

The role would have to look for the services list on each node and come
up with a deployment order on the fly, and the right list of nodes to
run for each service. Services could also be in different orders on
different nodes. That is too complex to implement at the moment.

This also partially reverts commit
0b7007e14bb09c4957f88358d4b6fa44178cdc08. That commit added the default
services list to the samples. The list needs to be moved out from under
the nodeTemplate and node sections to being under the spec directly
instead.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/313

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
@@ -917,16 +917,6 @@ spec:
                         default: nova
                         type: string
                     type: object
-                  services:
-                    default:
-                    - configure-network
-                    - validate-network
-                    - install-os
-                    - configure-os
-                    - run-os
-                    items:
-                      type: string
-                    type: array
                   userData:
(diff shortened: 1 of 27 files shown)


slagle:
recheck

fultonj:
+2

I ran a copy of dataplane-operator with this revert in my environment and it solved the following problem for me:

If you create a DataPlanNode CR which has nodes and roles and the roles template has a shortened services list, then the nodes are all created with a full services list. They should inherit the shortended services list instead.

fultonj (approved):

PR #225, opened 2023-06-09
Added CR to configure network with nmstate

rebtoor:
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/249

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_nmstate.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_nmstate.yaml
@@ -1,0 +1,163 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNodeSet
+metadata:
+  name: openstack-edpm
+spec:
+  env:
+  - name: ANSIBLE_FORCE_COLOR
+    value: "True"
+  - name: ANSIBLE_ENABLE_TASK_DEBUGGER
+    value: "True"
+  services:
+  - bootstrap
+  - download-cache
+  - configure-network
+  - validate-network
+  - install-os
+  - configure-os
+  - run-os
+  - ovn
+  - neutron-metadata
+  - libvirt
+  - nova
+  - telemetry
+  preProvisioned: true
+  nodes:
+    edpm-compute-0:
+      role: edpm-compute
+      hostName: edpm-compute-0
+      ansible:
+        ansibleHost: 192.168.122.100
+        ansibleVars:
+          ctlplane_ip: 192.168.122.100
+          internal_api_ip: 172.17.0.100
+          storage_ip: 172.18.0.100
+          tenant_ip: 172.19.0.100
+          fqdn_internal_api: edpm-compute-0.example.com
+  networkAttachments:
+  - ctlplane
+  nodeTemplate:
+    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+    managementNetwork: ctlplane
+    ansible:
+      ansibleUser: cloud-admin
+      ansiblePort: 22
+      ansibleVars:
+        service_net_map:
+          nova_api_network: internal_api
+          nova_libvirt_network: internal_api
+        edpm_chrony_ntp_servers:
+        - pool.ntp.org
+        # edpm_network_config
+        # Default nic config template for a EDPM compute node
+        # These vars are edpm_network_config role vars
+        edpm_network_config_hide_sensitive_logs: false
+        edpm_network_config_template: |
+          ---
+          {% set mtu_list = [ctlplane_mtu] %}
+          {% for network in role_networks %}
+          {{ mtu_list.append(lookup('vars', networks_lower[network] ~ '_mtu')) }}
+          {%- endfor %}
+          {% set min_viable_mtu = mtu_list | max %}
+          dns-resolver:
+            config:
+              search: {{ dns_search_domains }}
+              server: {{ ctlplane_dns_nameservers }}
+          interfaces:
+            - name: {{ neutron_public_interface_name }}
+              type: ethernet
+              state: up
+            - name: {{ neutron_physical_bridge_name }}
+              type: ovs-interface
+              state: up
+              mtu: {{ min_viable_mtu }}
+              ipv4:
+                enabled: true
+                address:
+                  - ip: {{ ctlplane_ip }}
+                    prefix-length: {{ ctlplane_subnet_cidr }}
+          {% for network in role_networks %}
+            - name: {{ "vlan" ~ lookup('vars', networks_lower[network] ~ '_vlan_id') }}
+              type: ovs-interface
+              state: up
+              mtu: {{ lookup('vars', networks_lower[network] ~ '_mtu') }}
+              ipv4:
+                enabled: true
+                address:
+                  - ip: {{ lookup('vars', networks_lower[network] ~ '_ip') }}
+                    prefix-length: {{ lookup('vars', networks_lower[network] ~ '_cidr') }}
+          {% endfor %}
+            - name: {{ neutron_physical_bridge_name }}
+              type: ovs-bridge
+              bridge:
+                options:
+                  fail-mode: standalone
+                port:
+                - name: {{ neutron_public_interface_name }}
+                - name: {{ neutron_physical_bridge_name }}
+          {% for network in role_networks %}
+                - name: {{ "vlan" ~ lookup('vars', networks_lower[network] ~ '_vlan_id') }}
+                  vlan:
+                    mode: access
+                    tag: {{ lookup('vars', networks_lower[network] ~ '_vlan_id') }}
+          {% endfor %}
+          routes:
+            config:
+            - destination: {{ ctlplane_host_routes.ip_netmask }}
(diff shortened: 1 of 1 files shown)


rebtoor:
recheck

rebtoor:
recheck

rebtoor:
recheck

rebtoor:
recheck

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
```diff
@@ -0,0 +1,266 @@
...
+          edpm_sshd_configure_firewall: true
+          edpm_sshd_allowed_ranges: ['192.168.122.0/24']
+          # SELinux module
+          edpm_selinux_mode: enforcing
+          edpm_hosts_entries_undercloud_hosts_entries: []
+          # edpm_hosts_entries role
+          edpm_hosts_entries_extra_hosts_entries:
+          - 172.17.0.80 glance-internal.openstack.svc neutron-internal.openstack.svc cinder-internal.openstack.svc nova-internal.openstack.svc placement-internal.openstack.svc keystone-internal.openstack.svc
+          - 172.17.0.85 rabbitmq.openstack.svc
+          - 172.17.0.86 rabbitmq-cell1.openstack.svc
```
stuggi, line 261: when using install_yamls, dns is now deployed and available per default on 192.168.122.80 [1]. the internal api endpoints can be resolved using dns instead of adding those hosts entries. Should we change this PR to use 192.168.122.80 as the `ctlplane_dns_nameservers` and remove adding these hosts entries?

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/config/samples/network_v1beta1_dnsmasq.yaml

rebtoor: It sounds good! I haven't thought about that, let me fix the CR.

rebtoor: @stuggi I'll add the dns as `network_state.dns_resolver.config.server` for each compute node host vars since `ctlplane_dns_nameservers` is used by os-net-config.

rebtoor:
recheck

rebtoor:
recheck

fultonj (commented):
+2

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
```diff
@@ -0,0 +1,258 @@
...
+          edpm_nodes_validation_validate_gateway_icmp: false
+
+          edpm_ovn_metadata_agent_DEFAULT_transport_url: rabbit://default_user@rabbitmq.openstack.svc:5672
+          edpm_ovn_metadata_agent_metadata_agent_ovn_ovn_sb_connection: tcp:10.217.5.121:6642
+          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_nova_metadata_host: 127.0.0.1
+          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_metadata_proxy_shared_secret: 12345678
+          edpm_ovn_metadata_agent_DEFAULT_bind_host: 127.0.0.1
+          edpm_chrony_ntp_servers:
+          - clock.redhat.com
+          - clock2.redhat.com
```
stuggi, line 229: should those be public ones for the sample?

slagle: Probably, but this is what we have in our other samples, so I think we can clean it up across the board separately.

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
```diff
@@ -0,0 +1,258 @@
...
+                type: ethernet
+                state: up
+              - name: br-ex
+                type: ovs-interface
+                state: up
+                mtu: 1500
+                ipv4:
+                  enabled: true
+                  address:
+                    - ip: "192.168.122.100"
```
rabi, line 38: I think we would need templates for network_state where ansible vars are jinja substituted like we do for os-net-config[1].  After the IPAM patches merge user would know these IPs, vlans etc beforehand.

[1]  https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_network_config/templates/single_nic_vlans/single_nic_vlans.j2

rabi (commented):
As an example this could be ok, but we would need templates like we have for os-net-config.

rabi (changes requested):
Why is this in the PR?

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
rebtoor: Indeed this is an example. A lot of things could(should!) be improved, but it wasn't the purpose of this PR since the beginning.

BTW, It was my first thought having an approach similiar to actual os-net-config templates, but that means that we need to "wrap" the `network` role and that is something that @slagle would like to avoid.

We will think about it after the patches will be merged though. :)

rabi: not sure I understand what you mean by "wrap the network role". We need the templates so that ansible vars for IPs etc could be substituted when using IPAM.

rebtoor:
recheck

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
rebtoor: Sorry, I'll try to explain my point of view better.

in `edpm_network_config` role we rely on `linux_system_roles.network` [role](https://github.com/linux-system-roles/network) to configure interfaces via `nmstate`. 

We should add tasks like [these](https://github.com/openstack-k8s-operators/edpm-ansible/blob/23b28886268f8725c0772305c4ca5e7ef2a1aa67/roles/edpm_network_config/tasks/os_net_config.yml#L23-L43) to render template files that should be passed to the `network` role as variable but when i asked James about that he told me to call the role directly and users can/should configure the variables it provides directly.

I could also misunderstood his words BTW, I'm not against templating but IMHO this isn't the right PR for that.

bshephar: I think we should handle this similar to how we're working with os-net-config j2 files in the latest proposals here:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/243
https://github.com/openstack-k8s-operators/dataplane-operator/pull/284

These PR's will stop creating the `network` key in the inventory configmap and instead `copy` the yaml / jinja2 content directly to the nodes via the `edpm_network_config` role. We can either:

1. Re-use the same `spec.networkConfig.template` CR interface and handle it in Ansible;
2. Add a new key in the `networkConfig` struct specifically for nmstate configuration, but still render out j2 so that we can still populate it using data from IPAM. The same way we do with `os-net-config` today.

I think option one is easy enough, we can just have an additional variable to define it as a nmstate config vs os-net-config. But ultimately, nesting it under that `networkConfig` struct will be a cleaner user experience and provide more consistency. Provided we use Ansible's `copy` module the same way we do for `os-net-config`, then it will be fine with jinja2 and I think that should address the concern Rabi is raising in this last point.
https://github.com/openstack-k8s-operators/edpm-ansible/pull/243/files#diff-9add306dc693f43a54d6d0e77739b1fa2cefa165ddb6f2f55a0f65e6d7043bc7R29

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
```diff
@@ -0,0 +1,258 @@
...
+                next-hop-interface: br-ex
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+      deployStrategy:
+        deploy: false
+    edpm-compute-1:
+      role: edpm-compute
+      hostName: edpm-compute-1
+      ansibleHost: 192.168.122.101
+      node:
+        ansibleVars: |
```
bshephar, line 100: Do we need to add `edpm_network_config_nmstate: true` to the vars?
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_network_config/tasks/network_config.yml#L154

And yeah, just reiterating, but I think this would be cleaner if nested under `networkConfig`

config/samples/dataplane_v1beta1_openstackdataplane_nmstate.yaml
rebtoor: @bshephar the additional var is already there! :D `edpm_network_config_tool` variable can take the values `nmstate` or `os-net-config` (which is the default). That flag let the user choose between `linux-system-role.network` config or the "standard' os-net-config.

I'd like to nest the configuration under the `networkConfig` struct, but given that os-net-config developers have already introducted the support for nmstate, does it have sense to use a different kind of config under the same var?

rebtoor:
recheck

rebtoor:
recheck

slagle (commented):
I think we still want to merge this one as a sample, but it should be updated to reflect the changes in https://github.com/openstack-k8s-operators/dataplane-operator/pull/303

slagle:
@rebtoor will you be able to update this one? I think it would be good to still have as a sample.

rebtoor:
recheck

slagle (approved):

PR #234, opened 2023-06-13
Document bootstrap command for system registration

slagle:
- Rename composable-services.md to composable_services.md
- Add docs for most common configurations and system registration


fultonj (approved):
+2

PR #236, opened 2023-06-14
Create services as part of Role reconcile

slagle:
Create the default services (OpenStackDataPlaneService) resoures during
role reconcile. This will remove the manual step of the user having to
create the services. The docs are updated accordingly to remove that
step.


controllers/openstackdataplanerole_controller.go
```diff
@@ -138,6 +141,11 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		instance.Status.Conditions.MarkFalse(dataplanev1beta1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
 	}
 
+	err = r.EnsureServices(ctx, helper, instance)
```
fao89, line 144: 👍🏽

rebtoor (approved):

rabi (changes requested):

rebtoor:
recheck

slagle:
recheck

controllers/openstackdataplanerole_controller.go
```diff
@@ -294,6 +305,84 @@ func (r *OpenStackDataPlaneRoleReconciler) ReconcileBaremetalSet(ctx context.Con
...
+			"defaulting to ", servicesPath)
+	}
+
+	r.Log.Info("Ensuring services", "servicesPath", servicesPath)
+	services, err := os.ReadDir(servicesPath)
+	if err != nil {
+		return err
+	}
+
+	for _, service := range services {
```
rabi, line 332: It would be better to reconcile only the services that the role uses, have in the spec an not all.

controllers/openstackdataplanerole_controller.go
rabi: OK, if we're doing it elsewhere we can do it  here too, but sounds weird from the design pov.

controllers/openstackdataplanerole_controller.go
```diff
@@ -294,6 +305,84 @@ func (r *OpenStackDataPlaneRoleReconciler) ReconcileBaremetalSet(ctx context.Con
...
+		}
+
+		serviceObjSpec := &dataplanev1beta1.OpenStackDataPlaneServiceSpec{}
+		err = serviceObj.Spec.Decode(serviceObjSpec)
+		if err != nil {
+			r.Log.Info("Service Spec decode error")
+			return err
+		}
+
+		ensureService := &dataplanev1beta1.OpenStackDataPlaneService{
```
rabi, line 368: Are these service definitions CRs immutable? If someone wants to make changes in service definition by changing the file, they would only be patched with reconcile of an old role or a new role is created, right? And how and when these services would be cleaned up?

controllers/openstackdataplanerole_controller.go
slagle: Ok, I made that change.

controllers/openstackdataplanerole_controller.go
```diff
@@ -294,6 +305,95 @@ func (r *OpenStackDataPlaneRoleReconciler) ReconcileBaremetalSet(ctx context.Con
...
+
+		roleContainsService := false
+		for _, roleServiceName := range instance.Spec.Services {
+			if roleServiceName == serviceObjMeta.Name {
+				roleContainsService = true
+				break
+			}
+		}
+		if !roleContainsService {
+			r.Log.Info("Skipping ensure service since it is not a service on this role", "service", serviceObjMeta.Name)
```
rabi, line 369: probably missed continue here.

slagle: Indeed I did. thanks, fixed now.

slagle:
/retest

   * could not run steps: step dataplane-operator-build-deploy-kuttl failed: failed to wait for the created cluster claim to become ready: timed out waiting for the condition

fao89 (approved):

rabi (approved):
/lgtm

PR #240, opened 2023-06-16
Raise kuttl test timeout to 10m

slagle:
kuttl tests are timing out in CI[1]. Raise the timeout to 600s (10m)
which is reasonable for a resource constrained CI environment.


slagle:
this doesn'thave anything to do with the rdo job, so i think it's ok to merge.

PR #242, opened 2023-06-18
Use AEE instead of the Job to set status

bshephar:
This change uses the AnsibleEE object rather than the job to determine the status of the deployment. This ensures that we can still set the status even if the jobs are cleaned up after execution and prevents us from being stuck in an unknown state.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,11 @@
 repos:
 - repo: local
   hooks:
+    - id: gotidy
+      name: gotidy
+      language: system
+      entry: make
+      args: ["tidy"]
     - id: make-manifests
       name: make-manifests
       language: system
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -30,6 +30,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	novav1beta1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
+	ansibleeev1alpha1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -251,10 +252,12 @@ func ConditionalDeploy(
 		log.Info(fmt.Sprintf("Condition %s unknown", readyCondition))
 		return nil
 
-	} else if status.Conditions.IsFalse(readyCondition) {
-		ansibleEEJob, err := dataplaneutil.GetAnsibleExecutionJob(ctx, helper, obj, deployLabel)
+	}
+
+	if status.Conditions.IsFalse(readyCondition) {
+		ansibleEE, err := dataplaneutil.GetAnsibleExecution(ctx, helper, obj, deployLabel)
 		if err != nil && k8s_errors.IsNotFound(err) {
-			log.Info(fmt.Sprintf("%s OpenStackAnsibleEE Job not yet found", readyCondition))
+			log.Info(fmt.Sprintf("%s OpenStackAnsibleEE not yet found", readyCondition))
 			return nil
 		} else if err != nil {
 			log.Error(err, fmt.Sprintf("Error getting ansibleEE job for %s", deployName))
@@ -265,14 +268,29 @@ func ConditionalDeploy(
 				readyErrorMessage,
 				err.Error()))
 			return err
-		} else if ansibleEEJob.Status.Succeeded > 0 {
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusSucceeded {
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			status.Conditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
-		} else if ansibleEEJob.Status.Failed > 0 {
+			return nil
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusRunning || ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusPending {
+			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Status: %s", ansibleEE.Name, ansibleEE.Status.JobStatus))
+			status.Conditions.Set(condition.FalseCondition(
+				readyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				readyWaitingMessage))
+			return nil
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusFailed {
 			log.Info(fmt.Sprintf("Condition %s error", readyCondition))
-			err = fmt.Errorf("failed: job.name %s job.namespace %s", ansibleEEJob.Name, ansibleEEJob.Namespace)
+			err = fmt.Errorf("Execution.name %s Execution.namespace %s Execution.status.jobstatus: %s", ansibleEE.Name, ansibleEE.Namespace, ansibleEE.Status.JobStatus)
 			status.Conditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.ErrorReason,
@@ -280,9 +298,6 @@ func ConditionalDeploy(
 				readyErrorMessage,
 				err.Error()))
 			return err
-		} else {
-			log.Info(fmt.Sprintf("Condition %s not yet ready", readyCondition))
-			return nil
 		}
 
 	}
--- a/pkg/util/ansible_execution.go
+++ b/pkg/util/ansible_execution.go
@@ -22,11 +22,9 @@ import (
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
-	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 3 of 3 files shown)


pkg/deployment/deployment.go
```diff
@@ -33,6 +33,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
```
rabi, line 36: No need to define them here.. There are there in the API for ansibleee and exported https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/171/files

rabi (changes requested):
Just one change, but you've to wait for ansibleee bump here

api/go.sum
```diff
@@ -228,8 +228,7 @@ github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-202306130620
 github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-20230613062027-d886a7879256/go.mod h1:cAXLyeKp6TiLW8qongnw6WR63Jnsm8CeRC5y07z3TAY=
 github.com/openstack-k8s-operators/lib-common/modules/storage v0.0.0-20230613062027-d886a7879256 h1:08Kn1j5Eqn0Uma3rx7a1GgIz0ejeOau4AtvdNip1rSI=
 github.com/openstack-k8s-operators/lib-common/modules/storage v0.0.0-20230613062027-d886a7879256/go.mod h1:7sbo6DydOwpl8Ex1atTbXIrWYvZ++eSOJ0Z6RphJJ44=
-github.com/openstack-k8s-operators/openstack-ansibleee-operator/api v0.0.0-20230612175620-ea563267160e h1:0FPvGnqTdW9GRcuNk2uxG6ky0TwOX1qJPSpX26CMK58=
-github.com/openstack-k8s-operators/openstack-ansibleee-operator/api v0.0.0-20230612175620-ea563267160e/go.mod h1:4TO/S5RdOodpNH70bVTZ3xLonaXamngG/gYRkZd+mw8=
+github.com/openstack-k8s-operators/openstack-ansibleee-operator/api v0.0.0-20230619001113-d4bec53e04a9 h1:UNaryEltysLdCyA+BgXQkW9osSln77W9fkb1j9O/+vk=
```
rabi, line 231: This would fail with pre-commit I guess with the go.mod

rebtoor:
recheck

rebtoor:
+2 for me

.pre-commit-config.yaml
```diff
@@ -1,6 +1,11 @@
 repos:
 - repo: local
   hooks:
+    - id: gotidy
+      name: gotidy
+      language: system
+      entry: make
+      args: ["tidy"]
```
fao89, line 8: nice!

fao89 (approved):

.pre-commit-config.yaml
rebtoor: ++

bogdando:
relying on persistent AEE objects' statuses instead of "disposable jobs" those produce is a great idea. @gibizer something to check for the approaches we take in Nova operator in that regard

pkg/deployment/deployment.go
```diff
@@ -265,24 +268,33 @@ func ConditionalDeploy(
...
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusSucceeded {
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			status.Conditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
-		} else if ansibleEEJob.Status.Failed > 0 {
+			return err
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusRunning || ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusPending {
```
kajinamit, line 281: this apparently works though I tend to prefer using else if here to make this is the continuous logic from L273.

bshephar: I did this to simplify the logic. Since we return from the previous if statement if that condition is met:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/242/files#diff-b44566f9caa976421532eb049c77b7eb247418f138516fab277a1ec33d3a7992R278

We shouldn't be able to arrive here any other way, so it didn't seem like the else was necessary in this case.

pkg/deployment/deployment.go
```diff
@@ -265,24 +268,33 @@ func ConditionalDeploy(
...
-		} else if ansibleEEJob.Status.Succeeded > 0 {
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusSucceeded {
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			status.Conditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
-		} else if ansibleEEJob.Status.Failed > 0 {
+			return err
```
kajinamit, line 278: we probably want to return nil instead.

pkg/deployment/deployment.go
kajinamit: I get your point but in that case then we probably want to return nil directly after L 287 ? My concern is that if-else logics depending on status are partially continued. If we prefer separate blocks then we probably have to make all blocks separate.

bshephar: Yeah, I guess we can. There's only one remaining unhandled condition which would be:
`	JobStatusFailed = "Failed"`

I guess my thinking there was that  the else block would also capture anything else that I wasn't already handling. But since it's just one remaining status, I think we can remove that too.

bshephar: Updated this now.

pkg/deployment/deployment.go
bshephar: Done

pkg/deployment/deployment.go
```diff
@@ -265,24 +268,36 @@ func ConditionalDeploy(
...
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusRunning || ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusPending {
+			log.Info(fmt.Sprintf("AnsibleEE job is still running: Execution: %s, Status: %s", ansibleEE.Name, ansibleEE.Status.JobStatus))
+			status.Conditions.Set(condition.FalseCondition(
+				readyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				readyWaitingMessage))
+			return err
```
kajinamit, line 288: should we return nil ?

pkg/deployment/deployment.go
```diff
@@ -265,24 +268,36 @@ func ConditionalDeploy(
...
 			log.Info(fmt.Sprintf("Condition %s error", readyCondition))
-			err = fmt.Errorf("failed: job.name %s job.namespace %s", ansibleEEJob.Name, ansibleEEJob.Namespace)
+			err = fmt.Errorf("Execution.name %s Execution.namespace %s Execution.status.jobstatus: %s", ansibleEE.Name, ansibleEE.Namespace, ansibleEE.Status.JobStatus)
 			status.Conditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.ErrorReason,
 				condition.SeverityError,
 				readyErrorMessage,
 				err.Error()))
 			return err
```
kajinamit, line 300: ditto

pkg/deployment/deployment.go
```diff
@@ -265,24 +268,36 @@ func ConditionalDeploy(
...
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			status.Conditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
-		} else if ansibleEEJob.Status.Failed > 0 {
+			return nil
+		}
+
+		if ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusRunning || ansibleEE.Status.JobStatus == ansibleeev1alpha1.JobStatusPending {
+			log.Info(fmt.Sprintf("AnsibleEE job is still running: Execution: %s, Status: %s", ansibleEE.Name, ansibleEE.Status.JobStatus))
```
kajinamit, line 282: This might be a nit pick but JobStatusPending does not mean the job is running (registered by not yet executed) probably we should rephrase the error to `job is not yet completed` or should use a different message (`running` vs `pending`)

bshephar: Yeah, that's a good point. Would be less confusing if it said job is not yet completed.

pkg/deployment/deployment.go
kajinamit: Ignore this. We should return an error defined above :-(

bshephar:
Kuttl failed because of:
```
Trying to pull quay.io/openstack-k8s-operators/infra-operator-bundle:8503e53378d606f739207da5c845e826f66831f8...
error: build error: error creating build container: 4 errors occurred while pulling:
 * initializing source docker://registry.redhat.io/golang:1.19: reading manifest 1.19 in registry.redhat.io/golang: unknown: Not Found
 * initializing source docker://registry.access.redhat.com/golang:1.19: reading manifest 1.19 in registry.access.redhat.com/golang: name unknown: Repo not found
 * initializing source docker://quay.io/golang:1.19: reading manifest 1.19 in quay.io/golang: StatusCode: 404, <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final/...
 * reading blob sha256:67f4091d98eedad4cde8b1cd42b2040fb0bbff6119f91b70bb92fa640798ed11: Get "https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/67/67f4091d98eedad4cde8b1cd42b2040fb0bbff6119f91b70bb92fa640798ed11/data?verify=1687483061-d1B7lqdTxX52cnXLAioWkinIsqs%3D": remote error: tls: handshake failure
```

rebtoor (approved):

rebtoor:
recheck

bshephar:
I guess the `|` breaks the `sed` command here:
```
f64908bbab384226aa25efce48140c594b1e652c|'
sed: -e expression #1, char 172: unterminated `s' command
```

bshephar:
recheck

fao89:
recheck

api/go.sum
```diff
@@ -234,6 +235,18 @@ github.com/openstack-k8s-operators/openstack-ansibleee-operator/api v0.0.0-20230
 github.com/openstack-k8s-operators/openstack-ansibleee-operator/api v0.0.0-20230620085730-fddf65f65f88/go.mod h1:e4hw70sdvXG8anfBSGq5VMq11h52aN/TI3HYxjPyDGE=
 github.com/openstack-k8s-operators/openstack-baremetal-operator/api v0.0.0-20230626134559-c08170fb5cff h1:ui7RJZdA9SedC1joPNfvkNQ/SeF5LSJmlTyfJhzaDxY=
 github.com/openstack-k8s-operators/openstack-baremetal-operator/api v0.0.0-20230626134559-c08170fb5cff/go.mod h1:JTCJLOYuhAr5GZiPzctpcOTh3IvqJn6RvvbiTMRoUQ4=
+=======
```
rabi: Can you please squash the commits as they would be committed as separate ones. Having merge conflicts in  commit like this is not good.

fao89:
recheck
/lgtm

rebtoor:
recheck

PR #246, opened 2023-06-23
Cleanup configmaps after node/role cleanup

rabi:
This also changes configmap management to use lib-common api.

### DIFF
--- a/pkg/deployment/inventory.go
+++ b/pkg/deployment/inventory.go
@@ -25,13 +25,11 @@ import (
 	"strings"
 
 	yaml "gopkg.in/yaml.v3"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	dataplanev1beta1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/ansible"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	utils "github.com/openstack-k8s-operators/lib-common/modules/common/util"
 )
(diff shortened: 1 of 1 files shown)


pkg/deployment/inventory.go
```diff
@@ -99,40 +97,29 @@ func GenerateRoleInventory(ctx context.Context, helper *helper.Helper,
...
+		"network":   string(instance.Spec.NodeTemplate.NetworkConfig.Template),
+	}
 	configMapName := fmt.Sprintf("dataplanerole-%s", instance.Name)
-	cm := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      configMapName,
-			Namespace: instance.Namespace,
-			Labels:    instance.ObjectMeta.Labels,
+	cms := []utils.Template{
+		// ConfigMap
```
bshephar, line 111: Not sure of the purpose of this comment? Should it be more explicit about which configmap is being created? Is the comment needed at all?

pkg/deployment/inventory.go
```diff
@@ -184,40 +171,29 @@ func GenerateNodeInventory(ctx context.Context, helper *helper.Helper,
...
+		"network":   string(networkConfig.Template),
+	}
 	configMapName := fmt.Sprintf("dataplanenode-%s", instance.Name)
-	cm := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      configMapName,
-			Namespace: instance.Namespace,
-			Labels:    instance.ObjectMeta.Labels,
+	cms := []utils.Template{
+		// ConfigMap
```
bshephar, line 185: And this one. Do we need a comment?

bshephar (commented):
Nice idea to use the common interface. Seems to ensure that we set the `ControllerReference`.

pkg/deployment/inventory.go
rabi: Yes there is a puprose. utils.Template can be used for configmaps and secrets and this says it's a configmap[1].

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/controllers/memcached/memcached_controller.go#L248

pkg/deployment/inventory.go
rabi: as explained above.

rebtoor (commented):
+2

fao89 (approved):

PR #251, opened 2023-06-27
Bump growvols_args: /var to 80%

raukadah:
https://github.com/openstack-k8s-operators/install_yamls/pull/350 reduces the growvols_args /var to 80% to fix swap issue.

The same needs to be updated in default sample config.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplane_baremetal.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane_baremetal.yaml
@@ -64,7 +64,7 @@ spec:
             nova_libvirt_network: internal_api
           edpm_chrony_ntp_servers:
             - clock.redhat.com
-          growvols_args: '/var=100%'
+          growvols_args: '/var=80%'
           # edpm_network_config
           # Default nic config template for a EDPM compute node
           # These vars are edpm_network_config role vars


config/samples/dataplane_v1beta1_openstackdataplane_baremetal.yaml
```diff
@@ -64,7 +64,7 @@ spec:
             nova_libvirt_network: internal_api
           edpm_chrony_ntp_servers:
             - clock.redhat.com
-          growvols_args: '/var=100%'
+          growvols_args: '/var=80%'
```
rabi, line 67: Let's just remove it from here and fix the default in edpm-ansible role. We don't need to do it in the sample now that we kustomize in install_yamls and the current tuning is our CI specific.

rebtoor: 100% agree with @rabi here. Let me propose the pr.

rebtoor: Opened https://github.com/openstack-k8s-operators/edpm-ansible/pull/200 but slightly changed approach.

rebtoor:
probably we can close this one too...

@raukadah  @rabi

rabi:
I had suggested earlier to drop that line and use the default.

PR #252, opened 2023-06-28
Make AnsibleSSHPrivateKeySecret validation Optional

rabi:
It can be specified either for the nodes or in `nodeTemplate` of the role. Keeping it `Required` means we have to specify it at all the places.

Let's make this optional. We can add a validationwebhook later to verify either or both can be specified for role/node combination.

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodes.yaml
@@ -935,8 +935,6 @@ spec:
                         type: string
                     type: object
                     x-kubernetes-map-type: atomic
-                required:
-                - ansibleSSHPrivateKeySecret
                 type: object
               role:
                 type: string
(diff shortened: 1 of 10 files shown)


kajinamit:
This makes AnsibleSSHPrivateKeySecret in all layers optional, but I'm wondering if we want to keep it mandatory in some point (probably in nodeTemplate). At least we need a better validation to catch the case where user does not give any value at all layers.

rabi:
> This makes AnsibleSSHPrivateKeySecret in all layers optional, but I'm wondering if we want to keep it mandatory in some point (probably in nodeTemplate). At least we need a better validation to catch the case where user does not give any value at all layers.

It can be provided for nodes or at the role level (nodeTemplate) with node overriding the role. We can't make one required. I mentioned that we will check that in a validationwebhook like we do in other operators. 

Also, current CR design uses the same structure at both places.

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/api/v1beta1/openstackbaremetalset_webhook.go#L171-L177

kajinamit:
/lgtm

ok that sounds reasonble

PR #254, opened 2023-06-29
Use HostName in BaremetalSet Spec

rabi:
node.Name and node.Spec.HostName could be different with HostName
being the fqdn. This changes to use HostName consistently.

This fixes a few additional things:

- Removes network from fqdn as we seem to define dnsDomain
for every network which has to be unique.
- Sets ansible_host to ctlplane ip as the ansible runner pod
  currently can't resolve the hostnames from the DNS Server.

### DIFF
--- a/pkg/deployment/baremetal.go
+++ b/pkg/deployment/baremetal.go
@@ -49,7 +49,11 @@ func DeployBaremetalSet(
 		instance.Spec.BaremetalSetTemplate.DeepCopyInto(&baremetalSet.Spec)
 		for _, node := range nodes.Items {
 			ipSet, ok := ipSets[node.Name]
-			instanceSpec := baremetalSet.Spec.BaremetalHosts[node.Name]
+			hostName := node.Spec.HostName
+			if hostName == "" {
+				hostName = node.Name
+			}
+			instanceSpec := baremetalSet.Spec.BaremetalHosts[hostName]
 			if !ok {
 				utils.LogForObject(helper, "IPAM Not configured for use, skipping", instance)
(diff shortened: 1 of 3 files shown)


rabi:
recheck

pkg/deployment/baremetal.go
```diff
@@ -49,7 +49,7 @@ func DeployBaremetalSet(
 		instance.Spec.BaremetalSetTemplate.DeepCopyInto(&baremetalSet.Spec)
 		for _, node := range nodes.Items {
 			ipSet, ok := ipSets[node.Name]
-			instanceSpec := baremetalSet.Spec.BaremetalHosts[node.Name]
+			instanceSpec := baremetalSet.Spec.BaremetalHosts[node.Spec.HostName]
```
stuggi, line 52: I am not into the details of the dataplane-operator. The HostName is an optional parameter [1], should it be required now?

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanenode_types.go#L32

rabi: oops.. I thot is is required field, but I see at many places we do[1]. Though we can do it in the webhook defaulting, I'll add the same here for the time being.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/ipam.go#L78C1-L82C5

rebtoor (approved):

rabi:
recheck

rabi:
Hit RETRY_LIMIT, will give one more try.

rabi:
recheck

rebtoor:
recheck

PR #260, opened 2023-07-02
Drop edpm_host_entries usage

rabi:
As we use internal DNS service, let's drop the usage of the role.


bshephar (approved):
Sounds good to me. If we have a DNS service providing this functionality it's nice to remove manual `/etc/hosts` management.

We should probably also update our documentation to inform folks that we expect DNS resolution to happen via the `dnsmasq` resource. But we can do that in another PR.

PR #262, opened 2023-07-02
Add defaulting and validation webhook

rabi:
Rather than doing validations and defaulting in reconcilers we would now do them in the webhooks.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/367


rabi:
recheck

rabi:
recheck

api/v1beta1/openstackdataplane_webhook.go
```diff
@@ -0,0 +1,125 @@
...
+var _ webhook.Defaulter = &OpenStackDataPlane{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *OpenStackDataPlane) Default() {
+	openstackdataplanelog.Info("default", "name", r.Name)
+	r.Spec.Default()
+
+}
+
+// Default - set defaults for this Glance spec
```
kajinamit, line 56: this OpenStackDataplane Spec

api/v1beta1/openstackdataplane_webhook.go
```diff
@@ -0,0 +1,125 @@
...
+func (spec *OpenStackDataPlaneSpec) Default() {
+	for nodeName, node := range spec.Nodes {
+		if node.HostName == "" {
+			node.HostName = nodeName
+		}
+		if node.Node.AnsibleSSHPrivateKeySecret == "" {
```
kajinamit, line 62: So this copies the value to node level when the node is initially added, and updating the key reference later requires updating the field in each node.

I'm ok with this but we might want to revisit this later if we consider supporting key rotation for all nodes(though that can be done by easy sed replacement, I guess)

rabi:
recheck

api/v1beta1/openstackdataplane_webhook.go
rabi: done

rabi:
recheck

rabi:
recheck

rabi:
recheck

rabi:
recheck

api/v1beta1/openstackdataplane_webhook.go
kajinamit: yes I understand your point though it might be a bit confusing that the result can be different according to the API.

This would not block the change. We can discuss this further if we want. So I'll close this conversation now.

rabi: if it's about using  PATCH to update the ssh keys as an use-case, we can probably implement PATCH for the webhook at some point.

rabi:
recheck

PR #264, opened 2023-07-04
Allow to set custom runner image for default services

karelyatin:
With [1] default Services are automatically created as part of role reconcile and since then the only way to use custom runner image for these default Services is by patching openstack-ansibleee-operator csv and that applies default settings globally.

This patch allows the default services to use custom runner image per role.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/236

### DIFF
--- a/pkg/deployment/service.go
+++ b/pkg/deployment/service.go
@@ -148,6 +148,9 @@ func EnsureServices(ctx context.Context, helper *helper.Helper, instance *datapl
 			},
 		}
 		_, err = controllerutil.CreateOrPatch(ctx, helper.GetClient(), ensureService, func() error {
+			if len(instance.Spec.OpenStackAnsibleEERunnerImage) > 0 {
+				serviceObjSpec.OpenStackAnsibleEERunnerImage = instance.Spec.OpenStackAnsibleEERunnerImage
+			}
 			ensureService.Spec = *serviceObjSpec
 			return nil
 		})
--- a/api/bases/dataplane.openstack.org_openstackdataplaneroles.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplaneroles.yaml
@@ -1086,6 +1086,8 @@ spec:
                     type: object
                     x-kubernetes-map-type: atomic
                 type: object
+              openStackAnsibleEERunnerImage:
+                type: string
               preProvisioned:
                 type: boolean
               services:
--- a/api/bases/dataplane.openstack.org_openstackdataplanes.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanes.yaml
@@ -2010,6 +2010,8 @@ spec:
                           type: object
                           x-kubernetes-map-type: atomic
                       type: object
+                    openStackAnsibleEERunnerImage:
+                      type: string
                     preProvisioned:
                       type: boolean
                     services:
--- a/api/v1beta1/openstackdataplanerole_types.go
+++ b/api/v1beta1/openstackdataplanerole_types.go
@@ -63,6 +63,10 @@ type OpenStackDataPlaneRoleSpec struct {
 	// +kubebuilder:default={configure-network,validate-network,install-os,configure-os,run-os}
 	// Services list
 	Services []string `json:"services"`
+
+	// +kubebuilder:validation:Optional
+	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
+	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty"`
 }
 
 //+kubebuilder:object:root=true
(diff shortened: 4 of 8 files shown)


pkg/deployment/service.go
```diff
@@ -148,6 +148,9 @@ func EnsureServices(ctx context.Context, helper *helper.Helper, instance *datapl
 			},
 		}
 		_, err = controllerutil.CreateOrPatch(ctx, helper.GetClient(), ensureService, func() error {
+			if len(instance.Spec.OpenStackAnsibleEERunnerImage) > 0 {
+				serviceObjSpec.OpenStackAnsibleEERunnerImage = instance.Spec.OpenStackAnsibleEERunnerImage
```
rabi, line 152: We probably can't do that. If provided for a role it would override for all services including custom services which may need a different image than the standard services.  This is tricky. My initial PR[1] was doing the samething (though checking if service has a runner image or not and then using the one provided with role).

[1] https://github.com/openstack-k8s-operators/dataplane-operator/commit/0c2050069baf11cfbdf712f9550c1164f4df488a

karelyatin: Thanks @rabi , i was not aware about your patch but that looks quite different.

But here with this patch i see it will not impact custom services(L91-96), the same i verfied in my local tests using different image for default and custom service(repo-setup)

rabi: My patch was before the role started reconciling the services. 

- I this patch, if you have OpenStackAnsibleEERunnerImage in the role specit would override for all services including the custom service(which we don't want). We won't be able to have different images for different services used by the role. 
-  Also, service CRs are global and not per role. So every role reconcile would try to update them and impact the services that don't want it to be overridden at the role level.

rabi: > But with this patch i was able to use different images for custom service(ex. repo-setup created by install_yamls)

Yeah, you're right, it's working because that service is not in location $OPERATOR_SERVICES[1] and not reconciled by the role.  I was thinking that ENV var can be patched in the CSV to point to a location where all services (default+custom) would be there, but looks like it's used when building the operator image[2]. So it looks like custom service creation would be manual and would not be created by patching $OPERATOR_SERVICES.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/service.go#L81
[2] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/Dockerfile#L64

karelyatin: Thanks yes that's too my understanding. @slagle may be you can check/confirm this?

slagle: I'm not a fan of adding the Image field back to the role, and having that override the default services. That would affect the services for all roles (as pointed out).

I don't see what the problem is with editing the CSV? Is that not the point of the environment variables, so that we can configure things for different environments (dev, CI, etc)?

I'm not sure how CI builds/deploys the CSV's, but perhaps we could plug in there somewhere?

Another option would be to change the service reconcile logic (EnsureServices in service.go), to not only ensure a service exists, but to not ensure it's full spec. On the first reconcile, it would create the service and it's full spec. On any subsequent reconcile if the service is already found, it would not modify any fields on the spec. That would allow for setting the images on existing services in a way that it does not get overwritten.

karelyatin: > I'm not a fan of adding the Image field back to the role, and having that override the default services. That would affect the services for all roles (as pointed out).
> 
Actually i did add Image field in role as i didn't see other option to do it. And wanted to have opinions from others who have more understanding around it.

> I don't see what the problem is with editing the CSV? Is that not the point of the environment variables, so that we can configure things for different environments (dev, CI, etc)?

Yes CI is patching the openstack-ansibleee CSV as of now.
Actually for csv we would need to patch openstack-ansibleee csv and it looked wrong to modify openstack-ansibleee csv(as i believe it could be used independently of CRs from dataplane-operator) just to provide custom AnsibleEE image to default services for "DataplaneRoles". But if it's just me who thought this way then i can live with the patching approach.
> 
> I'm not sure how CI builds/deploys the CSV's, but perhaps we could plug in there somewhere?
> 
It's modified as part of https://github.com/openstack-k8s-operators/ci-framework/pull/339

> Another option would be to change the service reconcile logic (EnsureServices in service.go), to not only ensure a service exists, but to not ensure it's full spec. On the first reconcile, it would create the service and it's full spec. On any subsequent reconcile if the service is already found, it would not modify any fields on the spec. That would allow for setting the images on existing services in a way that it does not get overwritten.

Not fully understood this but yes sounds good if that's the only option. May be the idea here is to not allow custom services with the same name as default services and if one creates those will be overridden with the default spec in config/services. But not sure how it will differentiate with the services pre created(before role create) and default services created in first reconcile as part of EnsureServices.

slagle: > Yes CI is patching the openstack-ansibleee CSV as of now. Actually for csv we would need to patch openstack-ansibleee csv and it looked wrong to modify openstack-ansibleee csv(as i believe it could be used independently of CRs from dataplane-operator) just to provide custom AnsibleEE image to default services for "DataplaneRoles". But if it's just me who thought this way then i can live with the patching approach.
> 
> > I'm not sure how CI builds/deploys the CSV's, but perhaps we could plug in there somewhere?
> 
> It's modified as part of [openstack-k8s-operators/ci-framework#339](https://github.com/openstack-k8s-operators/ci-framework/pull/339)

Along these lines though, how does CI modify the env vars it needs for other types of tests? For example, suppose we need to run with a different keystone api image, which is specified like so in the keystone-operator csv:

`                env:
                - name: KEYSTONE_API_IMAGE_URL_DEFAULT
                  value: quay.io/podified-antelope-centos9/openstack-keystone:current-podified
`

How would CI do that? And can we make use of the same mechanism to modify the env var for the default runner image?

raukadah: I have used the above role to update the ansiblee runner image via this pr https://github.com/openstack-k8s-operators/edpm-ansible/pull/210 and here are the results: https://github.com/openstack-k8s-operators/edpm-ansible/pull/210#issuecomment-1628513234 please have a look, thank you :-)

slagle:
Can we drop this PR now? I don't think adding the image back to the role is the right approach.

PR #267, opened 2023-07-04
Removing imports of tasks from main, as that is the default behavior

jpodivin:
The import_role module uses "main" as a default for the tasks_from parameter. Therefore there is not need to explicitly request it.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_role_module.html#parameter-tasks_from


fao89 (approved):

jpodivin:
recheck

rebtoor (approved):

bshephar (approved):
Yeah, sounds good.

jpodivin:
Kuttl tests are failing with something that seems to be a genuine error.

jpodivin:
I think I found out what was the problem. Kuttle tests were still expecting the explicit main.yml import.

rebtoor (approved):

PR #273, opened 2023-07-06
Adding netlify for previewing docs

fao89:


odyssey4me:
It took me a little time to figure out where to click, but nice to see that we can get a preview. Is there a way we can use one of our existing CI runners (maybe zuul) to run and produce the same?

fao89:
recheck

fao89:
recheck

fao89:
recheck

netlify.toml
```diff
@@ -0,0 +1,17 @@
+[build]
```
raukadah, line 1: Please add `netlify.toml` under zuul job irrelevant-files list here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/zuul.d/jobs.yaml#L11 so that zuul jobs does not trigger on this change.

raukadah (changes requested):

fao89:
recheck

rebtoor (commented):
+2

raukadah:
recheck

fao89:
closing in favor of https://github.com/openstack-k8s-operators/dataplane-operator/pull/282

PR #274, opened 2023-07-07
Wait for DNSMasq to be ready and reconcile for DNSMasq changes

rabi:
If dnsmasq service exists and not ready, we should wait for it. Also, if there are any changes in DNSMasq we should reconcile. This won't normally happen as the service would be created by the controlplane and won't be changed.

But it's posible that DNSMasq service has been deleted and role has to be reconciled once it's reconciled.

### DIFF
--- a/pkg/deployment/ipam.go
+++ b/pkg/deployment/ipam.go
@@ -136,8 +136,11 @@ func EnsureDNSData(ctx context.Context, helper *helper.Helper,
 		return nil, "", false, err
 	}
 	if len(dnsmasqList.Items) == 0 {
-		util.LogForObject(helper, "No NetConfig CR exists yet, DNS Service won't be used", instance)
+		util.LogForObject(helper, "No DNSMasq CR exists yet, DNS Service won't be used", instance)
 		return nil, "", true, nil
+	} else if !dnsmasqList.Items[0].IsReady() {
+		util.LogForObject(helper, "DNSMasq service exists, but not ready yet ", instance)
+		return nil, "", false, nil
 	}
 
 	// Create or Patch DNSData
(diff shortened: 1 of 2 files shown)


pkg/deployment/ipam.go
```diff
@@ -136,8 +136,11 @@ func EnsureDNSData(ctx context.Context, helper *helper.Helper,
 		return nil, "", false, err
 	}
 	if len(dnsmasqList.Items) == 0 {
-		util.LogForObject(helper, "No NetConfig CR exists yet, DNS Service won't be used", instance)
+		util.LogForObject(helper, "No DNSMasq CR exists yet, DNS Service won't be used", instance)
 		return nil, "", true, nil
+	} else if !dnsmasqList.Items[0].IsReady() {
+		util.LogForObject(helper, "DNSMasq service exists, but not ready yet ", instance)
+		return nil, "", false, nil
```
fao89, line 143: sweet!

fao89 (approved):

pkg/deployment/ipam.go
```diff
@@ -136,8 +136,11 @@ func EnsureDNSData(ctx context.Context, helper *helper.Helper,
 		return nil, "", false, err
 	}
 	if len(dnsmasqList.Items) == 0 {
-		util.LogForObject(helper, "No NetConfig CR exists yet, DNS Service won't be used", instance)
+		util.LogForObject(helper, "No DNSMasq CR exists yet, DNS Service won't be used", instance)
 		return nil, "", true, nil
+	} else if !dnsmasqList.Items[0].IsReady() {
+		util.LogForObject(helper, "DNSMasq service exists, but not ready yet ", instance)
```
abays, line 142: nit: `s/but not ready yet "/but not ready yet"`

PR #275, opened 2023-07-07
Update Role sample for networkConfig template

bshephar:
We changed the way we handle networkConfig to allow for users to provide a jinja2 formatted template in the spec:
https://github.com/openstack-k8s-operators/dataplane-operator/commit/5c635f6e18b22455887c38661a46c8db15d096a8

This change is updating the role sample file to correct the usage of the networkConfig.template section.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanerole.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanerole.yaml
@@ -11,7 +11,51 @@ spec:
     - run-os
   nodeTemplate:
     networkConfig:
-      template: templates/net_config_bridge.j2
+      template: |
+        ---
+        network_config:
+        - type: interface
+          name: nic2
+          mtu: 1500
+          addresses:
+            - ip_netmask:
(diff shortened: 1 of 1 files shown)


fao89:
recheck

fao89 (approved):

config/samples/dataplane_v1beta1_openstackdataplanerole.yaml
```diff
@@ -11,7 +11,51 @@ spec:
     - run-os
   nodeTemplate:
     networkConfig:
```
rabi, line 13: Looking at the patch linked in this PR , I noticed that we're creating a file in runner pod and then using it. I noticed that empty file in the runner pod the other day. Why're doing all this rather than just converting the yaml blob to json and set  `edpm_network_config_override` ansibe var[1] that ansible would have taken care of.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_network_config/tasks/os_net_config.yml#L26-L34

bshephar:
recheck

PR #278, opened 2023-07-10
Add finalizers for controllers

bshephar:
None of the controllers currently use finalizers. This changes adds the finalizers and handling required for them within the deletion process.

### DIFF
--- a/controllers/openstackdataplane_controller.go
+++ b/controllers/openstackdataplane_controller.go
@@ -115,6 +115,13 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 	}()
 
+	// If the service object doesn't have our finalizer, add it.
+	controllerutil.AddFinalizer(instance, helper.GetFinalizer())
+	// Register the finalizer immediately to avoid orphaning resources on delete
+	if err := r.Update(ctx, instance); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	if instance.Status.Conditions == nil {
 		instance.InitConditions()
 		// Register overall status immediately to have an early feedback e.g. in the cli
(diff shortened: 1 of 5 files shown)


controllers/openstackdataplane_controller.go
```diff
@@ -115,6 +115,13 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 	}()
 
+	// If the service object doesn't have our finalizer, add it.
+	controllerutil.AddFinalizer(instance, helper.GetFinalizer())
```
rabi, line 119: Nah, we don't need to add finalizer unless we want to do some cleanup manually in delete reconcile. All resources with ownerref/controller reference are anyway cleaned up. You can see an example of how we removed it from controllplane controller[1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/commit/e6fc9b5951c70e498b10f328b0d7a2e985351f62

rabi (changes requested):

controllers/openstackdataplane_controller.go
bshephar: We didn't want to append a finalizer to the `NovaExternalCompute`, Baremetal, ipset, objects that we create? I assumed these would be where finalizers would be leverage by the DP operator

I just saw all of the finalizer `rbac` comments that exist today and assumed it was all copied without the implementation for finalizers. But if we don't want to use them, then I'll send a PR to remove the remainder of the finalizer `rbac` references. eg:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplane_controller.go#L55

rabi: > We didn't want to append a finalizer to the NovaExternalCompute, Baremetal, ipset, objects that we create?

Add finalizers for what? All these CRs are owned by dataplanerole/node and would be deleted when the owner is deleted. If those are deleted they would again be created by role/node.

bshephar: If a user creates the OpenStackDataPlane CR with references to nodes and roles, then we creates the CRs for OpenStackDataPlane{Node,Role}. We don’t want a user to delete the Node and Role that are managed by our controllers, since the recreation of either of them would trigger another Ansible execution. 

It ties in to the redeployProtection work. But it might not matter after our discussion about the overall CRD design anyway.

rabi: At present they're owned by dataplane[1], once we move to a different CRD design we can change if required. The bottom line is, finalizer is added if you want to do something that won't be done automatically. Also, as discussed yesterday I don't see why we would need dataplane CR at all.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplane_controller.go#L386

PR #280, opened 2023-07-10
Fix kuttl tests

fao89:


tests/kuttl/tests/dataplane-deploy-no-nodes-test/01-assert.yaml
```diff
@@ -23,589 +23,3 @@ status:
...
-  JobStatus: Succeeded
-  conditions:
-  - message: AnsibleExecutionJob complete
-    reason: Ready
-    status: "True"
-    type: Ready
-  - message: AnsibleExecutionJob complete
-    reason: Ready
-    status: "True"
-    type: AnsibleExecutionJobReady
```
slagle, line 611: why are you removing all of this? Can you explain in the commit message what caused the change and why this would fix it?

fao89:
https://github.com/openstack-k8s-operators/install_yamls/pull/397#issuecomment-1631109438

PR #281, opened 2023-07-10
CSV: Set operator-type: non-standalone

dprince:
This should hide the dataplane package/bundle in the OpenShift console as we only want the main OpenStack operator to show up there.

### DIFF
--- a/config/manifests/bases/dataplane-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/dataplane-operator.clusterserviceversion.yaml
@@ -4,6 +4,7 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
+    operators.operatorframework.io/operator-type: non-standalone
   name: dataplane-operator.v0.0.0
   namespace: placeholder
 spec:


abays (approved):
/lgtm

bshephar (changes requested):
I'm not too sure about doing it here. There's a larger discussion around architecture for this operator happening at the moment. I won't link to a RH internal Jira here, but I think our users should be able to click on the Operator in "Installed Operators" and create their `OpenStackDataPlane` CR to define their compute infrastructure.

If this switch hides the operator from the console, does that mean they wouldn't be able to do this in that case? If so, I would be inclined to skip this change on the dataplane operator. Which would give the users a `OpenStackControlPlane` and a `OpenStackDataPlane` operator in their consoles to define the environment.

Please let me know if I'm misunderstanding this switch.

fao89:
recheck

slagle:
> I'm not too sure about doing it here. There's a larger discussion around architecture for this operator happening at the moment. I won't link to a RH internal Jira here, but I think our users should be able to click on the Operator in "Installed Operators" and create their `OpenStackDataPlane` CR to define their compute infrastructure.
> 
> If this switch hides the operator from the console, does that mean they wouldn't be able to do this in that case? If so, I would be inclined to skip this change on the dataplane operator. Which would give the users a `OpenStackControlPlane` and a `OpenStackDataPlane` operator in their consoles to define the environment.
> 
> Please let me know if I'm misunderstanding this switch.

I believe this just hides dataplane-operator from the list of operators that can be installed in the console. The OpenStackDataPlane CRD still shows up in the UI when creating custom resources for openstack-operator. Its still in visibleCRDList in csv-merger.go

@dprince is that correct?

fao89:
recheck

dprince:
@slagle has it correct. This is just "hidding" the Dataplane operator in the UI.

The Dataplane CRD here is being listed as a top level resource and should always remain visible at the top (even within the UI). Unless there has been a name change or something is misaligned I think we are still okay here: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/cmd/csv-merger/csv-merger.go#L120

bshephar:
> @slagle has it correct. This is just "hidding" the Dataplane operator in the UI.
> 
> The Dataplane CRD here is being listed as a top level resource and should always remain visible at the top (even within the UI). Unless there has been a name change or something is misaligned I think we are still okay here: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/cmd/csv-merger/csv-merger.go#L120

Right, so looks like it just hides it from the OperatorHub page specifically, not from the Installed Operators page, which is why my question was related to:
https://github.com/openshift/console/blob/682e0090ef91c73f0641e08d9280b25b48a43652/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-page.tsx#L105-L109

Understand the requirement to hide it from OperatorHub, I'll clear my vote.

bshephar (approved):

dprince:
> @slagle has it correct. This is just "hidding" the Dataplane operator in the UI.
> 
> The Dataplane CRD here is being listed as a top level resource and should always remain visible at the top (even within the UI). Unless there has been a name change or something is misaligned I think we are still okay here: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/cmd/csv-merger/csv-merger.go#L120

I need to clarify something that was incorrect about the 2nd half of my reply above. Because we no longer merge the dataplane CSV into openstack-operator (it is just pulled in as a dependency) we actually do need to handle this as a special case.

So... we could do as Brent suggested and leave dataplane-operator "visible" in the catalog. I'm not a huge fan of that as I think it is a bit confusing (you'll see 2 OpenStack operators, one depending on the other).

The other options would be: we could just merge in the dataplane CSV into openstack-operator. This would be much more lightweight than what we had before.

Lastly, we could move the CRDs, api/controllers for whatever dataplane stuff into openstack-operator.

Wanted to clarify things which I think have been broken since the last bundle split up.

PR #283, opened 2023-07-11
Fix dataplane status when roles aren't found

### LINKED ISSUE #250: Dataplane status is wrongly set to ready when initialized
I've observed the following behavior when watching the dataplane resource
```
[dpadev@osp-dev-05 devsetup]$ oc get osdp -w
NAME        STATUS   MESSAGE
openstack            
openstack   Unknown   Init
openstack   True      DataPlane ready
openstack   True      DataPlane ready
openstack   False     Deployment in progress
openstack   True      DataPlane ready
```

fao89:
Currently, dataplane is set to ready when the roles aren't in place yet 

closes #250

### DIFF
--- a/controllers/openstackdataplane_controller.go
+++ b/controllers/openstackdataplane_controller.go
@@ -157,6 +157,10 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 		if err != nil {
 			return ctrl.Result{}, err
 		}
+		r.Log.Info("found roles", "total", len(roles.Items))
+		if len(instance.Spec.Roles) > len(roles.Items) {
+			shouldRequeue = true
+		}
 
 		for _, role := range roles.Items {
 			logger.Info("DataPlane deploy", "role.Name", role.Name)
--- a/tests/kuttl/tests/dataplane-service-custom-image/00-assert.yaml
+++ b/tests/kuttl/tests/dataplane-service-custom-image/00-assert.yaml
@@ -29,7 +29,6 @@ status:
     reason: Ready
     status: "True"
     type: SetupReady
-  deployed: true
 ---
 apiVersion: dataplane.openstack.org/v1beta1
 kind: OpenStackDataPlaneRole
--- a/controllers/openstackdataplanerole_controller.go
+++ b/controllers/openstackdataplanerole_controller.go
@@ -189,6 +189,7 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	r.Log.Info("found nodes", "total", len(nodes.Items))
 
 	// Order the nodes based on Name
 	sort.SliceStable(nodes.Items, func(i, j int) bool {
--- a/tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml
+++ b/tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml
@@ -9,9 +9,9 @@ spec:
     edpm-compute-no-nodes:
       nodeTemplate:
         ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
-        services:
-          - configure-network
-          - validate-network
-          - install-os
-          - configure-os
-          - run-os
+      services:
+        - configure-network
+        - validate-network
+        - install-os
+        - configure-os
+        - run-os


fao89:
recheck

tests/kuttl/tests/dataplane-create-test/00-assert.yaml
```diff
@@ -9,12 +9,12 @@ spec:
 status:
   conditions:
   - message: Deployment not started
-    reason: NotRequested
+    reason: Requested
```
slagle, line 12: This doesn't look right. A deployment hasn't been requested in this test, so I don't think this should change.

tests/kuttl/tests/dataplane-create-test/00-assert.yaml
```diff
@@ -9,12 +9,12 @@ spec:
...
     severity: Info
     status: "False"
     type: Ready
   - message: Deployment not started
-    reason: NotRequested
+    reason: Requested
```
slagle, line 17: ditto

controllers/openstackdataplane_controller.go
```diff
@@ -205,6 +213,12 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 			err.Error()))
 		return ctrl.Result{}, err
 	}
+	if !foundRoles {
+		r.Log.Info("couldn't find roles with the label", "openstackdataplane", instance.Name)
+		instance.Status.Conditions.Set(condition.FalseCondition(condition.DeploymentReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.DeploymentReadyInitMessage))
```
slagle, line 218: This would be the reason you had to change the dataplane-create-test. You set the DeploymentReadyCondition to Requested, even though a deployment has not  been requested.

slagle: I don't think you need to set/change any status condition here at all. Why not just move this entire "if" to before the "if" for the deploy. And then not mark SetupReady as True until all the roles are ready.

fao89: done!

tests/kuttl/tests/dataplane-service-custom-image/00-assert.yaml
```diff
@@ -29,7 +29,6 @@ status:
     reason: Ready
     status: "True"
     type: SetupReady
-  deployed: true
```
rabi, line 32: This does not seem right.  Maybe we just need re-queue when all the roles are not created. Below piece of code before [1] after we process the existing items in roles.

```
if len(instance.Spec.Roles) > len(roles.Items) {
	shouldRequeue = true
}
```

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplane_controller.go#L208

fao89: I don't understand, are you asking to move or to remove the code?

rabi: If there are roles/nodes in the spec or not, I think expected final status of dataplane should be deployed: true. So you don't have to remove it from that assert. 

Also, I'm also suggesting to simplify the current code changes and move the check to L208.

fao89: but if the reconcile loop isn't finding the roles, considering deployed = true isn't misleading?

rabi: Roles and Nodes are both optional in the spec[1]. If there are roles in the spec like this kuttl test[2] and it can't find them, should not it requeue and wait for it.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplane_types.go#L32-L35
[2] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/tests/kuttl/tests/dataplane-service-custom-image/00-dataplane-create.yaml#L24-L26



[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplane_types.go#L32-L35

fao89:
recheck

tests/kuttl/tests/dataplane-service-custom-image/00-assert.yaml
fao89: I've made some changes to the code, so deployed will stay true in this case

fao89:
recheck

fao89:
recheck

controllers/openstackdataplane_controller.go
```diff
@@ -205,12 +205,21 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		r.Log.Info("found roles", "total", len(roles.Items))
+		shouldRequeue = true
+	}
 	if shouldRequeue {
 		r.Log.Info("one or more roles aren't ready, requeueing")
 		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
 	}
 	if instance.Spec.DeployStrategy.Deploy && len(deployErrors) == 0 {
 		instance.Status.Deployed = true
+		if len(roles.Items) == 0 {
```
rabi, line 218: Isn't dataplane created without any roles a valid scenario? If that's  not true, then we should make `roles` required in dataplane CRD. Here we won't set DeploymentReadyCondition for that scenario.

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

rebtoor (commented):
+2

fao89:
@rabi @slagle could you please take another look?

controllers/openstackdataplane_controller.go
```diff
@@ -205,12 +205,21 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if len(roles.Items) > 0 && len(instance.Spec.Roles) > len(roles.Items) {
+		shouldRequeue = true
+	}
 	if shouldRequeue {
 		r.Log.Info("one or more roles aren't ready, requeueing")
 		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
 	}
 	if instance.Spec.DeployStrategy.Deploy && len(deployErrors) == 0 {
 		instance.Status.Deployed = true
+		if len(instance.Spec.Roles) > 0 && len(roles.Items) == 0 {
```
rabi, line 218: I think we don't need this here and why would you set `instance.Status.Deployed = true` a line before when `len(instance.Spec.Roles) >0`.  Would not removing condition `len(roles.Items) > 0` from L209 would requeue for the roles to be created if `len(roles.Items) == 0`.

fao89: last time I tried to requeue when `len(roles.Items) == 0` it didn't reconcile `OpenstackDataplaneRole`, it kept requeuing `OpenstackDataplane`.
Let me test again to confirm

fao89: Actually, the infinite requeue was not cause by this. I was running the debugger with webhooks disabled and the node wasn't getting the SSH secret 🤦🏽‍♂️

rabi (commented):
/lgtm

kuttl-test.yaml
```diff
@@ -19,7 +19,7 @@ kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-dataplane
 namespace: openstack
-timeout: 600
+timeout: 900
```
fao89, line 22: bumping the timeout so we don't need to change the assertions

fao89: 900 wasn't enough to get rid of the errors: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_dataplane-operator/283/pull-ci-openstack-k8s-operators-dataplane-operator-main-dataplane-operator-build-deploy-kuttl/1682020391662391296

fao89:
@rabi @rebtoor I had to push changes for fixing the kuttl tests, please take another look

fao89:
recheck

controllers/openstackdataplane_controller.go
```diff
@@ -137,11 +137,11 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	var deployErrors []string
 	shouldRequeue := false
+	roles := &dataplanev1.OpenStackDataPlaneRoleList{}
```
slagle, line 140: roles will only be populated if deploy=true. I think you need to move the the  r.Client.List for the roles to be outside the deploy "if".

fao89: nice catch!

fao89:
recheck

fao89:
recheck

rebtoor (approved):

PR #286, opened 2023-07-12
Add ability to override OpenStackDataPlane spec

bshephar:
This change adds a new method for overriding `OpenStackDataPlaneSpec` and removed duplicated logic to simplify the envtest implementation. This will increase the flexibility of these methods and allow for a greater range of scenarios to be tested without the need to introduce duplicated blocks of code.


fao89 (approved):
sweet!

PR #288, opened 2023-07-12
Ensure InputReady condition for role/node

rabi:
We don't have a condition to ensure that inputs are ready for the role/node. This adds it and ensures that we check all required keys in the secret.

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -128,4 +128,7 @@ const (
 
 	// RoleDNSDataReadyErrorMessage error
 	RoleDNSDataReadyErrorMessage = "RoleDNSDataReady error occurred"
+
+	// InputReadyWaitingMessage not yet ready
+	InputReadyWaitingMessage = "Waiting for input %s, not yet ready"
 )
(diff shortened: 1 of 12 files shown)


rabi:
recheck

rabi:
recheck

controllers/common.go
```diff
@@ -0,0 +1,90 @@
...
+	AnsibleSSHPrivateKey = "ssh-privatekey"
+	// AnsibleSSHPublicKey ssh public key
+	AnsibleSSHPublicKey = "ssh-publickey"
+	// AnsibleSSHAuthorizedKeys authorized keys
+	AnsibleSSHAuthorizedKeys = "authorized_keys"
+)
+
+// ensureSSHSecret - ensures that the Secret object exists and the expected fields
+// are in the Secret. It returns a hash of the values of the expected fields.
+func ensureSSHSecret(
```
bshephar, line 46: Isn't this essentially just re-implementing the `lib-common` `VerifySecret` function with some condition handling built in?

Were we not able to set these conditions based on the return value of the lib-common function? Just trying to understand the need to reimplement the logic here.

rabi: It's little different, other than checking the secret and fields it updates conditions and does not generate hash which we won't use.

rabi: We can reuse that though. Let me update the PR.

rabi: Hmm.. I remembered the reason I created a different one, to differentiate between status messages (waiting vs error).

bshephar (approved):
lgtm.

PR #290, opened 2023-07-14
Remove DeleteNamespaces function

bshephar:
We are using the test-helpers DeleteNamespace function: https://github.com/openstack-k8s-operators/lib-common/blob/db7193b646f229eed2a06d46a1f6326734a4281c/modules/common/test/helpers/namespace.go#L52-L59

We don't need to duplicate this function in our base_test. This change removes our implementation.


fao89:
recheck

fao89 (approved):

slagle (approved):

fao89:
recheck

PR #293, opened 2023-07-16
Ensure conditions are set based on subconditions

bshephar:
This change ensures that we only set the Ready Condition when all subconditions are true. This change is in-line with the new standard being used within the oko project. See for example: https://github.com/openstack-k8s-operators/heat-operator/commit/f3a97a1678dd585c73b6453e5e3c539084f22ce3


controllers/openstackdataplane_controller.go
```diff
@@ -96,20 +96,20 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
-		// update the overall status condition if service is ready
-		if instance.IsReady() {
```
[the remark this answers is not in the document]
kajinamit: IsReady is now used to check when a CR depends on another CR and we want to ensure the dependent CR is ready before continue reconcile (eg Heat -> KeystoneAPI, OpenStackControlPlane -> Heat). This might not needed in dataplane-operator unless we want to create OpenStackDataplane CRs via meta-operator, but I tend to leave it for consistency.

I'd avoid using `AllSubConditionsIsTrue` to IsReady. The `Ready` condition is meant to present readiness of the CR, and controller is responsible to update it. Instead we should use ReadyCondition, instead of DeploymentReadyCondition, in IsReady, I think

controllers/openstackdataplane_controller.go
```diff
@@ -96,20 +96,20 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
 			// and recalculate it based on the state of the rest of the conditions
-			instance.Status.Conditions.Set(instance.Status.Conditions.Mirror(condition.ReadyCondition))
+			instance.Status.Conditions.Set(
+				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
-
 		err := helper.PatchInstance(ctx, instance)
 		if err != nil {
-			r.Log.Error(_err, "PatchInstance error")
```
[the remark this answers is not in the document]
kajinamit: I suspect that `_err` in the existing code is a typ (it should be `err`). I think we can remove this now assuming this would be logged at an upper layer.

controllers/openstackdataplane_controller.go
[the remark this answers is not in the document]
bshephar: I think that is the conversation I'm recalling too. About the status conditions being set in the controller, and the `IsReady()` used as an interface. I remember there were a few iterations on this in the other operators before arriving at current state.

Thanks for the clarification @kajinamit ++

bshephar:
Baremetal job failing while waiting for nmstate:
```
No resources found in openshift-nmstate namespace.
--
No resources found in openshift-nmstate namespace.
No resources found in openshift-nmstate namespace.
No resources found in openshift-nmstate namespace.
error: timed out waiting for the condition on pods/nmstate-handler-4ds7w
```

Deployment job was RETRY_LIMIT

bshephar:
recheck

controllers/openstackdataplane_controller.go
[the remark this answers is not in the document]
slagle: ok, that works. IsReady and AllSubConditionIsTrue both equate to the CR being ready.

controllers/openstackdataplane_controller.go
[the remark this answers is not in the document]
slagle: Not sure it's a typo. It's the named return value.

controllers/openstackdataplane_controller.go
```diff
@@ -96,20 +96,20 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
-		// update the overall status condition if service is ready
-		if instance.IsReady() {
-			instance.Status.Conditions.MarkTrue(condition.ReadyCondition, dataplanev1.DataPlaneReadyMessage)
+		// update the Ready condition based on the sub conditions
+		if instance.Status.Conditions.AllSubConditionIsTrue() {
+			instance.Status.Conditions.MarkTrue(
+				condition.ReadyCondition, condition.ReadyMessage)
```
fao89, line 102: you are changing the message, so you need to update the kuttl tests

controllers/openstackdataplanenode_controller.go
```diff
@@ -118,24 +118,24 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
+	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
-		// update the overall status condition if service is ready
-		if instance.IsReady() || instanceRole.IsReady() {
+		// update the Ready condition based on the sub conditions
+		if instance.Status.Conditions.AllSubConditionIsTrue() {
 			instance.Status.Deployed = true
-			instance.Status.Conditions.MarkTrue(condition.ReadyCondition, dataplanev1.DataPlaneNodeReadyMessage)
+			instance.Status.Conditions.MarkTrue(
+				condition.ReadyCondition, condition.ReadyMessage)
```
fao89, line 127: ditto

controllers/openstackdataplanerole_controller.go
```diff
@@ -124,23 +124,23 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
-	// persist any changes.
+	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
-		// update the overall status condition if service is ready
-		if instance.IsReady() {
-			instance.Status.Conditions.MarkTrue(condition.ReadyCondition, dataplanev1.DataPlaneRoleReadyMessage)
+		// update the Ready condition based on the sub conditions
+		if instance.Status.Conditions.AllSubConditionIsTrue() {
+			instance.Status.Conditions.MarkTrue(
+				condition.ReadyCondition, condition.ReadyMessage)
```
fao89, line 132: ditto

rebtoor:
recheck

bshephar:
Ok, so we're initialising all of the status conditions for everything even if they aren't going to be used, so when we try to MarkTrue based on all sub conditions, it will never hit that:
```
❯ oc get osdprole -o yaml edpm-compute-no-nodes | yq '.status.conditions[].status'
False
False
True
Unknown
Unknown
Unknown
True
Unknown
Unknown
Unknown
Unknown
Unknown
```

So, I think what I need to do here is go through and only initialise status conditions if they're going to be relevant. Or, set them to Ready if they aren't used. Although, the latter might be confusing.

bshephar:
> Ok, so we're initialising all of the status conditions for everything even if they aren't going to be used, so when we try to MarkTrue based on all sub conditions, it will never hit that:
> 
> ```
> ❯ oc get osdprole -o yaml edpm-compute-no-nodes | yq '.status.conditions[].status'
> False
> False
> True
> Unknown
> Unknown
> Unknown
> True
> Unknown
> Unknown
> Unknown
> Unknown
> Unknown
> ```
> 
> So, I think what I need to do here is go through and only initialise status conditions if they're going to be relevant. Or, set them to Ready if they aren't used. Although, the latter might be confusing.

I see OpenStack operator is still doing it the same way:
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_types.go#L539-L561

Maybe what we need is a "Not Enabled" status message. Then we can add a Not Enabled message and set Ready true in that case. I wonder if that would be less confusing?

bshephar:
> > Ok, so we're initialising all of the status conditions for everything even if they aren't going to be used, so when we try to MarkTrue based on all sub conditions, it will never hit that:
> > ```
> > ❯ oc get osdprole -o yaml edpm-compute-no-nodes | yq '.status.conditions[].status'
> > False
> > False
> > True
> > Unknown
> > Unknown
> > Unknown
> > True
> > Unknown
> > Unknown
> > Unknown
> > Unknown
> > Unknown
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So, I think what I need to do here is go through and only initialise status conditions if they're going to be relevant. Or, set them to Ready if they aren't used. Although, the latter might be confusing.
> 
> I see OpenStack operator is still doing it the same way: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_types.go#L539-L561
> 
> Maybe what we need is a "Not Enabled" status message. Then we can add a Not Enabled message and set Ready true in that case. I wonder if that would be less confusing?

So, something like this:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/293/commits/a18d2aebf9846f0896a3204d18d1f346d594c4a1

controllers/openstackdataplanerole_controller.go
```diff
@@ -264,6 +264,11 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		if err != nil || !isReady {
 			return ctrl.Result{}, err
 		}
+	} else {
+		// We don't require any baremetal nodes, so we will remove these conditions from the status
+		instance.Status.Conditions.Remove(dataplanev1.RoleBareMetalProvisionReadyCondition)
+		instance.Status.Conditions.Remove(dataplanev1.RoleDNSDataReadyCondition)
+		instance.Status.Conditions.Remove(dataplanev1.RoleIPReservationReadyCondition)
```
bshephar, line 271: Removing these is cleaner, and matches how we handle similar situations in the openstack-operator:
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/heat.go#L31

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -107,11 +107,21 @@ func (instance *OpenStackDataPlaneRole) InitConditions() {
...
 	)
 
+	// Only set Baremetal related conditions if we have baremetal hosts included in the
+	// baremetalSetTemplate.
+	if len(instance.Spec.BaremetalSetTemplate.BaremetalHosts) >= 1 {
+		bmConditionsList := []condition.Type{
+			RoleBareMetalProvisionReadyCondition,
+			RoleIPReservationReadyCondition,
+			RoleDNSDataReadyCondition,
+		}
```
bshephar, line 119: This list could probably go into the api package where these constants are defined. That might be a bit cleaner than having them here.

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -107,11 +107,21 @@ func (instance *OpenStackDataPlaneRole) InitConditions() {
...
 		condition.UnknownCondition(condition.InputReadyCondition, condition.InitReason, condition.InitReason),
 		condition.UnknownCondition(SetupReadyCondition, condition.InitReason, condition.InitReason),
-		condition.UnknownCondition(RoleBareMetalProvisionReadyCondition, condition.InitReason, condition.InitReason),
-		condition.UnknownCondition(RoleIPReservationReadyCondition, condition.InitReason, condition.InitReason),
-		condition.UnknownCondition(RoleDNSDataReadyCondition, condition.InitReason, condition.InitReason),
 	)
 
+	// Only set Baremetal related conditions if we have baremetal hosts included in the
+	// baremetalSetTemplate.
+	if len(instance.Spec.BaremetalSetTemplate.BaremetalHosts) >= 1 {
```
kajinamit, line 114: `> 0` would be consistent with the logic in controller

controllers/openstackdataplanerole_controller.go
kajinamit: I personally prefer keeping these to make it more explicit that these steps are skipped but this is jut my own preference.

bshephar: I originally did, but it seemed a little less clean, possibly confusing for users to have things there that they didn't ask for?
https://github.com/openstack-k8s-operators/dataplane-operator/compare/d648e9123b760cc7604f163d2182be8cfa4e31de..d8a00a56bfd98b0610f21310665193bfbae8cf6b

api/v1beta1/openstackdataplanerole_types.go
bshephar: +1 fixed this in the latest commit.

controllers/openstackdataplanenode_controller.go
```diff
@@ -118,24 +118,26 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
...
+				// and recalculate it based on the state of the rest of the conditions
+				instance.Status.Conditions.Set(
+					instance.Status.Conditions.Mirror(condition.ReadyCondition))
+			}
 		}
-
 		err := helper.PatchInstance(ctx, instance)
 		if err != nil {
-			r.Log.Error(_err, "PatchInstance error")
+			r.Log.Error(_err, "Error updating instance status conditions")
```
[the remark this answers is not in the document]
bshephar: I'm not sure why I changed this tbh. I'll revert it

controllers/openstackdataplanerole_controller.go
```diff
@@ -124,23 +124,25 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+				// and recalculate it based on the state of the rest of the conditions
+				instance.Status.Conditions.Set(
+					instance.Status.Conditions.Mirror(condition.ReadyCondition))
+			}
 		}
-
 		err := helper.PatchInstance(ctx, instance)
 		if err != nil {
-			r.Log.Error(_err, "PatchInstance error")
+			r.Log.Error(_err, "Error updating instance status conditions")
```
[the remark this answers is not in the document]
bshephar: ditto. :)

controllers/openstackdataplanenode_controller.go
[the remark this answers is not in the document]
abays: Yeah, logging `err` here seems like the right thing to do.  Then setting `_err = err` on L139 ensures that the error is returned to the `Reconcile()` caller as you would normally do for any other error outside this specific context of a embedded deferred function.

bshephar:
recheck

fao89 (commented):
+2

fao89:
recheck

bshephar:
Looks like `install-os` service was never marked as Ready. But I'm not sure it actually finished:

https://review.rdoproject.org/zuul/build/2476d931d6b44c0689f70983c2544292/log/ci-framework-data/logs/crc/pods/openstack_dataplane-deployment-install-os-edpm-compute-2wkhs_d0261907-7b2e-4dc6-baff-165e6a0b6dad/openstackansibleee/0.log

bshephar:
recheck

bshephar:
Baremetal worked, the other one didn't. What's our stance here?

slagle:
> Baremetal worked, the other one didn't. What's our stance here?

Override! :v:

slagle (approved):

controllers/openstackdataplanerole_controller.go
```diff
@@ -267,7 +267,10 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 			return ctrl.Result{}, err
 		}
 	} else {
+		// We don't require any baremetal nodes, so we will remove these conditions from the status
 		instance.Status.Conditions.Remove(dataplanev1.RoleBareMetalProvisionReadyCondition)
+		instance.Status.Conditions.Remove(dataplanev1.RoleDNSDataReadyCondition)
```
rabi, line 272: This is wrong. IPReservation and Dns is used for both pre-provisoned and baremetal deployments.

api/v1beta1/openstackdataplanerole_types.go
```diff
@@ -107,12 +107,22 @@ func (instance *OpenStackDataPlaneRole) InitConditions() {
...
-		condition.UnknownCondition(RoleDNSDataReadyCondition, condition.InitReason, condition.InitReason),
 	)
 
-	if instance.Spec.Services != nil {
+	// Only set Baremetal related conditions if we have baremetal hosts included in the
+	// baremetalSetTemplate.
+	if len(instance.Spec.BaremetalSetTemplate.BaremetalHosts) > 0 {
+		bmConditionsList := []condition.Type{
+			RoleBareMetalProvisionReadyCondition,
+			RoleIPReservationReadyCondition,
```
rabi, line 117: This is wrong too, we should not remove  these conditions. The earlier code was correct.

PR #294, opened 2023-07-17
Only mark SetupCondition after resources created

bshephar:
We currently always set the SetupCondition to True regardless of anything else. This change moves the declaration for Marktrue to be after the createOrPathDataPlaneResources() function. This way we only set SetupComplete if the dataplane resources were configured without error.

### DIFF
--- a/controllers/openstackdataplane_controller.go
+++ b/controllers/openstackdataplane_controller.go
@@ -125,15 +125,15 @@ func (r *OpenStackDataPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 		instance.Status.Conditions.MarkFalse(dataplanev1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
 	}
 
-	// all setup tasks complete, mark SetupReadyCondition True
-	instance.Status.Conditions.MarkTrue(dataplanev1.SetupReadyCondition, condition.ReadyMessage)
-
 	ctrlResult, err := createOrPatchDataPlaneResources(ctx, instance, helper)
 	if err != nil {
 		return ctrl.Result{}, err
 	} else if (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, nil
 	}
+
+	// all setup tasks complete, mark SetupReadyCondition True
+	instance.Status.Conditions.MarkTrue(dataplanev1.SetupReadyCondition, condition.ReadyMessage)
 
 	var deployErrors []string
 	shouldRequeue := false


abays:
I'm curious as to why there is a `SetupReadyCondition` as well as the `lib-common` `ReadyCondition` here.  The comment description of the former says `"Overall setup condition"`.  Isn't that what the standard `lib-common` `ReadyCondition` is used to indicate?  Or is there some special distinction needed for the DataPlane operator that other operators don't need to consider?  I just ask because if this is more or a less a duplicate, it seems like the sub-condition consideration logic [1] would cover what we're trying to ensure in this PR.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/4f5043fc7637d1c3b0e2dca9fa0ef8e97e97aab8/controllers/openstackdataplane_controller.go#L100-L108

bshephar:
> I'm curious as to why there is a `SetupReadyCondition` as well as the `lib-common` `ReadyCondition` here. The comment description of the former says `"Overall setup condition"`. Isn't that what the standard `lib-common` `ReadyCondition` is used to indicate? Or is there some special distinction needed for the DataPlane operator that other operators don't need to consider? I just ask because if this is more or a less a duplicate, it seems like the sub-condition consideration logic [1] would cover what we're trying to ensure in this PR.
> 
> [1]
> 
> https://github.com/openstack-k8s-operators/dataplane-operator/blob/4f5043fc7637d1c3b0e2dca9fa0ef8e97e97aab8/controllers/openstackdataplane_controller.go#L100-L108

Yeah, we discussed this today in the DF Slack channel. But tbh, I'm not sure why we need it at all, but we should either remove it or move it to somewhere actually doing something. As it stands currently, we set it to False and then regardless of anything else we set it to true before actually applying it to the cluster.

I'm happy to remove it if no one has a reason why it should exist.

bshephar:
> I'm curious as to why there is a `SetupReadyCondition` as well as the `lib-common` `ReadyCondition` here. The comment description of the former says `"Overall setup condition"`. Isn't that what the standard `lib-common` `ReadyCondition` is used to indicate? Or is there some special distinction needed for the DataPlane operator that other operators don't need to consider? I just ask because if this is more or a less a duplicate, it seems like the sub-condition consideration logic [1] would cover what we're trying to ensure in this PR.
> 
> [1]
> 
> https://github.com/openstack-k8s-operators/dataplane-operator/blob/4f5043fc7637d1c3b0e2dca9fa0ef8e97e97aab8/controllers/openstackdataplane_controller.go#L100-L108

Just another note that complete removal would require some more careful consideration and discussion around the intended use case. While the implementation in the `dataplane_controller` isn't doing anything currently, and is made functional by this PR. We have the same Status condition defined in each of the three Controllers and the other two are currently functionally effective.

slagle:
> I'm curious as to why there is a `SetupReadyCondition` as well as the `lib-common` `ReadyCondition` here. The comment description of the former says `"Overall setup condition"`. Isn't that what the standard `lib-common` `ReadyCondition` is used to indicate? Or is there some special distinction needed for the DataPlane operator that other operators don't need to consider? I just ask because if this is more or a less a duplicate, it seems like the sub-condition consideration logic [1] would cover what we're trying to ensure in this PR.

Yes, it's different from other operators due to how the ansible is handled. I think if we weren't trying to straddle across 2 different platforms with dataplane-operator, it would be the same as other operators.

SetupReady is meant to indicate exactly what this PR is fixing, that all k8s resources are Ready (or Setup at least in the case of other dataplane-operator resources).

DeploymentReadyCondition (from lib-common) is used indicate the status of ansible execution, since this can be triggered indepently from creating the Dataplane resources (deploy: true|false).

abays:
> > I'm curious as to why there is a `SetupReadyCondition` as well as the `lib-common` `ReadyCondition` here. The comment description of the former says `"Overall setup condition"`. Isn't that what the standard `lib-common` `ReadyCondition` is used to indicate? Or is there some special distinction needed for the DataPlane operator that other operators don't need to consider? I just ask because if this is more or a less a duplicate, it seems like the sub-condition consideration logic [1] would cover what we're trying to ensure in this PR.
> 
> Yes, it's different from other operators due to how the ansible is handled. I think if we weren't trying to straddle across 2 different platforms with dataplane-operator, it would be the same as other operators.
> 
> SetupReady is meant to indicate exactly what this PR is fixing, that all k8s resources are Ready (or Setup at least in the case of other dataplane-operator resources).
> 
> DeploymentReadyCondition (from lib-common) is used indicate the status of ansible execution, since this can be triggered indepently from creating the Dataplane resources (deploy: true|false).

You mentioned `DeploymentReadyCondition`, which is then a third condition in addition to the two (`SetupReadyCondition` from this repo and `ReadyCondition` from `lib-common`).  Just wanted to make sure I understand what you're saying and that you weren't conflating `DeploymentReadyCondition` with `ReadyCondition`.  But I otherwise get what you are saying about the DataPlane operator extending across two platforms.

rebtoor:
recheck

slagle (approved):

slagle:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

slagle (approved):

PR #297, opened 2023-07-18
EDPM Configuration Generation

slagle:
Commits:

- Add service Secret data to ExtraMounts
- Add service ConfigMap data to ExtraMounts
- Add ConfigMaps and Secrets OpenStackDataPlaneService CRD


config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
```diff
@@ -0,0 +1,18 @@
...
+  secrets:
+    - nova-cell1-compute-config
+  role:
+    name: "Deploy EDPM Nova"
+    hosts: "all"
+    strategy: "linear"
+    tasks:
+      - name: "nova"
+        import_role:
+          name: "osp.edpm.edpm_nova"
```
SeanMooney, line 16: this is  in https://github.com/openstack-k8s-operators/edpm-ansible/pull/217

https://github.com/openstack-k8s-operators/edpm-ansible/pull/217 is on top of 215
but i have updated 215 a few times as i have been working on it ill rebase 217

slagle:
2 of the kuttl tests will fail until https://github.com/openstack-k8s-operators/edpm-ansible/pull/217 is merged.

config/samples/dataplane_v1beta1_openstackdataplanerole.yaml
```diff
@@ -4,58 +4,8 @@ metadata:
...
 spec:
   services:
-    - configure-network
-    - validate-network
     - install-os
-    - configure-os
-    - run-os
   nodeTemplate:
-    networkConfig:
-      template: |
```
fao89, line 14: why removing this?

go.mod
```diff
@@ -5,7 +5,6 @@ go 1.19
 replace github.com/openstack-k8s-operators/dataplane-operator/api => ./api
 
 require (
-	dario.cat/mergo v1.0.0
```
fao89, line 8: don't we need this?
@rebtoor

main.go
```diff
@@ -60,6 +61,7 @@ func init() {
 	utilruntime.Must(novav1beta1.AddToScheme(scheme))
 	utilruntime.Must(baremetalv1.AddToScheme(scheme))
 	utilruntime.Must(infranetworkv1.AddToScheme(scheme))
+	utilruntime.Must(dataplanev1beta1.AddToScheme(scheme))
```
fao89, line 64: it is duplicating line 58

pkg/deployment/deployment.go
```diff
@@ -313,3 +256,111 @@ func ConditionalDeploy(
...
+			name := fmt.Sprintf("%s-%s", secretName, strconv.Itoa(idx))
+			volume := corev1.Volume{
+				Name: name,
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: secretName,
+						Items: []corev1.KeyToPath{
+							{
+								Key:  key,
+								Path: key,
```
fao89, line 344: why setting path with key?

tests/kuttl/tests/dataplane-service-config/00-create.yaml
```diff
@@ -0,0 +1,67 @@
...
+  name: kuttl-service
+spec:
+  label: dataplane-deployment-kuttl-service
+  serviceTask: kuttl-service
+  configMaps:
+    - kuttl-service-cm-0
+    - kuttl-service-cm-1
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneServiceTask
```
fao89, line 42: I couldn't find its API

go.mod
rebtoor: Nope, it's correct its removal since `pkg/deployment/nova.go` has been removed.

config/samples/dataplane_v1beta1_openstackdataplanerole.yaml
slagle: Bad rebase most likely. I'll put it back.

rebtoor (approved):

pkg/deployment/deployment.go
```diff
@@ -313,3 +256,111 @@ func ConditionalDeploy(
...
+	ctx context.Context,
+	helper *helper.Helper,
+	aeeSpec dataplanev1.AnsibleEESpec,
+	service dataplanev1.OpenStackDataPlaneService,
+) (dataplanev1.AnsibleEESpec, error) {
+
+	client := helper.GetClient()
+	baseMountPath := path.Join(ConfigPaths, service.Name)
+
+	for _, cmName := range service.Spec.ConfigMaps {
```
bshephar, line 271: I guess it's possible that both of these parameters are empty, since they are optional. Maybe we can exit early in that case before starting all of this logic?

```
if len(service.Spec.ConfigMaps) == 0 && len(service.Spec.Secrets) == 0 {
    return aeeSpec, nil
}
```

Because this function is handling both cases, we can't check one without the other otherwise we might skip required logic. But maybe an early out if both are empty will save a little time.

It's not a huge deal though, if you're ready to merge it I wouldn't worry about it now. I think a for loop over a nil value would be fairly quick anyway. We're probably only talking a difference of a couple of ms.

slagle:
I've split the commit that migrates nova and libvirt to the composable service interface into a seperate PR: https://github.com/openstack-k8s-operators/dataplane-operator/pull/337

fao89:
recheck

fao89 (approved):

PR #298, opened 2023-07-19
Populate node inventory from IPAM

rabi:
Though in normal scenarios we reconcile the role to deploy the nodes, in the current design there is also a need to deploy separate nodes. This updates the node inventory from IPAM to ensure that it works.


rabi:
recheck

slagle:
recheck

slagle (approved):

PR #299, opened 2023-07-19
configure internal dns in extracompute aee pods

rabi:
Though we would be moving them to edpm-ansible in the future, this would wire dns configuration so that nova ansibleee pods can reach edpm nodes using hostname.


fao89 (approved):

fao89:
recheck

PR #300, opened 2023-07-19
Ensure that all nodes are in SetupReady state before starting deploy

rebtoor:
All nodes should be in SetupReady state before the generation of the `roleConfigMap`, otherwise the first osaee service job will start leaving few nodes out of its inventory.

Here's an example of an inventory extracted from a `setup-repo` (the first osaee service) pod during its execution on 20 nodes:

```
$ ansible -i /tmp/hosts --list-hosts all
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
  hosts (16):
    edpm-compute-0
    edpm-compute-1
    edpm-compute-2
    edpm-compute-3
    edpm-compute-4
    edpm-compute-5
    edpm-compute-6
    edpm-compute-7
    edpm-compute-10
    edpm-compute-11
    edpm-compute-12
    edpm-compute-13
    edpm-compute-14
    edpm-compute-15
    edpm-compute-16
    edpm-compute-18
```

### DIFF
--- a/api/v1beta1/openstackdataplanenode_types.go
+++ b/api/v1beta1/openstackdataplanenode_types.go
@@ -85,9 +85,14 @@ func init() {
 	SchemeBuilder.Register(&OpenStackDataPlaneNode{}, &OpenStackDataPlaneNodeList{})
 }
 
-// IsReady - returns true if the DataPlane is ready
+// IsReady - returns true if the DataPlaneNode is ready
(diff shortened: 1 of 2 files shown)


fao89 (commented):
+2

rebtoor:
recheck

slagle:
recheck

slagle (approved):

PR #304, opened 2023-07-21
Remove usage of NAD ready condition

kajinamit:
NetworkAttachmentsReady condition explains whether the requested NAD resource exists, but is also used to explain whether the attachments are present in pod deployed. Because of this the messages we have for NAD conditions contains "creating" which does not match fully the implementation in Dataplne Node (it just checks existence).

This drops NAD condition used in Dataplane node, and squash the check into input peady condition check. This also resolves the problems with NAD conditions (It is not initialized. Also it is not set to True when but only set to False when something goes wrong).

### DIFF
--- a/controllers/openstackdataplanenode_controller.go
+++ b/controllers/openstackdataplanenode_controller.go
@@ -170,8 +170,8 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 				condition.InputReadyCondition,
 				condition.RequestedReason,
 				condition.SeverityInfo,
-				fmt.Sprintf(dataplanev1.InputReadyWaitingMessage,
-					"secret/"+ansibleSSHPrivateKeySecret))
+				dataplanev1.InputReadyWaitingMessage,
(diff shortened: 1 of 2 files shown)


kajinamit:
recheck

controllers/openstackdataplanenode_controller.go
```diff
@@ -170,8 +170,8 @@ func (r *OpenStackDataPlaneNodeReconciler) Reconcile(ctx context.Context, req ct
 				condition.InputReadyCondition,
 				condition.RequestedReason,
 				condition.SeverityInfo,
-				fmt.Sprintf(dataplanev1.InputReadyWaitingMessage,
-					"secret/"+ansibleSSHPrivateKeySecret))
+				dataplanev1.InputReadyWaitingMessage,
+				"secret/"+ansibleSSHPrivateKeySecret)
```
bshephar, line 174: Yeah, this looks better than `fmt.Sprintf`

kajinamit:
recheck

fao89 (approved):

PR #310, opened 2023-07-24
Remove config/manager/kustomization.yaml from .gitignore

kajinamit:
This file is required to build a bundle image. The file is actually maintained in the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ testbin/*
 bundle/*
 bundle.Dockerfile
 custom-bundle.Dockerfile.pinned
-config/manager/kustomization.yaml
 
 # Test binary, build with `go test -c`
 *.test


fao89 (approved):

.gitignore
```diff
@@ -11,7 +11,6 @@ testbin/*
 bundle/*
 bundle.Dockerfile
 custom-bundle.Dockerfile.pinned
-config/manager/kustomization.yaml
```
rabi, line 14: I thought there is a reason this file is in .gitignore. The image[1] in kustomization.yaml gets updated when you run docker-build/docker-push and push.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/manager/kustomization.yaml#L15

rabi (changes requested):

.gitignore
abays: It does get updated when you do that.  My understanding was that we have a base `config/manager/kustomization.yaml` and that already exists in the repo.  So the `.gitignore` for it was okay because we already have what we need in the repo.  But I suppose the format might change and at some point we might need a new base?

kajinamit: Ah, ok. I overlooked the fact that the file is modified by `make bundle`. Thanks for pointing that out.

PR #314, opened 2023-07-26
Allow the creation of Kubernetes services if openstackdataplaneservic…

jlarriba:
…e defines them


controllers/openstackdataplane_controller.go
```diff
@@ -54,6 +54,7 @@ type OpenStackDataPlaneReconciler struct {
 //+kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplanes/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplanes/finalizers,verbs=update
 //+kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplaneservices,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete;
```
slagle, line 57: you'd need this in dataplanerole_controller as well.

slagle (commented):
I didn't test it, but it lgtm other than the one comment.

pkg/deployment/kube_service.go
```diff
@@ -0,0 +1,134 @@
...
+		err := controllerutil.SetControllerReference(instance, service, helper.GetScheme())
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	return service, err
+}
```
bshephar, line 90: Any reason why we wouldn't use the helper functions in lib-common to create these services?
https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service.go

jlarriba: Because there are no helper functions to create EndpointSlices. As we need create that using the generic CreateOrUpdate function, I think that using also the same structure for the Service gives better readability and consistency in this code.

controllers/openstackdataplane_controller.go
jlarriba: I can do, but I have been testing this and with just this works perfectly. However, I am adding this to dataplanerole_controller just in case there is some case that I have not considered in my tests.

abays: Technically one annotation in one controller will then allow for all the other controllers to have the same permission, but for consistency's sake it's good to be explicit and annotate each controller to indicate exactly what permissions you expect them to have.

jlarriba:
recheck

jlarriba:
recheck

jlarriba:
I consider myself defeated by Zuul's stubborness :(

abays:
> I consider myself defeated by Zuul's stubborness :(

We can ovveride the check if we can just get the `dataplane-operator-crc-podified-edpm-baremetal ` Zuul job to pass, since the other jobs have passed independently across the attempted runs.

slagle (approved):
This lgtm, and I'll approve it. A kuttl test would however be nice, and it looks like it would work and we could test for the actual Service and EndpointSlice getting created.

I also think we may want to look at renaming Role.Services to avoid the field name overlap. Perhaps Role.DataPlaneServices or Role.AnsibleServices.

jlarriba:
> This lgtm, and I'll approve it. A kuttl test would however be nice, and it looks like it would work and we could test for the actual Service and EndpointSlice getting created.
> 
> I also think we may want to look at renaming Role.Services to avoid the field name overlap. Perhaps Role.DataPlaneServices or Role.AnsibleServices.

I think that this suggestions are very applicable, and I will work on them on a future PR, once this one is merged.

It has lgtm, but still lacks approval.

PR #315, opened 2023-07-26
Add $SKIP_FIREWALL

slagle:
firewall-cmd is distro dependent, so it can be skipped in some cases
with SKIP_FIREWALL=true.

SKIP_FIREWALL=true make run-with-webhooks

This matches functionality from other operators (nova-operator).


bshephar (approved):
We could have probably generalised on something like `iptables` or `nftables`. But this is helpful regardless.

PR #320, opened 2023-08-01
Use consistent logger within functions

bshephar:
We currently mix usage of r.Log and logger within the reconcile function. This change replaces the inconsistent usage within reconcile by leveraging the defined contextual logger.


slagle (commented):
+2

slagle:
recheck

fao89 (approved):

PR #321, opened 2023-08-01
Option to select required DDP package

Jaganathancse:


fao89:
recheck

fao89:
recheck

slagle (changes requested):
kuttl test needs to be fixed since a task was added.

     case.go:366: resource OpenStackAnsibleEE:openstack/dataplane-deployment-configure-network-edpm-compute-no-nodes: .spec.roles.tasks: slice length mismatch: 6 != 7

Issue #327, opened 2023-08-03
dataplane operator loops on role reconcile due to self update of deploy stragey field


SeanMooney:
if a contoler updates its own spec filed on an instance it will trigger a reconcile

the dataplen role controller sets the deploy stragey filed to false explicitly after deployment is compelted

https://github.com/openstack-k8s-operators/dataplane-operator/blob/0efa4a48fc8ab796158ff12d549b9ae94108f909/controllers/openstackdataplanerole_controller.go#L333-L338

to prevent this we either need to stop updating the CR we are reconsiling outside the status filed
or we need to check instance.Status.Deployed 

https://github.com/openstack-k8s-operators/dataplane-operator/blob/0efa4a48fc8ab796158ff12d549b9ae94108f909/controllers/openstackdataplanerole_controller.go#L329C6-L329C6

and early out of the reconsile function here
before the defer function.

PR #328, opened 2023-08-03
early out if role already deployed

SeanMooney:
This change add a check to early out from
the reconcile loop if the openstack dataplane role
status is deployed and it is not marked for deletion.

role reconciliation once completed should only happen again
if the end user sets deployed to false.
This prevents looping of reconciling the NovaExternalCompute
CRs by preventing the role reconcilation form triggering
a reconciliation of all the NovaExternalComputes.

Closes: #327


controllers/openstackdataplanerole_controller.go
```diff
@@ -125,6 +125,13 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
+	if instance.Status.Deployed && instance.DeletionTimestamp.IsZero() {
+		// The role is already deployed and not being deleted, so reconcileation
+		// is already complete.
+		logger.Info("Role already deployed", "instance", instance)
+		return ctrl.Result{}, nil
+	}
+
```
slagle, line 134: This look ok. I'll test it and see if it addresses the issue. We would want this check much later though, as we want the other reconcile logic to still happen (inventory, ensureServices, etc).

controllers/openstackdataplanerole_controller.go
```diff
@@ -125,6 +125,13 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, err
 	}
 
+	if instance.Status.Deployed && instance.DeletionTimestamp.IsZero() {
+		// The role is already deployed and not being deleted, so reconcileation
```
slagle, line 129: reconciliation

controllers/openstackdataplanerole_controller.go
slagle: We should also likely set instance.Status.Deployed=false early within the "if instance.Spec.DeployStrategy.Deploy {" loop (~line 297)

SeanMooney: i belive that is the default but sure i can set this

slagle: This still needs to be moved to probably just before the "if Deploy" loop.

slagle (commented):
I still see looping behavior. Although it does not loop infinitely. Seems to stop after 2 loops.

```
Thu Aug 03 12:25:37 (jslagle@teletran-1) [~/code/osp-ng/openstack-k8s-operators/dataplane-operator] (gh/327)$ oc get openstackansibleee -w
NAME                                                 NETWORKATTACHMENTS                              STATUS   MESSAGE
dataplane-deployment-validate-network-edpm-compute   ["ctlplane","internalapi","storage","tenant"]            
dataplane-deployment-validate-network-edpm-compute   ["ctlplane","internalapi","storage","tenant"]   Unknown   AnsibleExecutionJob not started
dataplane-deployment-validate-network-edpm-compute   ["ctlplane","internalapi","storage","tenant"]   False     AnsibleExecutionJob is running
dataplane-deployment-validate-network-edpm-compute   ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]             
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   Unknown   AnsibleExecutionJob not started
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]             
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   False     AnsibleExecutionJob is running
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   Unknown   AnsibleExecutionJob not started
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   False     AnsibleExecutionJob is running
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]             
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   Unknown   AnsibleExecutionJob not started
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]             
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   Unknown   AnsibleExecutionJob not started
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   False     AnsibleExecutionJob is running
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   False     AnsibleExecutionJob is running
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-libvirt                   ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
nova-edpm-compute-0-deploy-nova                      ["ctlplane","internalapi","storage","tenant"]   True      AnsibleExecutionJob complete
```

SeanMooney:
this is also in the commit messge but the process to re run a dataplane service after everging has deployed once is as follows

oc delete -n openstack openstackAnsibleEE dataplane-deployment-validate-network-edpm-compute
oc patch openstackdataplanerole edpm-compute --subresource status --type merge  -p '{"status":{"conditions":[{"type":"validate-network service ready", "status": "Unknown", "lastTransitionTime": "2023-08-03T17:32:39Z" }]}}'
oc patch openstackdataplanerole edpm-compute --subresource status --type json -p='[{"op": "replace", "path": "/status/deployed", "value": false}]'

note the lastTransitionTime is required in the condition filed but its value can be anything

so the process is delete the AEE for the dataplane service to be executed.
reset the condition to Unkown for that dataplane service
then mark the role as not deployed.

you can also do the last two steps interacitvly with

oc edit openstackdataplanerole edpm-compute --subresource status

or you could combine them into one patch.

--subresource status is important to make sure the update is propagated to db as status changes are discarded by default.

controllers/openstackdataplanerole_controller.go
```diff
@@ -281,6 +281,21 @@ func (r *OpenStackDataPlaneRoleReconciler) Reconcile(ctx context.Context, req ct
...
+	// This can be done by deleting the dataplane service AEE CRs and then
+	// patching the role to set  dataplane service condition ready to "Unknown"
+	// then patching the Deployed flag to false.
+	if instance.Status.Deployed && instance.DeletionTimestamp.IsZero() {
+		// The role is already deployed and not being deleted, so reconciliation
+		// is already complete.
+		logger.Info("Role already deployed", "instance", instance)
+		return ctrl.Result{}, nil
+	}
+
```
slagle, line 298: why not put this after GenerateInventory is called?

bshephar: We should probably introduce the hash change check either while we do this, or before we do this. It's not a great user experience to need to delete a completely different CR to trigger re-runs of Ansible for your dataplane nodes. But in terms of the general objective proposed here, I think it's beneficial to execute the least amount of code possible so it seems like a good idea.

Moving to after `GenerateInventory` as Slagle suggested would probably give us more of a chance to hash and compare changes and trigger that redeployment.

SeanMooney: i didnt put it after the inventory generation as i wanted to prevent the inventory and the existing AEE results getting out of sync.

the second we update the inventory we need to return all the ansible because the previous result are partially invalid.

i.e. i updated ansibel vars and as a result the previous execution no longer match what would happen if we ran with the new inventory.

james suggested that doing the auto retriggering was too complex to backprot into the dev preview at this time and to do it separately  which is why i added this todo to make sure we do that as a follow up.


needing to delete a cr (AEE) to retrigger the execution of the datapalne service is not a side effect of this change. its an existing feature gap.
currently the way the dataplane operator is written if a dataplane service is reconsiled successfully it will never run again. i am just documenting what was required to re run it without deleting and recreating the OpenstackDataplanerole.
deleting and recreation the OpenstackDataplanerole is not nessisarly safe if it cause reconciliation of the bearmetal reosucs.

slagle: Thanks for the explanation about leaving it before GenerateInventory. I agree with that reasoning.

I think the hashing is too big of a change to get into dev preview at this point. We agreed to hold on that when we punted on the DeployIdentifier work. A large chunk of it is already implemented, but we held it due to other changes.

slagle: That being said, I think I'm good to go with this PR. We will need some more testing before we patch it on top of v0.1.0. My testing so far was with it just applied to main.

slagle (commented):
+2

fao89 (approved):

PR #333, opened 2023-08-07
Use consistent timer format

bshephar:
We have a mix of using time.Second * n and time.Duration(n) * time.Second. We should use a consistent method of doing this to make things easier to maintain. This change moves the instances of time.Duration over to time.Second * n


fao89:
recheck

fao89 (approved):

PR #334, opened 2023-08-07
Don't remove IPSet/DNS Ready condtions for pre-provisioned

rabi:
We provision IPs and create DNSData when netconfig/dnsmasq services exist, using the networks provided with the node/role spec. Let's not remove them for all pre-provisioned scenarios.


fao89:
recheck

fao89 (approved):

slagle (approved):

PR #335, opened 2023-08-07
Removed roles from `install-os` service since they already called in `configure-network` service

rebtoor:
The `edpm_bootstrap` and the `edpm_kernel` roles are already defined and called in the `configure-network` service [1], so their presence in `install-os` it's useless.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/130162a8b3727d9dbc7b34ea441c2ea7364db3e6/config/services/dataplane_v1beta1_openstackdataplaneservice_configure_network.yaml#L24-L34


rebtoor:
recheck

fao89:
recheck

PR #338, opened 2023-08-08
Don't attach all networks to ansibleee pods

rabi:
We use mcvlan with NADs, so you can't reach the metallb lb endpoints with these networks. API endpoints inside the cluster should be reached using cluster network.


fao89 (approved):

fao89:
recheck

rabi:
recheck

rabi:
error: unable to default to a user name: the server is currently unable to handle the request (get users.user.openshift.io ~)
gmake: *** [Makefile:448: namespace] Error 1

rabi:
recheck

PR #339, opened 2023-08-08
Remove --leader-elect for controller manager

rabi:
We use a single replica for the controller-manager and there is probably no need to use multiple of them for high availability of the operator pods. This patch removes the leader election with single replica. When we move to use multiple replicas we can use ENV var to patch both.

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -36,4 +36,3 @@ spec:
         args:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
-        - "--leader-elect"
--- a/config/manager/controller_manager_config.yaml
+++ b/config/manager/controller_manager_config.yaml
@@ -7,7 +7,7 @@ metrics:
 webhook:
   port: 9443
 leaderElection:
-  leaderElect: true
+  leaderElect: false
   resourceName: e12e763d.openstack.org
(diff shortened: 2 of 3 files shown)


rabi:
recheck

rabi:
recheck

slagle (commented):
This is a divergence from all the other operators isn't it? Or are you planning on proposing the same change to all of openstack-k8s-operators?

abays:
> > This is a divergence from all the other operators isn't it? Or are you planning on proposing the same change to all of openstack-k8s-operators?
> 
> Yes, the plan is to propose to other operators. Already proposed to few others like [openstack-k8s-operators/openstack-baremetal-operator#63](https://github.com/openstack-k8s-operators/openstack-baremetal-operator/pull/63)

Since we're only ever using one controller-manager replica, the idea is that this leader election is unnecessary and in fact can be obstructive when it fails occasionally in CI.

rabi:
>  can be obstructive when it fails occasionally in CI

My understanding of the issue is as below.

As CI is resource constrained, the default[1] `LeaseDuration` and `RenewDeadline` are not enough to get the lock from etcd. We could have increased those, but all those operations to get a lock to assume leadership is redundant when using only one replica. It's clearly documented[2] that we should enable it when running replicated  components.


[1] https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/manager.go#L209-L220
[2] https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/

slagle (approved):

PR #325, opened 2023-08-02
Get inventory data from cluster

fao89:
https://github.com/openstack-k8s-operators/install_yamls/pull/453

PR #342, opened 2023-08-08
Don't error if resources aren't found when auto setting vars

slagle:
If the needed resources aren't found when attempting to automatically
set needed ansible vars, don't fail. Instead, just don't set the var.
This allows for still creating the dataplane inventory even when the
control plane may not be fully deployed or deployed at all.

### DIFF
--- a/pkg/deployment/inventory.go
+++ b/pkg/deployment/inventory.go
@@ -75,7 +75,7 @@ func GenerateRoleInventory(ctx context.Context, helper *helper.Helper,
 		if _, ok := roleNameGroup.Vars["edpm_ovn_metadata_agent_metadata_agent_DEFAULT_nova_metadata_host"]; !ok {
 			novaSvc := &corev1.Service{}
 			err = helper.GetClient().Get(ctx, types.NamespacedName{Name: "nova-metadata-internal", Namespace: instance.Namespace}, novaSvc)
-			if err != nil {
+			if err != nil && !k8s_errors.IsNotFound(err) {
 				return "", err
 			}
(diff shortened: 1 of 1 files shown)


fao89 (dismissed):

bshephar (commented):
I wouldn't say I'm opposed to it, but just wondering what the use case might be? I mean, if rabbitmq or OVN aren't available, we wont be able to use the dataplane anyway? I thought in that case it might have been better to fail earlier rather than after configuring compute service that don't work?

fao89:
I agree with @bshephar 
maybe we just need it for nova, as not all nodes will be compute ones, but the others will be nice to early fail

slagle:
> I wouldn't say I'm opposed to it, but just wondering what the use case might be? I mean, if rabbitmq or OVN aren't available, we wont be able to use the dataplane anyway? I thought in that case it might have been better to fail earlier rather than after configuring compute service that don't work?

I don't think all nodes are OVN nodes or need rabbitmq either. 3rd party services may remove the need for those.

Requiring these to exist also puts a very strong dependency on the ctlpane, where currently one does not exist.

slagle:
We may actually just need to revert https://github.com/openstack-k8s-operators/dataplane-operator/pull/325

Everything edpm_ovn* should be set by the config generation interface when OVN is migrated to a proper composable service. Those vars/values should be coming through in a ConfigMap/Secret on the DataPlaneService that deploys OVN.

fao89:
+1, I'll revert it

fao89:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/344
https://github.com/openstack-k8s-operators/install_yamls/pull/472

slagle:
Abandoning this one.

PR #348, opened 2023-08-10
Use IPAM consistently

rabi:
If we use both IPs in ansibleVars and IPAM there can be issues with IPAM reserved IPs different from what's used in ansibleVars with duplicates. We've IPAM so that we can avoid duplicates.

There should be single interface to change those. For pre-provisoned nodes we shoud use fixedIP to ensure that we reserve the ctlplane IP we use for the node.

When using IPAM, IPs in ansibleVars would be ignored.


rabi:
recheck

fao89 (approved):

rabi:
recheck

fao89:
recheck

PR #349, opened 2023-08-10
Use appropriate message for final Ready condition

rabi:
We use `SetupReadyCondition` that we set before deployment has been started and use the same message i.e "Setup Complete" for final `Ready` condition. This is confusing. Let's use the custom messages that we've for `Ready` condition.

Also cleans up some of the redundant condition messages.


slagle (commented):
I'm thinking this will need some kuttl test updates.

slagle (approved):

abays:
recheck

abays:
Jobs independently succeeded.

PR #352, opened 2023-08-11
Page not found error

Jaganathancse:


README.md
```diff
@@ -6,12 +6,12 @@ workloads. The dataplane-operator prepares the nodes with enough operating
 system configuration so that they are ready for hosting other required
 OpenStack services and workloads.
 
-See [contributing](contributing.md) for notes for developers and
+See [contributing](docs/contributing.md) for notes for developers and
```
fao89, line 9: this change would break mkdocs
https://logserver.rdoproject.org/52/352/c703497978efd6bcb15b1c115eb5b99171dd52eb/github-check/dataplane-operator-docs-preview/588716e/docs/index.html

fao89 (commented):
I believe we should stop symlinking the `README` and start to have a dedicated file for `docs/index.md`

README.md
slagle: :+1: We don't want to break mkdocs.

slagle (changes requested):

README.md
Jaganathancse: Ack

PR #353, opened 2023-08-11
Force bump for renovate to catch up

fao89:


fao89:
recheck

fao89:
recheck

Makefile
```diff
@@ -348,6 +348,15 @@ gowork: ## Generate go.work file to support our multi module repository
 	go work use ./api
 	go work sync
 
+PHONY: force-bump
+force-bump: ## Force bump after tagging
+	for dep in $$(cat go.mod | grep  openstack-k8s-operators | grep -vE -- 'indirect|dataplane-operator' | awk '{print $$1}'); do \
```
jpodivin, line 353: You can use regular expressions in `awk` to make this a bit more concise.  Same goes for the second script.

jpodivin (changes requested):
If this is supposed to be triggered manually, I would suggest writing some notes about intended usage.

Makefile
fao89: I prefer readability to concision

slagle (approved):

PR #358, opened 2023-08-13
Remove nova leftovers

fao89:


abays:
recheck

docs/diagrams/out/edpm.svg
fao89, line 1: I'm not sure how the relationship with Nova should be represented, I just removed Nova for now

slagle (approved):
Thanks!

PR #359, opened 2023-08-14
Use full link path on README

fao89:


docs/README.md
slagle, line 1: this copies the content from the README in the root of the repo. Is there anyway to reuse that so we don't have to maintain 2 copies?

fao89: my idea didn't work, so I updated the link on README to avoid broken links when clicking from the github repo

slagle (approved):

PR #360, opened 2023-08-14
Remove explicit EDPM deploy pod name variable.

pablintino:
With the introduction of the new wait condition for EDPM deployment there is now no need to check for the status of the deploy pod, so it's name is no more needed.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/442


fao89 (approved):

jpodivin (approved):
Thanks Pablo.

fao89:
recheck

slagle (approved):

PR #361, opened 2023-08-15
Linking playbooks in dataplane os services

jpodivin:
Following services will now use playbooks from edpm-ansible collection, rather than the previous structured 'role' field:

- configure-os
- install-os
- run-os


fao89 (approved):

fao89:
recheck

PR #365, opened 2023-08-15
Linking playbooks in dataplane ceph services

jpodivin:
Following services will now use playbooks from edpm-ansible collection, rather than the previous structured 'role' field:

- ceph-client
- ceph-hci-pre

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/279


jpodivin:
recheck

fao89 (approved):

PR #367, opened 2023-08-16
Remove the yaml mapping for OpenStackAnsibleEERunnerImage

steveb:
Something is happening with the decode or clone of OpenStackDataPlaneServiceSpec which is resulting in OpenStackAnsibleEERunnerImage being stripped, meaning any custom runner image is not being applied for the actual run.

This is an attempt to fix that issue.

### DIFF
--- a/api/v1beta1/openstackdataplaneservice_types.go
+++ b/api/v1beta1/openstackdataplaneservice_types.go
@@ -70,7 +70,7 @@ type OpenStackDataPlaneServiceSpec struct {
 
 	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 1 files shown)


api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -70,7 +70,7 @@ type OpenStackDataPlaneServiceSpec struct {
 
 	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
 	// +kubebuilder:validation:Optional
-	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty" yaml:"openStackAnsibleEERunnerImage,omitempty"`
```
slagle, line 73: This yaml tag is required b/c gopkg.in/yaml.v3 marshals/unmarshals struct fields as all lowercase only. So, we need to tell it to use the k8s camelCase convention instead.

PR #383, opened 2023-09-03
Make with_ipam samples consistent with each other

steveb:
dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml is used by two different make targets in install_yamls, and
config/samples/dataplane_v1beta1_openstackdataplane_with_ipam.yaml by one.

This change aligns these samples for consistency, and to support work in install_yamls to make the baremetal workflow more practical and documented for developers.

The one functional change in this commit is the addition of the
`repo-setup` service to
dataplane_v1beta1_openstackdataplane_with_ipam.yaml. This service is
already patched in[1], and it is really time for
install_yamls/devsetup/scripts/edpm-compute-repos.sh to be retired.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L94-L96

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
@@ -26,6 +26,16 @@ spec:
           value: "True"
         - name: ANSIBLE_VERBOSITY
           value: "2"
+      services:
+        - repo-setup
+        - configure-network
+        - validate-network
+        - install-os
+        - configure-os
+        - run-os
+        - ovn
+        - libvirt
+        - nova
       networkAttachments:
         - ctlplane
       baremetalSetTemplate:
@@ -51,14 +61,13 @@ spec:
           service_net_map:
             nova_api_network: internal_api
             nova_libvirt_network: internal_api
-          edpm_chrony_ntp_servers:
-            - 0.pool.ntp.org
-            - 1.pool.ntp.org
+
           # edpm_network_config
           # Default nic config template for a EDPM compute node
           # These vars are edpm_network_config role vars
           edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
           edpm_network_config_hide_sensitive_logs: false
+          #
           # These vars are for the network config templates themselves and are
           # considered EDPM network defaults.
           neutron_physical_bridge_name: br-ex
@@ -77,8 +86,11 @@ spec:
           edpm_nodes_validation_validate_controllers_icmp: false
           edpm_nodes_validation_validate_gateway_icmp: false
 
-          edpm_ovn_metadata_agent_metadata_agent_default_metadata_proxy_shared_secret: 12345678
-          edpm_ovn_metadata_agent_default_bind_host: 127.0.0.1
+          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_metadata_proxy_shared_secret: 12345678
+          edpm_ovn_metadata_agent_DEFAULT_bind_host: 127.0.0.1
+          edpm_chrony_ntp_servers:
+          - clock.redhat.com
+          - clock2.redhat.com
           dns_search_domains: []
 
           registry_url: quay.io/podified-antelope-centos9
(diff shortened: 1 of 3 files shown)


config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
```diff
@@ -26,10 +26,20 @@ spec:
           value: "True"
         - name: ANSIBLE_VERBOSITY
           value: "2"
+      preProvisioned: false
```
rabi, line 29: This is the default.

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
```diff
@@ -26,10 +26,20 @@ spec:
...
+        - install-os
+        - configure-os
+        - run-os
+        - ovn
+        - libvirt
+        - nova
       networkAttachments:
         - ctlplane
       baremetalSetTemplate:
-        deploymentSSHSecret: dataplane-ansible-ssh-private-key-secret
```
rabi, line 32: This is required field atm in baremetalset spec, though we use the same secret.

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
```diff
@@ -89,12 +102,11 @@ spec:
           edpm_nova_compute_container_image: "{{ registry_url }}/openstack-nova-compute:{{ image_tag }}"
           edpm_nova_libvirt_container_image: "{{ registry_url }}/openstack-nova-libvirt:{{ image_tag }}"
           edpm_ovn_metadata_agent_image: "{{ registry_url }}/openstack-neutron-metadata-agent-ovn:{{ image_tag }}"
-
           gather_facts: false
           enable_debug: false
           # edpm firewall, change the allowed CIDR if needed
           edpm_sshd_configure_firewall: true
-          edpm_sshd_allowed_ranges: ['192.168.111.0/24']
```
rabi, line 97: This had .111 as normally folks would deploy with dev-scripts deployed OCP cluster with CBO. We can change it though.

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
steveb: I know, but this was more about kustomize behaviour if we ever unified to a single gen-edpm-kustomize.sh script. I think it fails with the `replace` op when the value is missing:

https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L70-L72

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
steveb: oh this was a mistake, as was the addition on line 60

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
steveb: since this value is patched by install_yamls scripts anyway, leaving the defaults to match dev-scripts might be best

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
steveb: I might restore this, since that kustomize operation isn't actually required

config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
rabi: Once we move to simplified CRD with NodeSet, we will use webhook to populate this and can remove it then.

rabi (approved):
/lgtm

PR #384, opened 2023-09-04
Remove repo-setup service

steveb:
This is an install-yamls only service and shouldn't be included in these samples.


fao89 (approved):
thank you!

PR #386, opened 2023-09-05
Simplifying the controller result conditional

jpodivin:
Rather than check against an empty object, we can ask the object itself if it's empty.


fao89 (approved):

jpodivin:
recheck

PR #303, opened 2023-07-20
Simplified CRD Proposal

bshephar:
This change is aimed at simplifying the OpenStack Dataplane interface. We're achieving this simplification by:

1. Removing the OpenStackDataPlaneNode controller and CR;
2. We remove logic from the OpenStackDataPlane to automatically create the OpenStackDataPlaneRole;
3. We consolidate inventory generation into the `GenerateRoleInventory` function, where we apply hierarchy of variables to ensure that users can override top level variables on a per node basis.

Remaining work on this PR to get it merged:

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/464
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/479

- [x]  Functional Tests
- [x]  Kuttl Tests 
- [ ] CI pipelines for baremetal and pre-provisioned
   - [ ] Changes to the CI Framework to get the right resource (OpenStackDataPlaneNodeSet instead of OpenStackDataPlaneRole / Node - WIP https://github.com/openstack-k8s-operators/ci-framework/pull/479)
   - [ ] Update install_yamls (partial: https://github.com/openstack-k8s-operators/install_yamls/pull/464)

PR #387, opened 2023-09-05
Removing duplicated logic in reconciliation loop

jpodivin:
The conditional check on DeploymentReadyCondition and nested code were duplicated with only difference being the comment.


slagle:
/hold

In order to avoid a merge conflict, we need to hold this until https://github.com/openstack-k8s-operators/dataplane-operator/pull/303 can be merged. PR 303 is the top priority in order to get merged at this time.

jpodivin:
No longer necessary after #303

PR #389, opened 2023-09-05
Test rewrite

jpodivin:
The operator test coverage is rather low. While we don't have cover every statement we should at very least run some tests on the main control flow branches.

The existing tests were tightened, so they make assertions more aggressively.

### DIFF
--- a/tests/functional/base_test.go
+++ b/tests/functional/base_test.go
@@ -4,6 +4,7 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 
 	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
@@ -102,6 +103,37 @@ func GetDataplaneNodeSet(name types.NamespacedName) *dataplanev1.OpenStackDataPl
 		return nil
 	}, timeout, interval).Should(Succeed())
 	return instance
+}
+
+// Create an empty OpenStackDataPlaneService struct
+// containing only given NamespacedName as metadata
+func DefaultDataplaneService(name types.NamespacedName) map[string]interface{} {
+
+	return map[string]interface{}{
+
+		"apiVersion": "dataplane.openstack.org/v1beta1",
+		"kind":       "OpenStackDataPlaneService",
+		"metadata": map[string]interface{}{
+			"name":      name.Name,
+			"namespace": name.Namespace,
+		}}
+}
+
+// Create an OpenStackDataPlaneService with a given NamespacedName, assert on success
+func CreateDataplaneService(name types.NamespacedName) *unstructured.Unstructured {
+	raw := DefaultDataplaneService(name)
+
+	return th.CreateUnstructured(raw)
+}
+
+// Get service with given NamespacedName, assert on successful retrieval
+func GetService(name types.NamespacedName) *dataplanev1.OpenStackDataPlaneService {
+	foundService := &dataplanev1.OpenStackDataPlaneService{}
+	Eventually(func(g Gomega) error {
+		g.Expect(k8sClient.Get(ctx, name, foundService)).Should(Succeed())
+		return nil
+	}, timeout, interval).Should(Succeed())
+	return foundService
 }
 
 func DeleteNamespace(name string) {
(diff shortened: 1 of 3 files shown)


slagle:
/hold

In order to avoid a merge conflict, we need to hold this until https://github.com/openstack-k8s-operators/dataplane-operator/pull/303 can be merged. PR 303 is the top priority in order to get merged at this time.

slagle:
/unhold

rdoproject.org/github-check jobs should be passing now.

jpodivin:
recheck

tests/functional/service_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var _ = Describe("Dataplane Role Test", func() {
```
gibizer, line 27: We don't have Roles any more

tests/functional/service_test.go
```diff
@@ -0,0 +1,70 @@
...
+			Expect(GetService(dataplaneServiceName).Spec.Services).NotTo(BeEmpty())
+
+			Eventually(func(g Gomega) {
+				service.Spec.Services = []dataplanev1.KubeService{}
+				g.Expect(k8sClient.Update(ctx, service)).To(Succeed())
+			}, timeout, interval).Should(Succeed())
+			Expect(GetService(dataplaneServiceName).Spec.Services).To(BeEmpty())
+		})
+	})
+})
```
gibizer, line 70: As there is no controller for the DataplaneService there is not much logic to test. So I guess this is enough.
If later a controller is added, or webhook validation / defaulting is added then we can add more test cases.

tests/functional/base_test.go
```diff
@@ -63,22 +74,125 @@ func DefaultDataplaneNodeSetTemplate(name types.NamespacedName, spec dataplanev1
 	}
 }
 
+func DefaultNodeSpec() dataplanev1.OpenStackDataPlaneNodeSetSpec {
```
gibizer, line 77: I suggest to switch to unstructured create so that i) you don't have to define unfilled fields ii) you can simulate if the client does not send a field in the json request.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/19e1d7506ae86e1f391f0bfbcac31e228507d999/modules/common/test/helpers/common.go#L87-L111 
[2] https://github.com/openstack-k8s-operators/nova-operator/commit/8aaaaee7d8ef68dd49172b786f117bb5e0aa50fb

gibizer (commented):
I think there is opportunity to test the following scenarios for the NodeSet (can be done in separate PRs):
* webhook logic:
  * spec.NodeTemplate.Nodes[nodeName].HostName gets populated from the nodeName if not set
  * spec.BaremetalSetTemplate.DeploymentSSHSecret gets populated from spec.NodeTemplate.AnsibleSSHPrivateKeySecret if not set
  * spec.BaremetalSetTemplate.BaremetalHosts gets populated
* controller logic:
  * The main happy path where a nodeset gets deployed and therefore the Ready Condition of the NodeSet becomes True. You probably need two main scenario on for the preprovisioned case and one for the non preprovisioned case
    * IpSet and DNSData is created
    * OpenStackBaremetalSet is created (non preprovisioned case)
    * Ansible inventory secret is created
    * for each DataPlaneService an AnsibleEE is created
  * Non happy paths:
    * The Secret missing the private key or the authorized key in case of non preprovisionded
    * Baremetal deployment fails so the service deployment is not started
    * an AnsibleEE execution fails so the deployment fails

For most of these you need to simulate external behavior. Like BaremetalHosts deployement failure / success or AnsibleEE failure / success. You can do that by simply modifying the status of those CRs during the test case.

Let me know if you need help.

jpodivin:
I'm afraid I'll have to rewrite all of this. Recent changes to the operator reconciliation loop have effectively made some of these tests incompatible.

tests/functional/base_test.go
jpodivin: Technically I could leave those fields unfilled. Decorators in type definition are taking care of these. Mostly I've left them like this to guide future test development. If defaults or the way controller manages them it will be more obvious what to adjust. Same goes for tightening tests.

But I understand it's a question of taste.

gibizer: While my first point can be a matter of taste. The second cannot. If you use golang structs as input in your test then you cannot simulate that the user left a field out from the json request. Take the Deploy flag as an example. It is defined as 
```golang
	// +kubebuilder:default=true
	Deploy bool `json:"deploy"`
```
If you want to test that it is properly defaulted to "true" then you need to send in an input json request where the deploy field is unset. You cannot do that with a golang struct.

jpodivin: Ack. I'll add rewrite of existing tests to use unstructured to this PR.

fao89 (approved):

gibizer (approved):

PR #390, opened 2023-09-06
[doc]Trigger nova discover hosts after EDPM deploy

gibizer:
The periodic host discovery is now disabled in nova-scheduler as it is not scaling well. Instead after EDPM compute node deployment the nova-manage cell_v2 discover_hosts CLI needs to be run to trigger the discovery.

This PR documents the new necessary manual step.

Related-To: openstack-k8s-operators/nova-operator#519
Related: https://issues.redhat.com/browse/OSPRH-319


fultonj:
+2

fao89 (approved):

PR #396, opened 2023-09-07
Removing redundant uniqueness check function

jpodivin:
Both the function and the related variable are not used anywhere in the project code base. I assume this is because they were gradually removed, culminating with recent CRD refactor. 

This also allows us to remove import of "reflect".

### DIFF
--- a/api/v1beta1/common.go
+++ b/api/v1beta1/common.go
@@ -18,7 +18,6 @@ package v1beta1
 
 import (
 	"encoding/json"
-	"reflect"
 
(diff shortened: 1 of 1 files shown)


fao89 (approved):

jpodivin:
recheck

rebtoor (approved):

jpodivin:
recheck

fao89:
recheck

slagle (approved):

PR #398, opened 2023-09-07
Fix Regression with IPAM

rabi:
We should use Spec.NodeTemplate.Networks if there are no node.Networks. Also, remove the block to remove ctlplane network. Fixes the sample used in CI.

- ansible_user: cloud-admin
- Remove all ansible vars for network ips
- Remove BMH namespace from the sample as it is also used with CBO. We patch it in install_yamls for CI.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/528

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
@@ -20,7 +20,6 @@ spec:
   deployStrategy:
       deploy: true
   baremetalSetTemplate:
-    bmhNamespace: openstack
     deploymentSSHSecret: dataplane-ansible-ssh-private-key-secret
     bmhLabelSelector:
       app: openstack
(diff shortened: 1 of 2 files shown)


fao89 (approved):

rabi:
recheck

rabi:
recheck

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
```diff
@@ -9,6 +9,7 @@ spec:
     - name: ANSIBLE_ENABLE_TASK_DEBUGGER
       value: "True"
   services:
+    - repo-setup
```
slagle, line 12: repo-setup should not be in the sample here. it's added by install_yamls.

slagle:
@rabi why drop the depends-on https://github.com/openstack-k8s-operators/ci-framework/pull/528. I believe it is still necessary.

slagle:
recheck

i added the depends-on back on https://github.com/openstack-k8s-operators/ci-framework/pull/528. I just want to make sure we don't introduce any new errors, so we don't keep chasing failures.

slagle:
recheck

another depends-on ( https://github.com/openstack-k8s-operators/install_yamls/pull/523) was added to https://github.com/openstack-k8s-operators/ci-framework/pull/528

slagle:
recheck

I pushed a new commit to https://github.com/openstack-k8s-operators/ci-framework/pull/528.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
bshephar: This PR is passing now. So, I think we just need to remove this and then we're good here.

rabi: It was added in https://github.com/openstack-k8s-operators/dataplane-operator/commit/8bdfcb1bf948a126c128fee86d8bd34af81aaaa2a and then removed in a later patch it seems.

rabi:
> @rabi why drop the depends-on [openstack-k8s-operators/ci-framework#528](https://github.com/openstack-k8s-operators/ci-framework/pull/528)? I believe it is still necessary.

It's not necessary for the baremetal job.

rabi:
> recheck
> 
> i added the depends-on back on [openstack-k8s-operators/ci-framework#528](https://github.com/openstack-k8s-operators/ci-framework/pull/528). I just want to make sure we don't introduce any new errors, so we don't keep chasing failures.

Really? You thought this would introduce new errors and break the pre-provisioned multinode job.  We've been merging half baked PRs  which resulted in the current issues.

PR #394, opened 2023-09-07
Remove NetworkConfig override handling

bshephar:
This change removes the CRD level override for network config files. Instead, this allows us to use the Ansible variable directly to set any require network config customizations.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/341
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/560

PR #401, opened 2023-09-08
Updating samples

fao89:
- remove the symlinks
- update IPAM sample

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
@@ -28,7 +28,6 @@ spec:
          hostName: edpm-compute-0
          ansible:
            ansibleHost: 192.168.122.100
-           ansibleUser: root
            ansibleVars:
              ctlplane_ip: 192.168.122.100
              internal_api_ip: 172.17.0.100
@@ -37,7 +36,7 @@ spec:
              fqdn_internal_api: edpm-compute-0.example.com
     managementNetwork: ctlplane
     ansible:
-      ansibleUser: root
+      ansibleUser: cloud-admin
       ansiblePort: 22
       ansibleVars:
          service_net_map:
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
@@ -3,7 +3,6 @@ kind: OpenStackDataPlaneNodeSet
 metadata:
   name: openstack-edpm-ipam
 spec:
-  ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
   env:
     - name: ANSIBLE_FORCE_COLOR
       value: "True"
@@ -22,6 +21,7 @@ spec:
     - libvirt
     - nova
   nodeTemplate:
+    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
     nodes:
        edpm-compute-0:
          hostName: edpm-compute-0
@@ -88,9 +88,6 @@ spec:
          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_nova_metadata_host: 127.0.0.1
          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_metadata_proxy_shared_secret: 12345678
          edpm_ovn_metadata_agent_DEFAULT_bind_host: 127.0.0.1
-         ctlplane_dns_nameservers:
-         - 192.168.122.1
-         dns_search_domains: []
          edpm_ovn_dbs:
          - 192.168.122.1
          registry_url: quay.io/podified-antelope-centos9
--- a/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
@@ -1,1 +1,0 @@
-dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
--- a/config/samples/dataplane_v1beta1_openstackdataplane_with_ipam.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplane_with_ipam.yaml
@@ -1,1 +1,0 @@
-dataplane_v1beta1_openstackdataplanenodeset.yaml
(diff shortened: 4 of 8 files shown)


config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -3,15 +3,11 @@ kind: OpenStackDataPlaneNodeSet
 metadata:
   name: openstack-edpm-ipam
 spec:
-  ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
   env:
     - name: ANSIBLE_FORCE_COLOR
       value: "True"
     - name: ANSIBLE_ENABLE_TASK_DEBUGGER
       value: "True"
-  preProvisioned: true
```
slagle, line 12: What's the reason for dropping this? We should decide on what we want the samples to represent, all fields with default values, or the minimal set of fields. For instance, why drop preProvisioned, but keep env?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -21,60 +17,27 @@ spec:
...
+            subnetName: subnet1
+            defaultRoute: true
+            fixedIP: 192.168.122.100
+          - name: InternalApi
+            subnetName: subnet1
+          - name: Storage
+            subnetName: subnet1
+          - name: Tenant
+            subnetName: subnet1
     managementNetwork: ctlplane
```
slagle, line 38: this is optional. should we drop it?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -21,60 +17,27 @@ spec:
...
     ansible:
-      ansibleUser: cloud-admin
+      ansibleUser: root
```
slagle, line 40: Aren't we sticking with cloud-admin?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -21,60 +17,27 @@ spec:
...
       ansiblePort: 22
```
slagle, line 41: this is the default. should we drop it?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -21,60 +17,27 @@ spec:
...
       ansibleVars:
          service_net_map:
```
slagle, line 43: not used anywhere. can be dropped.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -87,32 +50,11 @@ spec:
          # Default nic config template for a EDPM compute node
          # These vars are edpm_network_config role vars
          edpm_network_config_hide_sensitive_logs: false
+         edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
```
slagle, line 53: this is the default. drop it?

slagle (changes requested):

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
fao89: I know we use cloud-admin for baremetal, but I believe install_yamls still using root

bshephar: We should probably decide what we actually want here. With tripleo, we went through the stages of `heat-admin` and `tripleo-admin`. I guess we want to standardise on something similar here? But probably not root right?

Understand the install_yamls situation. Just saying that we probably want to aim for a non-root user before we GA.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
bshephar: Now that we have merged the new CRD design, I'll keep pushing forward with my other PR to change how this interface works anyway. 
https://github.com/openstack-k8s-operators/dataplane-operator/pull/394

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -130,11 +72,8 @@ spec:
          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_nova_metadata_host: 127.0.0.1
          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_metadata_proxy_shared_secret: 12345678
          edpm_ovn_metadata_agent_DEFAULT_bind_host: 127.0.0.1
-         ctlplane_dns_nameservers:
-         - 192.168.111.1
-         dns_search_domains: []
          edpm_ovn_dbs:
-         - 192.168.111.1
+         - 192.168.122.1
```
bshephar, line 76: I think I had issues with `192.168.122.0/24` which is why I ended up on `192.168.111.0/24` I notice the rdo checks haven't run for this change so we might not be seeing the issue with addressing. Provided this actually works in CI and doesn't require some other change, I have no other preference to any particular subnet. Just calling it out in-case there's more required for this subnet change.

fao89: I think this is another thing that is different when baremetal or not: https://github.com/openstack-k8s-operators/dataplane-operator/commit/9b19caf25ca568012a20738b2b5a12c9579ad4b4

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -28,7 +28,6 @@ spec:
          hostName: edpm-compute-0
          ansible:
            ansibleHost: 192.168.122.100
-           ansibleUser: root
```
slagle, line 31: There's no default set for ansibleUser. Shouldn't it be cloud-admin?

fao89: I'm adding cloud-admin on install_yamls: https://github.com/openstack-k8s-operators/install_yamls/pull/531

slagle: As long as it's set here in the samples as well. thanks.

PR #406, opened 2023-09-11
Fix ExtraMounts

slagle:
ExtraMounts were being dropped and not added to the created
OpenStackAnsibleEE resources. The reason was because as each composable
service is deployed, aeeSpec.ExtraMounts is reset back to the original
value so that there is no cross over between mounts from individual
composable services. The reset was not working because the initialied
value of aeeSpec.ExtraMounts did not have a lenght set, causing the
copy() function to not actually copy any data.

The commit switches to using make(...) with a length parameter to ensure
the copy() works as intended.

Adds a new kuttl test, dataplane-extramounts, to test this specific
functionality.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -87,7 +87,7 @@ func Deploy(
 
 		// Reset ExtraMounts to its original value, and then add in service
 		// specific mounts.
-		aeeSpec.ExtraMounts = []storage.VolMounts{}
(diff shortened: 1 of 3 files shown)


ASBishop (commented):
+1 this fixed the problem I experienced

fultonj:
+2 this also fixed the problem according to my testing

fultonj (approved):

PR #409, opened 2023-09-13
Move nodes to top level of nodeset spec

rabi:
We should not have nodes in nodeTemplate. NodeTemplate contains everything that's common for all nodes.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/535


bshephar (approved):
This sounds good to me.

PR #418, opened 2023-09-19
Add neutron-sriov-agent service to deploy this agent on the EDP nodes

slawqo:
Related: OSP-26189


config/samples/dataplane_v1beta1_openstackdataplanenodeset_sriov.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - name: ANSIBLE_ENABLE_TASK_DEBUGGER
+      value: "True"
+  services:
+    - download-cache
+    - configure-network
+    - validate-network
+    - install-os
+    - configure-os
+    - run-os
+    - neutron-sriov-agent
```
booxter, line 18: I'd suggest we use a more generic name (neutron-sriov instead of neutron-sriov-agent) in case / when we have to do something else but only the agent deployment as part of this Service.

booxter (changes requested):
The name suggestion is just a suggestion.

But you need to add kuttl tests for the service I think. (We have them for other services.)

slawqo:
> The name suggestion is just a suggestion.
> 
> But you need to add kuttl tests for the service I think. (We have them for other services.)

I'm not sure but I don't see such tests for each of the services in tests/kuttl/tests. Can You point me to the tests of other services?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_sriov.yaml
slawqo: Done

booxter:
You can find an example of OVN kuttl check here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/a7fd24b4d0692a401d6fdf73833cf688afc6cf27/tests/kuttl/tests/dataplane-deploy-no-nodes-test/01-assert.yaml#L381-L469

There are more in the file, e.g. libvirt one below.

bshephar:
recheck

bshephar (commented):
Sounds reasonable to me. Let's get CI passing.

Do we have a test plan for this though? Is this something we think would be tested downstream only? I assume there's not a great way to test it outside of e2e tests, since Kuttl and Functional tests would already cover all of the options here. The addition of this service doesn't really add anything complicated to the Kubernetes side of things.

slawqo:
ok, I updated kuttl test to have this one tested too. I hope it will be ok but it's my first time working with kuttl tests so I could do it in wrong way of course :)

booxter:
/lgtm

Code wise, it looks correct. We'll see if CI is happy.

fao89 (approved):

booxter:
@bshephar I think sriov testing will have to rely on downstream CI that has the necessary hardware. There are tempest scenarios that cover sriov.

@vkhitrin anything to add to what Brendan asked about?

vkhitrin:
@bshephar, as far as I am aware, in TripleO based deployments, all SR-IOV initiatives had to be tested downstream since there was no hardware dedicated for it upstream (maybe some organizations donated hardware in the past, but in recent years I don't recall any extra hardware upstream).

We have various tempest suites and functional/performance testing for downstream deployments based on SR-IOV data paths.

booxter:
recheck

viroel:
recheck

viroel:
recheck

telemetry-operator-bundle:232275280f55 is updated now

slawqo:
recheck

rabi:
recheck
 
cifmw-crc-podified-edpm-baremetal  fixed now

bshephar (approved):
Looks good to me. The two seperate `mount` dictionaries seems odd in `01-assert.yaml`, but I see we also have that in other kuttl jobs, so it must be fine. I would have thought they would be grouped together under a single `- mounts:` dictionary.

PR #420, opened 2023-09-20
Change NodeSetDeploymentCondition to condition.Type

bshephar:
We were using the NodeSetDeploymentCondition as a type string. This should be a condition.Type instead. This change fixes that issue and adds some initialisation tests for the dataplaneDeployment.

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -88,7 +88,7 @@ const (
 
 	// NodeSetDeploymentReadyCondition Status=True condition indicates if the
 	// NodeSet Deployment is finished and successful.
-	NodeSetDeploymentReadyCondition string = "%s Deployment ready"
+	NodeSetDeploymentReadyCondition condition.Type = "NodeSetDeploymentReady"
 
 	// NodeSetDeploymentReadyMessage ready
 	NodeSetDeploymentReadyMessage = "%s Deployment ready"
--- a/api/v1beta1/openstackdataplanedeployment_types.go
+++ b/api/v1beta1/openstackdataplanedeployment_types.go
@@ -100,7 +100,7 @@ func (instance *OpenStackDataPlaneDeployment) InitConditions() {
 
 	if instance.Spec.NodeSets != nil {
 		for _, nodeSet := range instance.Spec.NodeSets {
-			readyCondition := condition.Type(fmt.Sprintf(NodeSetDeploymentReadyCondition, nodeSet))
+			readyCondition := condition.Type(fmt.Sprintf(NodeSetDeploymentReadyMessage, nodeSet))
 			cl = append(cl, *condition.UnknownCondition(readyCondition, condition.InitReason, condition.InitReason))
 		}
 	}
--- a/controllers/openstackdataplanedeployment_controller.go
+++ b/controllers/openstackdataplanedeployment_controller.go
@@ -232,7 +232,7 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context, 
 			logger.Info("Set NodeSetDeploymentReadyCondition true", "nodeSet", nodeSet.Name)
 			instance.Status.Conditions.Set(
 				condition.TrueCondition(
-					condition.Type(fmt.Sprintf(dataplanev1.NodeSetDeploymentReadyCondition, nodeSet.Name)),
+					condition.Type(fmt.Sprintf(dataplanev1.NodeSetDeploymentReadyMessage, nodeSet.Name)),
 					condition.DeploymentReadyMessage))
 		}
 	}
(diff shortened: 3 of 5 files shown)


abays:
> @bshephar: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/dataplane-operator-build-deploy-kuttl 	[999ad10](https://github.com/openstack-k8s-operators/dataplane-operator/commit/999ad10935990de854298eb1ac542ddcbc261f1c) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_dataplane-operator/420/pull-ci-openstack-k8s-operators-dataplane-operator-main-dataplane-operator-build-deploy-kuttl/1704346875403964416) 	true 	`/test dataplane-operator-build-deploy-kuttl`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=dataplane-operator&pr=420). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:bshephar).

Given the error here:

```
case.go:366: resource OpenStackAnsibleEE:openstack/dataplane-deployment-nova-edpm-compute-no-nodes: .spec.image: value mismatch, expected: quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest != actual: quay.io/openstack-k8s-operators/openstack-ansibleee-runner@sha256:a1bb61a788f9ec9f82db1a81fa0c7692aa3197af022639bfa94276690203952e
```

It looks like the DataPlane operator KUTTL tests need some sort of "related images" assert like found here [1].

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/0512e2f335dd18713450c065975ac99ee35c12b1/tests/kuttl/common/assert_sample_deployment.yaml#L201-L221

fao89:
recheck

fao89 (approved):

api/v1beta1/conditions.go
```diff
@@ -88,7 +88,7 @@ const (
 
 	// NodeSetDeploymentReadyCondition Status=True condition indicates if the
 	// NodeSet Deployment is finished and successful.
-	NodeSetDeploymentReadyCondition string = "%s Deployment ready"
+	NodeSetDeploymentReadyCondition condition.Type = "NodeSetDeploymentReady"
```
slagle, line 91: I'm confused about the point of this change. NodeSetDeploymentReadyCondition is changed to condition.Type, but then you don't use it anywhere. You switch all the previous uses to use NodeSetDeploymentReadyMessage, which is still a string, as it needs to be so that we can have separate conditions for each NodeSet.

slagle (changes requested):

api/v1beta1/conditions.go
bshephar: Should all of the `*Condition` ones be `condition.Type`:
```
❯ grep Condition api/v1beta1/conditions.go | grep -Ev '^.\/\/'
	ServiceReadyCondition string = "%s service ready"
	SetupReadyCondition condition.Type = "SetupReady"
	RoleBareMetalProvisionReadyCondition condition.Type = "RoleBaremetalProvisionReady"
	NodeSetBareMetalProvisionReadyCondition condition.Type = "NodeSetBaremetalProvisionReady"
	NodeSetIPReservationReadyCondition condition.Type = "NodeSetIPReservationReady"
	NodeSetDNSDataReadyCondition condition.Type = "NodeSetDNSDataReady"
	NodeSetDeploymentReadyCondition condition.Type = "NodeSetDeploymentReady"
	NodeSetServiceDeploymentReadyCondition string = "%s %s Deployment ready"
```

It's more about the inconsistency. I was going to add a test for it, then also realised we weren't using it. But if it's going to exist, it seemed like it should be a `condition.Type`.

bshephar: Ah ServiceReadyCondition is also a string. Seems like that should be a `condition.Type` as well

slagle: condition.Type is just a string anyway. This PR doesn't fix the inconsistency since it just switches to use the Message (also a string) as the type parameter.

My intent is to customize the condition type based on the NodeSet name or Service name, so we can dynamically report the status. Alternatively, we could use a single condition with a static type string, and then append to the message to indicate the state of different NodeSets or services. Although that makes it much harder to interpret the status programatically.

api/v1beta1/openstackdataplanedeployment_types.go
```diff
@@ -100,7 +100,7 @@ func (instance *OpenStackDataPlaneDeployment) InitConditions() {
 
 	if instance.Spec.NodeSets != nil {
 		for _, nodeSet := range instance.Spec.NodeSets {
-			readyCondition := condition.Type(fmt.Sprintf(NodeSetDeploymentReadyCondition, nodeSet))
+			readyCondition := condition.Type(fmt.Sprintf(NodeSetDeploymentReadyMessage, nodeSet))
```
slagle, line 103: this is what I'm referring to ^^. You changed NodeSetDeploymentReadyCondition to be condition.Type, but then you just switch back to using the string here.

slagle (changes requested):

api/v1beta1/openstackdataplanedeployment_types.go
bshephar: Right, I see what you mean. I'll have a re-think about this entire PR. I might close it and start a fresh one with a more wholistic approach specifically around the conditions. Rather than trying to tack it on with my functional tests

PR #425, opened 2023-09-21
Functional tests for successful deployment

jpodivin:


gibizer (commented):
```diff
diff --git a/pkg/util/ansible_execution.go b/pkg/util/ansible_execution.go
index f7fd0cb..bd291bc 100644
--- a/pkg/util/ansible_execution.go
+++ b/pkg/util/ansible_execution.go
@@ -51,7 +51,10 @@ func AnsibleExecution(
 	var err error
 	var cmdLineArguments strings.Builder
 
+	helper.GetLogger().Info("!!! AnsibleExecution", "obj", obj, "label", label)
+
 	ansibleEE, err := GetAnsibleExecution(ctx, helper, obj, label)
+	helper.GetLogger().Info("!!! AnsibleExecution", "ansibleEE", ansibleEE)
 	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
@@ -68,6 +71,8 @@ func AnsibleExecution(
 		}
 	}
 
+	helper.GetLogger().Info("!!! AnsibleExecution create or patch", "ansibleEE", ansibleEE)
+
 	_, err = controllerutil.CreateOrPatch(ctx, helper.GetClient(), ansibleEE, func() error {
 		ansibleEE.Spec.NetworkAttachments = aeeSpec.NetworkAttachments
 		if aeeSpec.DNSConfig != nil {
@@ -168,6 +173,7 @@ func AnsibleExecution(
 		util.LogErrorForObject(helper, err, fmt.Sprintf("Unable to create AnsibleEE %s", ansibleEE.Name), ansibleEE)
 		return err
 	}
+	helper.GetLogger().Info("!!! AnsibleExecution create or patch succeeded", "ansibleEE", ansibleEE)
 
 	return nil
 }
diff --git a/tests/functional/openstackdataplanedeployment_controller_test.go b/tests/functional/openstackdataplanedeployment_controller_test.go
index a7be26d..0c8e3e3 100644
--- a/tests/functional/openstackdataplanedeployment_controller_test.go
+++ b/tests/functional/openstackdataplanedeployment_controller_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	. "github.com/openstack-k8s-operators/lib-common/modules/common/test/helpers"
 	ansibleeev1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
-	ansibleeev1alpha1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
 	baremetalv1 "github.com/openstack-k8s-operators/openstack-baremetal-operator/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -79,27 +78,29 @@ var _ = Describe("Dataplane Deployment Test", func() {
 						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyCondition,
 						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyMessage)
 					g.Expect(th.K8sClient.Status().Update(th.Ctx, &baremetal)).To(Succeed())
-					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &nodeSet)).To(Succeed())
-					g.Expect(th.K8sClient.Status().Update(th.Ctx, &nodeSet)).To(Succeed())
 				}).Should(Succeed())
 
+				service := GetService(dataplaneServiceName)
+
 				Eventually(func(g Gomega) {
-					service := GetService(dataplaneServiceName)
 					ansibleeeName := types.NamespacedName{
 						Name:      fmt.Sprintf("%s-%s", service.Spec.Label, dataplaneDeploymentName.Name),
 						Namespace: dataplaneDeploymentName.Namespace,
 					}
-					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneDeploymentName, &edpmDeployment))
+					logger.Info("getting deployment")
+					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneDeploymentName, &edpmDeployment)).Should(Succeed())
+					logger.Info("got deployment")
+
 					ansibleEE := &ansibleeev1.OpenStackAnsibleEE{
 						ObjectMeta: metav1.ObjectMeta{
 							Name:      ansibleeeName.Name,
 							Namespace: ansibleeeName.Namespace,
-							Labels: map[string]string{
-								service.Spec.Label: string(edpmDeployment.GetUID()),
-							},
 						}}
+					logger.Info("getting", "ansibleEE", ansibleEE)
 					g.Expect(th.K8sClient.Get(th.Ctx, ansibleeeName, ansibleEE)).To(Succeed())
-					ansibleEE.Status.JobStatus = ansibleeev1alpha1.JobStatusSucceeded
+					logger.Info("got", "ansibleEE", ansibleEE)
+
+					ansibleEE.Status.JobStatus = ansibleeev1.JobStatusSucceeded
 
 					g.Expect(th.K8sClient.Status().Update(th.Ctx, ansibleEE)).To(Succeed())
 				}, timeout, interval).Should(Succeed())
diff --git a/tests/functional/suite_test.go b/tests/functional/suite_test.go
index 7211532..26154ff 100644
--- a/tests/functional/suite_test.go
+++ b/tests/functional/suite_test.go
@@ -68,7 +68,7 @@ const (
 	ContainerImage       = "test://nova"
 	timeout              = 40 * time.Second
 	// have maximum 100 retries before the timeout hits
-	interval = 100
+	interval = timeout / 100
 )
 
 func TestAPIs(t *testing.T) {

```

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
 	})
 
 	When("A dataplaneDeployment is created with matching NodeSet", func() {
 		BeforeEach(func() {
-			DeferCleanup(th.DeleteInstance, CreateDataplaneNodeSet(dataplaneNodeSetName, DefaultDataPlaneNoNodeSetSpec()))
-			DeferCleanup(th.DeleteInstance, CreateDataplaneDeployment(dataplaneDeploymentName, DefaultDataPlaneDeploymentSpec()))
 			CreateSSHSecret(dataplaneSSHSecretName)
+			CreateSSHSecret(neutronOvnMetadataSecretName)
+			CreateSSHSecret(novaNeutronMetadataSecretName)
+			CreateSSHSecret(novaCellComputeConfigSecretName)
```
bshephar, line 62: We should rename this function if we want to use it to create generic secrets. Although, do we already have a helper function for this in lib-common? Maybe we can just re-use something there and remove our bespoke implementation of this logic.

bshephar (changes requested):
Nice, it will be really good to have full functional test coverage of the deployment. 

There are probably a couple of things we would want to verify actually happen from the baremetal perspective rather than just marking them as true. Maybe an intermediary step could have been testing the PreProvisioned deployment. But I would definitely prefer to have this coverage merged now and work on refining it over the next quarter.

One small change I would request wrt the secret creation. But otherwise, I'm pretty excited to have this coverage.

tests/functional/openstackdataplanedeployment_controller_test.go
gibizer: These are not SSH secrets so they don't have to have `ssh-privatekey` field, so I think simply reusing th.CreateSecret is enough as the dataplane operator really don't care about the content of these secrets. I even doubt that it care about the existence of these secrets.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, &nodeSet)).To(Succeed())
+				}).Should(Succeed())
+
+				// Set baremetal provisioning conditions to True
+				Eventually(func(g Gomega) {
+					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &baremetal)).To(Succeed())
+					baremetal.Status.Conditions.MarkTrue(
+						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyCondition,
+						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyMessage)
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, &baremetal)).To(Succeed())
```
gibizer, line 105: That is OK to set as the OpenStackBaremetalSet is created by the dataplane-operator but expected to be reconciled by the baremental operator. As the baremetal operator is not running in this test env you have to simulate its behavior form the test.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+						condition.Type(fmt.Sprintf(dataplanev1.NodeSetServiceDeploymentReadyCondition, nodeSet.Name, "download-cache")),
+						condition.RequestedReason,
+						condition.SeverityInfo,
+						fmt.Sprintf(dataplanev1.NodeSetServiceDeploymentReadyMessage, nodeSet.Name, "download-cache"))
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, &nodeSet)).To(Succeed())
+				}).Should(Succeed())
+
+				// Set baremetal provisioning conditions to True
+				Eventually(func(g Gomega) {
+					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &baremetal)).To(Succeed())
```
gibizer, line 101: to avoid confusion I would not reuse the dataplaneNodeSetName as the name of the baremetalset. In reality the name of the two CRs are the same, but reading the test code is confusing this way.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+					if serviceName == "ovn" {
+						service.Spec.ConfigMaps = append(service.Spec.ConfigMaps, "ovncontroller-config")
+
+						Eventually(func(g Gomega) {
+							g.Expect(k8sClient.Update(ctx, service)).To(Succeed())
+						}, th.Timeout, th.Interval).Should(Succeed())
+					}
+					//Retrieve service AnsibleEE and set JobStatus to Successful
+					Eventually(func(g Gomega) {
+						// Make an AnsibleEE name for each service
```
gibizer, line 127: update the comment it is not creating things just simulating successful exection

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+						}
+						ansibleEE := &ansibleeev1.OpenStackAnsibleEE{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      ansibleeeName.Name,
+								Namespace: ansibleeeName.Namespace,
+							}}
+						g.Expect(th.K8sClient.Get(th.Ctx, ansibleeeName, ansibleEE)).To(Succeed())
+						ansibleEE.Status.JobStatus = ansibleeev1.JobStatusSucceeded
+
+						g.Expect(th.K8sClient.Status().Update(th.Ctx, ansibleEE)).To(Succeed())
```
gibizer, line 140: again this is OK to do as AnsibleEE is reconciled by the ansibleee-operator wich does not run in this env.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
-			DeferCleanup(th.DeleteInstance, CreateDataplaneDeployment(dataplaneDeploymentName, DefaultDataPlaneDeploymentSpec()))
 			CreateSSHSecret(dataplaneSSHSecretName)
+			CreateSSHSecret(neutronOvnMetadataSecretName)
+			CreateSSHSecret(novaNeutronMetadataSecretName)
+			CreateSSHSecret(novaCellComputeConfigSecretName)
+			DeferCleanup(th.DeleteInstance, CreateDataplaneNodeSet(dataplaneNodeSetName, DefaultDataPlaneNodeSetSpec()))
+			DeferCleanup(th.DeleteInstance, CreateDataplaneDeployment(dataplaneDeploymentName, DefaultDataPlaneDeploymentSpec()))
 		})
 
 		It("should have conditions set", func() {
```
gibizer, line 67: I would name this "should successfully deploy the requested services"

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+			Eventually(func(g Gomega) {
```
gibizer, line 68: I think this outermost Eventually block is not needed.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+
+				// Set node set deployment ready condition to False
+				Eventually(func(g Gomega) {
+					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &nodeSet)).To(Succeed())
+					nodeSet.Status.Conditions.MarkFalse(
+						condition.Type(fmt.Sprintf(dataplanev1.NodeSetServiceDeploymentReadyCondition, nodeSet.Name, "download-cache")),
+						condition.RequestedReason,
+						condition.SeverityInfo,
+						fmt.Sprintf(dataplanev1.NodeSetServiceDeploymentReadyMessage, nodeSet.Name, "download-cache"))
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, &nodeSet)).To(Succeed())
```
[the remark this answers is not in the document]
gibizer: To be more precise: you should not manipulate the Status of a CR that is reconciled by the operator under test.

tests/functional/openstackdataplanedeployment_controller_test.go
gibizer: I was wrong about the latter. The existence of the secret is important that dataplane looks into the Secret to to the volume mount for AnsibleEE. (Probably can be removed from the impl but that is a separate issue from this PR)

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+			Eventually(func(g Gomega) {
+				nodeSet := dataplanev1.OpenStackDataPlaneNodeSet{}
+				baremetal := baremetalv1.OpenStackBaremetalSet{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      nodeSet.Name,
+						Namespace: nodeSet.Namespace,
+					},
+				}
+				// Create config map for OVN service
+				Eventually(func(g Gomega) {
```
gibizer, line 77: You dont need Eventually here, the CreateConfigMap helper already has one

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+
+				// Set baremetal provisioning conditions to True
+				Eventually(func(g Gomega) {
+					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &baremetal)).To(Succeed())
+					baremetal.Status.Conditions.MarkTrue(
+						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyCondition,
+						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyMessage)
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, &baremetal)).To(Succeed())
+
+				}).Should(Succeed())
```
gibizer, line 107: use the the timeout and interval value here so it can be finetuned globally

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+
+				// Create all services necessary for deployment
+				for _, serviceName := range nodeSet.Spec.Services {
+					dataplaneServiceName := types.NamespacedName{
+						Name:      serviceName,
+						Namespace: namespace,
+					}
+					CreateDataplaneService(dataplaneServiceName)
```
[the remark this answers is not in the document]
gibizer: Yeah this can be removed

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+					service := GetService(dataplaneServiceName)
+					//OVN needs config map resource
+					if serviceName == "ovn" {
+						service.Spec.ConfigMaps = append(service.Spec.ConfigMaps, "ovncontroller-config")
+
+						Eventually(func(g Gomega) {
+							g.Expect(k8sClient.Update(ctx, service)).To(Succeed())
+						}, th.Timeout, th.Interval).Should(Succeed())
+					}
```
gibizer, line 124: You don't need to modify the service. It is pre-created by the dataplane-operator automatically based on the yaml files stored under config/services. You only need the ConfigMap to exists.

tests/functional/suite_test.go
```diff
@@ -68,7 +68,7 @@ const (
 	ContainerImage       = "test://nova"
 	timeout              = 40 * time.Second
```
gibizer, line 69: This is a fairly big number that slows down the test execution. But the deployment controller has a 15 seconds Requeue request at some point so this 40 sec is probably needed today. I encourage to a) check if Requeue is really needed there or it can be replaced with some watch functionality. b) if a) is not possible then I suggest to make that 15 seconds configurable so that we can override that from the test to be smaller.  These are obviously outside of the scope of the current PR.

tests/functional/openstackdataplanedeployment_controller_test.go
jpodivin: I'd rather keep the name as it is. After all I'm only checking states of conditions, not actual deployment. When the test is expanded to make assertions on other things it may be time to revisit the issue.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,23 +38,116 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+				// Set baremetal provisioning conditions to True
+				Eventually(func(g Gomega) {
+					g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &baremetal)).To(Succeed())
+					baremetal.Status.Conditions.MarkTrue(
+						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyCondition,
+						baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyMessage)
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, &baremetal)).To(Succeed())
+
+				}).Should(Succeed())
+
```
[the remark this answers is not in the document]
jpodivin: I think that would be better left for the follow up.

tests/functional/suite_test.go
jpodivin: ACK

tests/functional/base_test.go
```diff
@@ -157,13 +166,15 @@ func SimulateIPSetComplete(name types.NamespacedName) {
...
 		g.Expect(th.K8sClient.Get(th.Ctx, name, IPSet)).Should(Succeed())
+		gateway := "172.20.12.1"
 		IPSet.Status.Reservation = []infrav1.IPSetReservation{
 			{
 				Address: "172.20.12.76",
 				Cidr:    "172.20.12.0/16",
 				MTU:     1500,
 				Network: "CtlPlane",
 				Subnet:  "subnet1",
+				Gateway: &gateway,
```
gibizer, line 177: If you want you can use `Gateway: ptr.To("172.20.12.1"),`

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -53,11 +102,77 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+			// Set node set deployment ready condition to False
+			Eventually(func(g Gomega) {
+				g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &nodeSet)).To(Succeed())
+				nodeSet.Status.Conditions.MarkFalse(
+					condition.Type(fmt.Sprintf(dataplanev1.NodeSetServiceDeploymentReadyCondition, nodeSet.Name, "download-cache")),
+					condition.RequestedReason,
+					condition.SeverityInfo,
+					fmt.Sprintf(dataplanev1.NodeSetServiceDeploymentReadyMessage, nodeSet.Name, "download-cache"))
+				g.Expect(th.K8sClient.Status().Update(th.Ctx, &nodeSet)).To(Succeed())
+			}, th.Timeout, th.Interval).Should(Succeed())
```
gibizer, line 132: You can remove this Eventually block (you should never set Status on a CR that is reconciled by the operator under test). When you remove this you need to add to read back the nodeSet once before you start iterating the services of it to have the defaulted service list in its Spec.
```
			nodeSet = *GetDataplaneNodeSet(dataplaneNodeSetName)
```

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -30,15 +41,53 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
 		Expect(err).NotTo(HaveOccurred())
 	})
 
 	When("A dataplaneDeployment is created with matching NodeSet", func() {
 		BeforeEach(func() {
-			DeferCleanup(th.DeleteInstance, CreateDataplaneNodeSet(dataplaneNodeSetName, DefaultDataPlaneNoNodeSetSpec()))
-			DeferCleanup(th.DeleteInstance, CreateDataplaneDeployment(dataplaneDeploymentName, DefaultDataPlaneDeploymentSpec()))
 			CreateSSHSecret(dataplaneSSHSecretName)
+			th.CreateSecret(neutronOvnMetadataSecretName, map[string][]byte{
+				"authorized_keys": []byte("blih"),
```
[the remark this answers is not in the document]
jpodivin: Actually, it will. We are using this value in our controller as on of the secrets.[0] If spec isn't declared preprovisioned the controller checks for this key. If it can't find it the reconciliation loop fails. 

[0] https://github.com/openstack-k8s-operators/dataplane-operator/blob/f3a157ac547c3290f0186b02eab4534857161d0d/controllers/openstackdataplanenodeset_controller.go#L88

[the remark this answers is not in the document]
gibizer:
```
$ oc get OpenStackDataPlaneServices/ovn -o yaml
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneService
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"dataplane.openstack.org/v1beta1","kind":"OpenStackDataPlaneService","metadata":{"annotations":{},"name":"ovn","namespace":"openstack"},"spec":{"configMaps":["ovncontroller-config"],"label":"dataplane-deployment-ovn","openStackAnsibleEERunnerImage":"quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest","playbook":"osp.edpm.ovn","secrets":["neutron-ovn-metadata-agent-neutron-config","nova-metadata-neutron-config"]}}
  creationTimestamp: "2023-10-02T09:41:38Z"
  generation: 2
  name: ovn
  namespace: openstack
  resourceVersion: "75183"
  uid: e48a079e-484d-4ce4-933b-6a0b899794d5
spec:
  configMaps:
  - ovncontroller-config
  label: dataplane-deployment-ovn
  playbook: osp.edpm.ovn
  secrets:
  - neutron-ovn-metadata-agent-neutron-config
  - nova-metadata-neutron-config
```

```
$ oc get Secret/neutron-ovn-metadata-agent-neutron-config -o yaml
apiVersion: v1
data:
  10-neutron-metadata.conf: IyBOT1RFOiBzb21lIG5lY2Vzc2FyeSBtZXRhZGF0YSBhZ2VudCBvcHRpb25zIGFyZSBzZXQgYnkgbm92YS4KIyBTcGVjaWZpY2FsbHksIG5vdmFfbWV0YWRhdGFfaG9zdCBhbmQgbWV0YWRhdGFfcHJveHlfc2hhcmVkX3NlY3JldC4KIwojIFRoZSBmaW5hbCBtZXRhZGF0YSBhZ2VudCBkZXBsb3ltZW50IHdpbGwgaGF2ZSB0byBtZXJnZSBzZXR0aW5ncyBmcm9tIG11bHRpcGxlCiMgc291cmNlcy4KCltvdm5dCm92bl9zYl9jb25uZWN0aW9uID0gdGNwOjE3Mi4xNy4wLjMwOjY2NDIK
kind: Secret
metadata:
  creationTimestamp: "2023-10-02T09:23:08Z"
  labels:
    neutron.openstack.org/name: neutron
    neutron.openstack.org/namespace: openstack
    neutron.openstack.org/uid: 6d46429a-208e-4b5f-92b1-9bd2e3a97696
  name: neutron-ovn-metadata-agent-neutron-config
  namespace: openstack
  ownerReferences:
  - apiVersion: neutron.openstack.org/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: NeutronAPI
    name: neutron
    uid: 6d46429a-208e-4b5f-92b1-9bd2e3a97696
  resourceVersion: "58082"
  uid: 2591ec65-65cb-439f-b32b-042fa7147abe
type: Opaque
```
No, It has a `10-neutron-metadata.conf` key.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -53,11 +102,68 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+				// OpenStackBaremetalSet has the same name as OpenStackDataPlaneNodeSet
+				g.Expect(th.K8sClient.Get(th.Ctx, dataplaneNodeSetName, &baremetal)).To(Succeed())
+				baremetal.Status.Conditions.MarkTrue(
+					baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyCondition,
+					baremetalv1.OpenStackBaremetalSetBmhProvisioningReadyMessage)
+				g.Expect(th.K8sClient.Status().Update(th.Ctx, &baremetal)).To(Succeed())
+
+			}, th.Timeout, th.Interval).Should(Succeed())
+
+			// Create all services necessary for deployment
```
gibizer, line 136: nit: you are not creating service here just simulating successful AnsibleEE execution

slagle (approved):

PR #426, opened 2023-09-22
Update network-config-template to new format

bshephar:
This commit updates the sample files to use the new format that will accept Jinja2 files instead of a file path. We maintain the edpm_network_config_override variable for now to be backwards compatible with other projects that depend on it existing.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/362
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/581


bshephar:
recheck

PR #428, opened 2023-09-22
Remove deployStrategy from samples and docs

rabi:
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/546


fao89 (approved):

slagle (commented):
+2.

This probably shouldn't have the lgtm label until the install_yamls actually merges. Otherwise, this one could merge first which would break local install_yamls environment since the samples are taken from a git checkout of dataplane-operator.

rabi:
recheck

depends-on merged.

PR #429, opened 2023-09-23
Run download-cache after configure-network

rabi:
We do gowvol in the configure-network service and we should do image downloads after that. This currently breaks the baremetal job as there is not enough space to create dirs as required.

Error: writing blob: adding layer with blob
"sha256:f39cd2eaaf10d896f872d413579dd8415bff1e3db4b0ed9476fb92cc5cf3ebb0": processing tar file(mkdir /etc/unbound: no space left on device): exit status 1


rabi:
recheck

RETRY_LIMIT

bshephar (approved):

PR #431, opened 2023-09-23
Add functional tests for Ansible inventory values

bshephar:
This change adds functional tests to validate the ansible_user override capabilities along with tests to ensure IPAM related variables are populated as expected.

### DIFF
--- a/tests/functional/base_test.go
+++ b/tests/functional/base_test.go
@@ -4,12 +4,14 @@ import (
 	"fmt"
 
 	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 
 	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
+	infrav1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 )
 
@@ -65,6 +67,12 @@ func CustomServiceImageSpec() map[string]interface{} {
 	}
 }
 
+func CreateNetConfig(name types.NamespacedName, spec map[string]interface{}) *unstructured.Unstructured {
+	raw := DefaultNetConfig(name, spec)
+	return th.CreateUnstructured(raw)
+}
+
+// Create SSHSecret
 func CreateSSHSecret(name types.NamespacedName) *corev1.Secret {
 	return th.CreateSecret(
 		types.NamespacedName{Namespace: name.Namespace, Name: name.Name},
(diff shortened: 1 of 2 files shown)


pkg/deployment/inventory.go
```diff
@@ -92,23 +92,14 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
...
 					dnsSearchDomains = append(dnsSearchDomains, res.DNSDomain)
+					host.Vars["dns_search_domains"] = dnsSearchDomains
 				}
-
-				host.Vars["ansible_user"] = getAnsibleUser(instance, nodeName)
-				host.Vars["ansible_port"] = getAnsiblePort(instance, nodeName)
-				host.Vars["management_network"] = getAnsibleManagementNetwork(instance, nodeName)
-				host.Vars["networks"] = getAnsibleNetworks(instance, nodeName)
-
-				ansibleVarsData, err := getAnsibleVars(helper, instance, nodeName)
```
bshephar, line 102: Does this still provide as much value now that we're using a single inventory file from the nodeSet?

Group vars should apply to the member nodes, so duplicating all values onto the node itself seems redundant?

bshephar:
recheck

bshephar:
recheck

bshephar:
recheck

pkg/deployment/inventory.go
fao89: I think we need this for the cases where we need to set some specific variable for a specific node

pkg/deployment/inventory.go
```diff
@@ -92,34 +92,19 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
...
 
 		err = resolveNodeAnsibleVars(&node, &host, &ansible.Group{})
 		if err != nil {
 			return "", err
 		}
-
-		ipSet, ok = allIPSets[nodeName]
-		if ok {
-			populateInventoryFromIPAM(&ipSet, host, dnsAddresses)
-		}
```
fao89, line 121: not sure if we should drop this, wdyt @rabi ?

rabi: No this should be the last and can't be removed. Seems this file is really messed up with the call to populateInventoryFromIPAM() twice and the lots of code duplicated inline.  This needs cleanup.

rabi (changes requested):

slagle (commented):
Instead of having to set the user on each node's vars, why don't we just set it in the group vars for the NodeSet? If it's set on the node, then set it that nodes host vars.

bshephar:
> Instead of having to set the user on each node's vars, why don't we just set it in the group vars for the NodeSet? If it's set on the node, then set it that nodes host vars.

We need to check it anyway right? Like, if a user overrides the ansible_user on just one of the nodes, we still need to check all of them, so we could do that in addition but we would still be doing an equal amount of work:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/inventory.go#L187-L192

So probably in addition but not instead of?

pkg/deployment/inventory.go
bshephar: Ok, I left the last one, removed the first one and all of the duplicated logic that was in this function.

pkg/deployment/inventory.go
bshephar: This would take care of the individual node overrides though?

https://github.com/openstack-k8s-operators/dataplane-operator/pull/431/files#diff-874aae70e738f15dc1bf30750cee0e87e3278c5c02d37e35c0943ca6c29393bbR70

pkg/deployment/inventory.go
rabi: Proposed https://github.com/openstack-k8s-operators/dataplane-operator/pull/437 to cleanup and simplify the whole thing.

fao89:
should we unhold?

slagle (approved):

PR #434, opened 2023-09-25
Revert "Run download-cache after configure-network"

rebtoor:
Reverts openstack-k8s-operators/dataplane-operator#429

We need to download packages for several roles referenced in the `configure_network` playbook so the `download-cache` playbook should be executed before.

Obviously the problem of the free disk space remains, so I added `edpm_growvols` role as dependency of the `download-cache` role to avoid it.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/369


rabi:
> We need to download packages for several roles referenced in the configure_network playbook so the download-cache playbook should be executed before.

I'm not really convinced  that we need to download all packages for all services in separate service.  I can understand pre-fetching containers, but not install all rpms beforehand. There can custom roles and custom services and I would prefer the packages are installed when those services are configured.

slagle:
> > We need to download packages for several roles referenced in the configure_network playbook so the download-cache playbook should be executed before.
> 
> I'm not really convinced that we need to download all packages for all services in separate service. I can understand pre-fetching containers, but not install all rpms beforehand. There can custom roles and custom services and I would prefer the packages are installed when those services are configured.

It's not install all rpms beforehand, it's just download. The isolation of the download vs the install has 2 purposes:

1. Clearly be able to measure the performance of download vs install for performance testing
2. Download and cache all artifacts (rpms, container images, etc) before an update/upgrade so these tasks can happen outside the maintenance window.

This functionality will need to be enhanced to work per-service, depending on which services are enabled. I believe @rebtoor is tracking this work for Q4.

rabi:
> 1. Clearly be able to measure the performance of download vs install for performance testing

We're splitting every package install task into two tasks anyway, won't profiling those tasks give what we want for performance testing?  Have we seen these downloads taking lot of deployment time when run with `free` strategy?  I think large number of nodes downloading the world at the same time can cause other issues, eventually delaying the whole thing (with multiple retries), though I understand we plan to do it outside the maintenance window.  Also, if we're refactoring these plays, I would like we move `edpm_bootstrap`, `edpm_kernel`, `edpm_tuned` etc from `configure_network` as well.

rebtoor:
> Also, if we're refactoring these plays, I would like we move `edpm_bootstrap`, `edpm_kernel`, `edpm_tuned` etc from `configure_network` as well.

Yeah, that's the idea.

slagle:
> > 1. Clearly be able to measure the performance of download vs install for performance testing
> 
> We're splitting every package install task into two tasks anyway, won't profiling those tasks give what we want for performance testing? Have we seen these downloads taking lot of deployment time when run with `free` strategy? I think large number of nodes downloading the world at the same time can cause other issues, eventually delaying the whole thing (with multiple retries), though I understand we plan to do it outside the maintenance window. Also, if we're refactoring these plays, I would like we move `edpm_bootstrap`, `edpm_kernel`, `edpm_tuned` etc from `configure_network` as well.

We're not using the free strategy. It makes it harder to debug the ansible output. We're sticking with the default ansible strategy, linear.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
```diff
@@ -12,7 +12,6 @@ spec:
   services:
     - configure-network
     - validate-network
-    - download-cache
```
slagle, line 15: why is not listed before configure-network?

slagle:
> > We're not using the free strategy. It makes it harder to debug the ansible output. We're sticking with the default ansible strategy, linear.
> 
> Oh.. I just saw [openstack-k8s-operators/openstack-ansibleee-operator#108](https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/108). Are the issues with `free` strategy documented somewhere? Now that we moved to not reconciling individual nodes, this could probably be a scaling issue in the future.

"free" provided some benefit with TripleO since there were several task includes that were dynamic based on the THT role.  There were several tasks that also loaded the task file dynamically based on the role name (include_task with the task file name dynamically set based on the THT role). "free" helped here some since each node could then run whatever tasks were in those task files, and everything wasn't in lock step. Controllers could be running  their tasks in parallel with Compute nodes, and the tasks were different.

That came at the expense of readability and ease of debugging activities. That was one of the complaints from TripleO users in that even though it was using ansible, you still couldn't tell what was actually going on.

We don't have that pattern in edpm. Each NodeSet is its own ansible execution, and the tasks are the same across a NodeSet. Thus the thinking is that we are just better off sticking with the ansible default linear which produces output and logs that folks are used to seeing with ansible executions.

rabi:
> We don't have that pattern in edpm. Each NodeSet is its own ansible execution, and the tasks are the same across a NodeSet

OK,  so we think we could scale with thousands of nodes in a NodeSet  with linear strategy where we would wait for all nodes to complete a task before moving to the next. We probably should do some scale testing to see the performance as we increase nodes in a nodeset and number of nodesets.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
rebtoor: it wasn't listed at all, probably i forgot to add it in the original PR and i think that other sample files could suffer of the same issue. I'll propose a followup PR to fix it after this one is get merged. WDYT?

slagle:
> > We don't have that pattern in edpm. Each NodeSet is its own ansible execution, and the tasks are the same across a NodeSet
> 
> OK, so we think we could scale with thousands of nodes in a NodeSet with linear strategy where we would wait for all nodes to complete a task before moving to the next. We probably should do some scale testing to see the performance as we increase nodes in a nodeset and number of nodesets.

No.  I don't think that at least. Ansible doesn't scale well with thousands of nodes with any configuration. We've shown that even when using free and tripleo_free in TripleO. You're also limited with the number of forks with either strategy. You are better off splitting it up into separate Ansible executions (multiple invocations of ansible-runner), which is what NodeSets enable.

Regardless, I don't think the recommendation will be to have thousands of nodes in the same NodeSet, even if they are configured similarly, which they most likely wouldn't be anyway (subnet and hardware differences).

config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml
slagle: Yes, please do a follow up. :+1:

slagle (approved):

slagle (approved):

PR #438, opened 2023-09-27
Remove networks from inventory

rabi:
This is not used and increases the size of inventory. Also, cleans up multiple loops to set vars.


fao89 (approved):

jpodivin (commented):
If we aren't using it, and there is no reasonable expectation of using it in the future, removal is a prudent move. Although it may have been better to split this in two patches, one for loop optimization and the other for the attribute removal. 

One last thing, are there any changes we should make to the docs?

PR #442, opened 2023-09-28
Initial commit for EDPM certs

vakwetu:
This is completely untested(although it does compile!), but it gives an idea of how EDPM certs could be generated and mounted for each of the services that need them.

While I will test and refine, I'm hoping that this PR starts conversations and questions so that we can decide on an implementation.

The basic idea is as follows:
1. After the DNS data is created in the openstackdataplanenodeset_controller, a call is made to GenerateTLSCerts to create all the certs for all the nodes in the deployment.
2. GenerateTLSCerts right now only generates a single cert for each node in the nodeset using the DNSNames for that node.  As the cert contains the DNSNames for all the node's interfaces, it should be usable on all interfaces.  Its likely that most services will be able to use that cert (which has the name "nodeName-cert")
3. You can easily see though how the code could be extended to create a cert with all the IPS for instance, or to use a different issuer, depending of the needs of the service.
4. The secret containing the cert will need to have an ownership reference updated so that we can trigger a reconcile when the cert changes or is renewed.
5. We create a secret per service (for those that need certs) that contains all the certs for that service on all the nodes.  This secret has the name <nodeSet.Name>-<service>-certs.  This secret will be mounted in the ansibleEE pod as an extraMount when the service is deployed.  This secret contains keys of the form nodeName-tls.key, nodeName-tls.crt, nodeName-ca.crt.
6. We create a new secret containing all the node certs because there could be hundreds of nodes.  I doubt we can have hundred of secrets. Instead, all the certs will be visible at /var/lib/openstack/service.name/certs
7. A new boolean is added to the Service spec (HasTLSCerts).  The secret containing the certs will only be mounted if HasTLSCerts=True


fultonj (commented):
This general approach looks good to me.

pkg/deployment/cert.go
```diff
@@ -0,0 +1,163 @@
...
+			Name: instance.Name + "-libvirt-certs",
+		},
+		Data: libvirtCertsData,
+	}
+	secret.CreateOrPatchSecret(ctx, helper, instance, libvirtCertsSecret)
+
+	return nil
+}
+
+func IssueTLSCert(ctx context.Context, helper *helper.Helper,
```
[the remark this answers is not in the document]
fultonj: > I think functions like this should go in `lib-common` so that we're able to generate certs in openstack-operator 

You have a good point about `lib-common` but I'm not sure that `openstack-operator` should be creating the certs for the EDPM nodes.

When would the openstack-operator know of the DNS names for the EDPM nodes? I think Ade is creating the certs in the dataplane-operator because when node sets are created, that is the moment the DNS names are known, and thus the certs can be created. I'm not sure it could be earlier in the deployment process.

Now, there is certificate related code in lib-common that the dataplane-operator could call.

  https://github.com/openstack-k8s-operators/lib-common/tree/main/modules/certmanager

Both this `cert.go` and the above `certificate.go` in lib-common use `github.com/cert-manager` so maybe certificate.go in lib-common should be used instead to create certificates and what's in the dataplane operator should be calling that code. I haven't looked at it closely  but  both seem to have certificate creating functions.

So I'm fine with the calls to generate the certs for edpm nodes, being made in the dataplane operator, though I agree it's worth considering using lib-common.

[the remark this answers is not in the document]
fmount: not saying that `openstack-operator` should create certs for the `edpm` nodes, sorry if I didn't write it properly. My only point is that some logic (e.g., cert generation) is required in the ctlplane, where you want to Create certs and use them to secure the generated Routes [1]. The same certs are required to re-encrypt the traffic from the Route to the service operator, hence the openstack operator should provide them to the service operators.
In terms of edm nodes creation/generation, I agree that the entire process should be driven by the `dataplane-operator`, that could call the same functions provided by `lib-common` with different input coming from the node sets definition. 

[1] https://docs.openshift.com/dedicated/networking/routes/secured-routes.html

pkg/deployment/cert.go
```diff
@@ -0,0 +1,163 @@
...
+	//
+	// The convention we use here is "<nodeset.name>-<service>-certs", so for example,
+	// openstack-epdm-nova-certs.
+	novaCertsData := map[string][]byte{}
+	libvirtCertsData := map[string][]byte{}
+
+	// for each node in the nodeset, issue all the TLS certs needed based on the
+	// ips or DNS Names
+	for nodeName, _ := range instance.Spec.Nodes {
+		var ips []string
```
gibizer, line 55: `ips` seems to be unused below other than appending to it

pkg/deployment/cert.go
```diff
@@ -0,0 +1,163 @@
...
+	cert := certmanager.NewCertificate(certReq, 5)
+	_, err = cert.CreateOrPatch(ctx, helper)
+	if err != nil {
+		err = fmt.Errorf("Error issuing certificate %s - %w", name, err)
+		return nil, err
+	}
+
+	certSecret, _, err := secret.GetSecret(ctx, helper, secretName, namespace)
+	if err != nil {
+		err = fmt.Errorf("Error retrieving secret %s - %w", secretName, err)
```
gibizer, line 158: nit: there are two cases:
1) the Secret is being created by the cert-manager so we got NotFound here so we just need to wait with a RequeueAfter
2) we could not reach the API or the API failed in other reasons. A real error.

I would split the two cases as 1) is not an error for me

pkg/deployment/const.go
```diff
@@ -47,4 +47,10 @@ const (
 
 	// ConfigPaths base path for volume mounts in OpenStackAnsibleEE pod
 	ConfigPaths = "/var/lib/openstack/configs"
+
+	// RootCAIssuerInternalLabel for internal RootCA to issue internal TLS Certs
+	RootCAIssuerInternalLabel = "osp-rootca-issuer-internal"
```
gibizer, line 52: Do we need to support user specified CA cert or root? I vaguely remember discussing this with @stuggi before.

pkg/deployment/deployment.go
```diff
@@ -326,6 +327,35 @@ func addServiceExtraMounts(
 
 		aeeSpec.ExtraMounts = append(aeeSpec.ExtraMounts, volMounts)
 	}
+
+	// Add mount for TLS certs
+	if service.Spec.HasTLSCerts {
+		volMounts := storage.VolMounts{}
+		secretName := nodeSet.Name + "-" + service.Name + "-certs"
```
gibizer, line 334: Lets move this name generation into a function so it is kept in sync between the place we create the secret and the place we use the secret

pkg/deployment/deployment.go
```diff
@@ -326,6 +327,35 @@ func addServiceExtraMounts(
...
+		sec := &corev1.Secret{}
+		err := client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: service.Namespace}, sec)
```
gibizer, line 336: This will fails for services other than nova and libvirt as the cert secret is only created for those services but each DataPlaneService can have HasTLSCerts: true configured.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -66,6 +66,11 @@ type OpenStackDataPlaneServiceSpec struct {
 	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
 	// +kubebuilder:validation:Optional
 	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty" yaml:"openStackAnsibleEERunnerImage,omitempty"`
+
+	// HasTLSCerts - Whether the nodes have TLS certs (True)
```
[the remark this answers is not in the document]
vakwetu: Sorry - True doesn't mean anything here.  I'll take it out.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -66,6 +66,11 @@ type OpenStackDataPlaneServiceSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	HasTLSCerts bool `json:"hasTLSCerts,omitempty"`
```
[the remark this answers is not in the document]
vakwetu: Ack - will do.

pkg/deployment/deployment.go
vakwetu: Ack, good idea.

pkg/deployment/cert.go
```diff
@@ -0,0 +1,163 @@
...
+		// For now we just add the hostname so we can select all the certs on one node
+		labels := map[string]string{
+			"hostname": nodeName,
+		}
+
+		ipSet, ok := allIPSets[nodeName]
+		if ok {
+			for _, res := range ipSet.Status.Reservation {
+				ips = append(ips, res.Address)
+				fqdnName := strings.Join([]string{nodeName, res.DNSDomain}, ".")
```
rabi, line 70: These are already available generated for dnsdata[1], we probbaly should not go through all IPSets and build them again. Can return a dict from there that can be used in nodeset controller when calling GenerateTLSCerts()

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/ipam.go#L87

pkg/deployment/cert.go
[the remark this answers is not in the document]
vakwetu: agreed.  this function should likely go into lib-common.  I'll discuss further with @stuggi

pkg/deployment/cert.go
vakwetu: ack

pkg/deployment/deployment.go
vakwetu: yeah -- I had simply put in nova and libvirt as examples.  But I've rewritten things so that any service that has hasTLSCerts: True will have certs created for them.  See the next version.

vakwetu: see the changes in controllers/openstackdataplanenodeset_controller.go

pkg/deployment/const.go
vakwetu: Thats a good question, but I think that at least initially the answer to this is no.  The compute nodes do not have a public interface, so all the interfaces are internal.  All internal certs right now are being served from a well-defined osp-rootca-issuer

vakwetu: @stuggi  ^^

pkg/deployment/cert.go
vakwetu: ack.  this was supposed to be illustrative in case ovn , for instance, needed to get certs with ips in them.  I've removed in the latest version

pkg/deployment/cert.go
```diff
@@ -0,0 +1,163 @@
...
+	allIPSets map[string]infranetworkv1.IPSet) error {
+
+	// We need to create a secret for each service that contains all the certs needed for
+	// that service for all the nodes in the nodeset.  This one secret will be mounted in
+	// the ansibleEE pod as an extra mount when that service is deployed.
+	//
+	// The convention we use here is "<nodeset.name>-<service>-certs", so for example,
+	// openstack-epdm-nova-certs.
+	novaCertsData := map[string][]byte{}
+	libvirtCertsData := map[string][]byte{}
```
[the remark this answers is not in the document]
vakwetu: done in latest version.

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -213,6 +213,13 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 	instance.Status.DNSClusterAddresses = dnsClusterAddresses
 	instance.Status.CtlplaneSearchDomain = ctlplaneSearchDomain
 
+	logger.Info("ADE: issue TLS Certs")
+	err = deployment.GenerateTLSCerts(ctx, helper, instance, allIPSets)
```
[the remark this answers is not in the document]
vakwetu: fixed in latest version

pkg/deployment/cert.go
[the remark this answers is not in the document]
stuggi: we already have one in lib-common https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L137

pkg/deployment/deployment.go
stuggi: fyi, there is now a tls pkg https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/tls/tls.go we use on the ctlplane side to mount the service/ca certs to the deployment, then using kolla_start to e.g. add the ca cert to the ca store. not sure if it can be used as is on the edpm side.

pkg/deployment/const.go
stuggi: right, for internal we do not plan (at least right now) to support custom CA, we do for public. We could do the same for internal if we see there is the need. I'll have added this label in lib-common https://github.com/openstack-k8s-operators/lib-common/pull/356

pkg/deployment/const.go
```diff
@@ -47,4 +47,13 @@ const (
 
 	// ConfigPaths base path for volume mounts in OpenStackAnsibleEE pod
 	ConfigPaths = "/var/lib/openstack/configs"
+
+	// RootCAIssuerInternalLabel for internal RootCA to issue internal TLS Certs
+	RootCAIssuerInternalLabel = "osp-rootca-issuer"
+
+	// RootCAIssuerInternalNamespaceLabel for internal RootCA to issue internal TLS Certs
+	RootCAIssuerInternalNamespaceLabel = "openstack"
```
stuggi, line 55: do we need this? at least right now we don't do any cross namespace actions, so the issuer will be created in the same namespace as the CR.

pkg/deployment/cert.go
```diff
@@ -0,0 +1,176 @@
...
+			Name:      GetServiceCertsSecretName(instance, serviceName),
+			Namespace: instance.Namespace,
+		},
+		Data: certsData,
+	}
+	_, _, err := secret.CreateOrPatchSecret(ctx, helper, instance, serviceCertsSecret)
+	if err != nil {
+		err = fmt.Errorf("Error creating certs secret for %s - %w", serviceName, err)
+		return err
+	}
```
stuggi, line 116: wondering if we have to create a single with the certs, or just could mount certs from several cert secrets based on the label using the `secretTamplate` as in https://cert-manager.io/docs/usage/certificate/ . would just have to update https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L167 for having the `secretTemplate`

Deydra71: In my opinion when we pursue the second option, then it means that we can mount only the certs that are needed. Which is in general safer approach.  The `serviceCertsSecret` aggregates all the individual node certificates into one secret, is this correct? Having one secret with all certificates for a service across all nodes is a point of potential issue.

stuggi: I submitted https://github.com/openstack-k8s-operators/lib-common/pull/358 so that annotations/labels can be passed for the `secretTemplate`

vakwetu: As I understand the way things work, there is a single ansibleEE pod that is created for a service, which installs the service on every node in the nodeset.  That means that if there are 1000 nodes in the nodeset, then there are 1000 secrets to be mounted.  Is there a limit to the number of mounts that a pod can have?  Does it make sense to have 1000 mounts?   I created this single secret because I was concerned with not breaching that limit.  

But, I also realize that there is a limit in the size of the secret (4M?)  So, how to avoid that limit?

* We can also create multiple secrets (with as many certs as can fit in a secret) - and then just mount all these secrets as volumes. 

* Another alternative is to create a single volume per service and copy all the certs and keys into that volume.  The main concern there though is that this volume will contain a bunch of secret stuff - ie. keys.  We can specify that this volume is ephemeral and should be deleted - but we don't have any control of when that happens.
 
So, I'm looking for guidance here.  What is the expectation of the size of a nodeset?

pkg/deployment/cert.go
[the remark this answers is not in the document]
vakwetu: Great.  I'll use that one.
That said  - when anyone (maybe ovn?) needs it - we'll need to add support for IPs instead of just hostnames

pkg/deployment/const.go
vakwetu: removed.

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
gibizer: This still needs to be addressed

pkg/deployment/cert.go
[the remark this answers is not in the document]
vakwetu: Latest code uses EnsureCerts in lib-common

pkg/deployment/cert.go
vakwetu: looks resolved to me.

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
vakwetu: should be addressed in this latest version

pkg/deployment/const.go
vakwetu: using the merged label now

pkg/deployment/cert.go
vakwetu: Yup - this is where I originally saw the code I used.

I'd have to return all the dnsNames and all the ips for each node - because its likely we'll need both.
We can do as you suggest, but it does make the code a little less readable IMHO.  Is it really that much of a performance improvement to go through the ips again?

pkg/deployment/cert.go
vakwetu: Based on some back of the envelope calculations, I figure we can end up with about 500 certs in a single secret.
That seems reasonable as an approach.

We do need to do some work where we create multiple secrets if we go over that amount -- well, really keep track of the secret size as we add certs - but I'd like to leave that to a new PR.

I will look into only removing the cacerts from the secret once the PRs with the cacert chains are merged.

stuggi:
+1 from me on this. looks good to me to start iterating on

pkg/deployment/cert.go
rabi: It's kind of duplicating the code. Also, using the `nodeName` to get the fqdn is not correct and we've changed that[1].  As a minimum we should have a helper function that's used at both places so that we keep them consistent.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/commit/14f9f9b236276d00d53d3991b7537faf9498fe44

vakwetu: Ok - I see that this code can diverge -- I've taken your suggestion in the next iteration.

vakwetu:
functional tests failing right now because the ca issuer doesn't exist - we need to add this into the ci framework.

vakwetu:
currently failing tests due to dependency on lib-common change

pkg/deployment/cert.go
bogdando: When we rotate a cert for a node, we should only touch its own secret. To avoid bind mounting housands of certs in large envs, let's use kolla to copy it in instead. This is how we could mitigate mounts count and secret size limits

vakwetu: @bogdando We're using a secret containing a bunch of other secrets here rather than a bind-mounted directory because we're concerned that the directory could stick around awhile after the ansible EE pod is completed, potentially exposing the secrets for all the compute nodes.  Secrets have some built-in protections.  They are stored encrypted in etcd, so they are safe at rest.  They are only sent to the nodes that need them and they are mounted using ramfs - so there is no place where they are written to disk.  The secret size limitation is because the data is stored in etcd.

We could potentially use a mounted  volume if that volume were encrypted in some way, but as far as I can tell, there is no openstack-style encrypted cinder volume available in k8s, where a volume is mounted with an encryptor/decryptor mechanism.  If you know of such a mechanism, I'd be happy to use it.  I did find mention of something like that here -- https://docs.portworx.com/portworx-enterprise/operations/key-management/kubernetes-secrets/pvc-encryption-using-annotations -- but not in openshift.

For openshift, there appear to be a couple of options. 
https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.6/html/planning_your_deployment/security-considerations_rhocs (configuring encryption at the container storage level)

https://red-hat-storage.github.io/ocs-training/training/ocs4/ocs4-encryption.html#_overview (integration with vault as a kms)

All of this is way beyond the scope of this patch :)

For the reconciliation loop issue, recall that there will be maybe 500 certs in a secret, so the number of extra reconciliations will be somewhat small.  Also, we can add a label to the secret of secrets and some code to tell the  reconciliation loop to ignore it.  We can handle that in a follow on patch when we test reconciations for cert changes.

 I propose we go with this mechanism for now - we know it works - and it provides the needed security without too much extra complexity.

vakwetu:
recheck

pkg/deployment/inventory.go
```diff
@@ -46,6 +46,12 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 		utils.LogErrorForObject(helper, err, "Could not resolve ansible group vars", instance)
 		return "", err
 	}
+
+	// add TLS ansible variable
+	if instance.Spec.TLSEnabled != nil && *instance.Spec.TLSEnabled {
+		nodeSetGroup.Vars["tls_certs_enabled"] = "true"
```
slagle, line 52: This is going to be a variable that multiple roles from edpm-ansible would use right? Could we call this edpm_tls_certs_enabled?

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -209,6 +209,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 	instance.Status.DNSClusterAddresses = dnsClusterAddresses
 	instance.Status.CtlplaneSearchDomain = ctlplaneSearchDomain
 
+	// Issue certs for TLS for services that need them
```
slagle, line 212: This probably needs to be refactored out into a function that can also be called from https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanedeployment_controller.go as that can use an entirely different set of services if servicesOverride is set.

In fact...maybe this logic should just live entirely in openstackdataplanedeployment_controller.go. I'm thinking that might be the way to go.

pkg/deployment/inventory.go
vakwetu: yes - we can do that

controllers/openstackdataplanenodeset_controller.go
vakwetu: ok - will refactor

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -76,6 +76,19 @@ type OpenStackDataPlaneServiceSpec struct {
 	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
 	// +kubebuilder:validation:Optional
 	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty" yaml:"openStackAnsibleEERunnerImage,omitempty"`
+
+	// HasTLSCerts - Whether the nodes have TLS certs
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	HasTLSCerts *bool `json:"hasTLSCerts,omitempty" yaml:"hasTLSCerts,omitempty"`
```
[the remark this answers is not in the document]
vakwetu: ok - will use TLSCertsEnabled

controllers/openstackdataplanenodeset_controller.go
vakwetu: actually -- I don't understand this comment entirely .. will sync with you,

PR #444, opened 2023-09-28
Add services to telemetry to expose node_exporter

jlarriba:


fao89 (approved):

slagle (approved):

jlarriba:
It seems that the KubeServices feature is broken, I will be properly debugging and fixing it.

jlarriba:
Failed to deploy due to the error solved by this PR: https://github.com/openstack-k8s-operators/install_yamls/pull/559. Will recheck later, when the workaround is merged.

jlarriba:
recheck

pkg/deployment/kube_service.go
```diff
@@ -34,18 +34,25 @@ func CreateKubeServices(
...
 		if err != nil {
 			return err
 		}
 
 		addresses := make([]string, len(nodeSet.Spec.Nodes))
+		i := 0
 		for _, item := range nodeSet.Spec.Nodes {
-			addresses = append(addresses, item.Ansible.AnsibleHost)
+			if len(item.Ansible.AnsibleHost) == 0 {
+				log.Info("Creating Services in an IPAM environment is not supported")
```
slagle, line 51: I don't think this log message is accurate, and I don't think the logic is right. If AnsibleHost is not set, we default the ansible var ansible_host to node.HostName in inventory.go.

Also, AnsibleHost not being set doesn't mean that IPAM is in use. And the log message doesn't make sense even if it did. If we're using IPAM, we'd *know* the IP. So why couldn't we create the service?

We also can't say, this just doesn't work when using IPAM. Most customers will use IPAM.

pkg/deployment/kube_service.go
```diff
@@ -34,18 +38,32 @@ func CreateKubeServices(
...
-		for _, item := range nodeSet.Spec.Nodes {
-			addresses = append(addresses, item.Ansible.AnsibleHost)
+		i := 0
+		for name, item := range nodeSet.Spec.Nodes {
+			namespacedName := &types.NamespacedName{
+				Name:      name,
+				Namespace: instance.GetNamespace(),
+			}
+			if len(item.Ansible.AnsibleHost) > 0 {
+				addresses[i], addressType = getAddressFromAnsibleHost(&item, namespacedName, helper)
```
slagle, line 58: If set, using AnsibleHost always overrides using IPSet. Except when AnsibleHost is just a hostname (short), then an IPSet is never checked with this logic.

pkg/deployment/kube_service.go
```diff
@@ -67,6 +85,59 @@ func CreateKubeServices(
...
+			return item.HostName, discoveryv1.AddressTypeFQDN
+		}
+		// No IP address or valid hostname found anywhere
+		helper.GetLogger().Info("Did not found a valid hostname or IP address")
+		return "", ""
+	}
+	// check that the reservations list is not empty
+	if len(ipset.Status.Reservation) > 0 {
+		// use the first IP in the reservations list. This is CtlPlane network.
+		return ipset.Status.Reservation[0].Address, discoveryv1.AddressTypeIPv4
```
slagle, line 107: Is CtlPlane always the first reservation? And is that always the right IP to use, even when using network isolation?

jlarriba: I dont have the answer to those questions, only more questions. 

If it is not the first reservation, we will get whatever is the first reservation, as this is not exactly important for the Service itself.

The second question is more of a functional one and impacts the previous one. I don't really know if using the CtlPlane connection is an issue or not. We can use the IP we want as long as it is accessible from the ControlPlane, the CtlPlane seemed the most reasonable, but at this point (until I get more information), it is largely irrelevant.

Maybe we can add a new parameter on the KubeService API so that the user itself is able to define the network the Service should use. Would that be reasonable?

pkg/deployment/kube_service.go
jlarriba: Maybe we want to search first always in IPSets if we have them and resource to NodeSet only when those are not available?

slagle: The ordering is ok. I don't have a strong opinion if one should take precedence over the other. But, if AnsibleHost is set, but you can't use it, then I'd suggest checking for IPSets at that point.

pkg/deployment/kube_service.go
slagle: Yes..I think adding network to KubeService makes sense. If not set, default to CtlPlane.

rabi:
One of the failures is as below in the baremetal job[1], that's why the nodeset did not reach ready status. I think we should look for the reason before merging this.

[1] https://logserver.rdoproject.org/44/444/819c673ff012808bf3989a2935bb09db880cd601/github-check/cifmw-crc-podified-edpm-baremetal/aab57f0/ci-framework-data/logs/crc/pods/openstack-operators_dataplane-operator-controller-manager-6bd5884bb6-f66px_226281a6-a72c-40f7-89b2-6d4498335a34/manager/0.log

`ipam","namespace":"openstack"},"namespace":"openstack","name":"openstack-edpm-ipam","reconcileID":"10c3a8ce-5b23-455a-af76-197e1e21484b","ObjectType":"*v1beta1.OpenStackDataPlaneDeployment","ObjectNamespace":"openstack","ObjectName":"openstack-edpm-ipam","error":"EndpointSlice.discovery.k8s.io \"node-exporter-0\" is invalid: [endpoints[0].addresses[0]: Invalid value: \"\": must be a valid IPv4 address, endpoints[0].addresses[0]: Invalid value: \"\": must be a valid IP address]","stacktrace":"github.com/openstack-k8s-operators/lib-common/modules/common/util.LogErrorForObject\n\t/go/pkg/mod/github.com/openstack-k8s-operators/lib-common/modules/common@v0.1.1-0.20230919113507-d74c2f31d216/util/log.go:65\ngithub.com/openstack-k8s-operators/dataplane-operator/controllers.(*OpenStackDataPlaneDeploymentReconciler).Reconcile\n\t/remote-source/controllers/openstackdataplanedeployment_controller.go:219\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:122\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:323\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:235"}`

jlarriba:
Thank you Rabi for the heads up Rabi, I am already working on it.

pkg/deployment/kube_service.go
jlarriba: I have rethinked this and I think that if we have IPSets and we allow the user to choose the network, it would be better to always look into IPSets first, as that option allows us to choose from the user's chosen network. If we do not have IPSets, then the network choice from the user is not important and we just will try to expose the AnsibleHost, as a backup.

slagle (commented):
This is looking good to me. I assume the latest commit addressed the CI failure in some way. Is that accurate?

jlarriba:
Yes, it addressed it because the baremetal CI works with ipam and ipsets so now the ips are being retrieved correctly

rabi:
> Yes, it addressed it because the baremetal CI works with ipam and ipsets so now the ips are being retrieved correctly

Both pre-provisioned[1] and baremetal jobs use IPAM. 

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L59-L71

fao89 (approved):

PR #448, opened 2023-09-30
Introduce HCI DataSet example and Update Ceph DataSet example

fultonj:
Update the Ceph OpenStackDataPlaneNodeSet sample to fit current conventions. In particular do not hard code IPs as it's not necessary and that data can be provided by IPAM. Also, use MTU 9000 for the storage network.

Introduce HCI OpenStackDataPlaneNodeSet sample which is like the updated Ceph example, except it adds the Storage Management network and uses MTU 9000 and has a shortended service list since HCI requires the list to be modified after the first run on EDPM Ansible so that the EDPM deployment can be paused so that
Ceph can be installed and then the rest of the EDPM deployment may be resolved.

Jira: [OSP-28997](https://issues.redhat.com//browse/OSP-28997)


abays (approved):
/lgtm

fmount (commented):
/lgtm

PR #455, opened 2023-10-04
Add servicesOverride to OpenStackDataPlaneDeployment

slagle:
servicesOverride is used to override the list of services executed by an
OpenStackDataPlaneDeployment. When set, instead of using the individual
list of services from each nodeSet listed in the deployment, the
servicesOverride list will be used instead. This allows for targeting
execution of specific services often for day 2 operations.


slagle:
@jistr This addresses the ovn upgrade example. Let me know if you think this would work.

jistr (commented):
Thanks, yes i think this does address the "targeted execution". Since Ansible is stateless (unlike operators), we may also want to put in some Ansible variables that will help us distinguish whether we are running a deploy vs. update vs. upgrade, unless we want to run the "most defensive and potentially downtime inducing" variant all the time.

cc @xbezdick @sathlan

jistr:
To be clear: the variables concern doesn't relate to this PR in particular. I think it's good to merge.

bshephar:
> Thanks, yes i think this does address the "targeted execution". Since Ansible is stateless (unlike operators), we may also want to put in some Ansible variables that will help us distinguish whether we are running a deploy vs. update vs. upgrade, unless we want to run the "most defensive and potentially downtime inducing" variant all the time.
> 
> cc @xbezdick @sathlan

I think you would use the targeted execution to run your update specific playbook though right? Would you need a variable to distinguish if you had update specific tasks that are only executed when you call them via this method?

bshephar (approved):

jistr:
For the OVN case maybe it could be a custom playbook. I think it gets more tricky with the rest of the upgrade though. In theory we could have an update/upgrade playbook to match each of the deployment playbooks, e.g. "nova_upgrade" instead of "nova", but then assembling a list of playbooks for upgrade that would correctly match what was used for deployment becomes somewhat of a burden on the operator. So from that perspective executing the same playbooks and just have a way to alter their behavior seems like nicer UX.

PR #456, opened 2023-10-06
Update nodeset status for existing non-Ready deployment

rabi:
This would update the status of nodeset when there is an excuting deployment.


bshephar (approved):
Looks good to me. This will be nice to inform users that we are working on their deployment.

PR #459, opened 2023-10-09
base64 encoded ovn-controller configmap data

bshephar:


jpodivin (changes requested):
Could you please explain in the commit why was this change necessary or what is it intended to accomplish?

bshephar:
> Could you please explain in the commit why was this change necessary or what is it intended to accomplish?

It wasn't sorry, I was just wondering how this is working in CI and whether the Kuttl test was actually testing it, because locally if I use that value I get:
```
TASK [osp.edpm.edpm_ovn : Load ovsdb configuration from file] ******************
fatal: [edpm-compute-0 -> localhost]: FAILED! => {"ansible_facts": {"edpm_ovn_configmap_facts": {}}, "ansible_included_var_files": [], "changed": false, "message": "/var/lib/openstack/configs/ovn/ovsdb-config must be stored as a dictionary/hash"}
```

I don't think base64 encoded would work at all in a configmap, so I was looking to see if kuttl failed with invalid values. It was just easier to run it through CI. Initially, I was just going to leave this PR in draft and delete it, but it didn't run kuttl while it was in draft.

Apologies for the noise. I'll send a proper PR with valid ovsdb-config data and a proper commit message. :)

bshephar:
> > Could you please explain in the commit why was this change necessary or what is it intended to accomplish?
> 
> It wasn't sorry, I was just wondering how this is working in CI and whether the Kuttl test was actually testing it, because locally if I use that value I get:
> 
> ```
> TASK [osp.edpm.edpm_ovn : Load ovsdb configuration from file] ******************
> fatal: [edpm-compute-0 -> localhost]: FAILED! => {"ansible_facts": {"edpm_ovn_configmap_facts": {}}, "ansible_included_var_files": [], "changed": false, "message": "/var/lib/openstack/configs/ovn/ovsdb-config must be stored as a dictionary/hash"}
> ```
> 
> I don't think base64 encoded would work at all in a configmap, so I was looking to see if kuttl failed with invalid values. It was just easier to run it through CI. Initially, I was just going to leave this PR in draft and delete it, but it didn't run kuttl while it was in draft.
> 
> Apologies for the noise. I'll send a proper PR with valid ovsdb-config data and a proper commit message. :)

In hindsight, it would never be tested by kuttl because we aren't actually executing these playbooks against any nodes so maybe it doesn't need to be valid.

It might be nice to have some valid examples of these configs in our documentation though I guess.

PR #460, opened 2023-10-09
Don't set vlan ansible var when it's nil

rabi:
We don't need to set the ansible vars when vlan is nil.


luis5tb:
Nice one! LGTM

abays (approved):
/lgtm

PR #461, opened 2023-10-09
Drop network related ansible vars

rabi:
They would be popluated from the networks defined in nodeset spec.


rabi:
recheck 

https://issues.redhat.com/browse/OSP-29534

slagle (approved):

PR #463, opened 2023-10-11
Replacing edpm_chrony_ntp_servers with timesync_ntp_servers

jpodivin:
The edpm_chrony role is getting removed, this variable has to be renamed in order to reflect the new reality.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/575

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
@@ -47,8 +47,8 @@ spec:
          service_net_map:
            nova_api_network: internal_api
            nova_libvirt_network: internal_api
-         edpm_chrony_ntp_servers:
-           - pool.ntp.org
+         timesync_ntp_servers:
+           - hostname: pool.ntp.org
          # edpm_network_config
          # Default nic config template for a EDPM compute node
          # These vars are edpm_network_config role vars
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
@@ -42,8 +42,8 @@ spec:
          service_net_map:
            nova_api_network: internal_api
            nova_libvirt_network: internal_api
-         edpm_chrony_ntp_servers:
-           - pool.ntp.org
+         timesync_ntp_servers:
+           - hostname: pool.ntp.org
          # edpm_network_config
          # Default nic config template for a EDPM compute node
          # These vars are edpm_network_config role vars
--- a/tests/kuttl/tests/dataplane-create-test/00-dataplane-create.yaml
+++ b/tests/kuttl/tests/dataplane-create-test/00-dataplane-create.yaml
@@ -43,8 +43,8 @@ spec:
          service_net_map:
            nova_api_network: internal_api
            nova_libvirt_network: internal_api
-         edpm_chrony_ntp_servers:
-           - clock.redhat.com
+         timesync_ntp_servers:
+           - hostname: clock.redhat.com
          # edpm_network_config
          # Default nic config template for a EDPM compute node
          # These vars are edpm_network_config role vars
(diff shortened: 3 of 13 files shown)


tests/kuttl/tests/dataplane-create-test/00-dataplane-create.yaml
```diff
@@ -41,7 +41,7 @@ spec:
          service_net_map:
            nova_api_network: internal_api
            nova_libvirt_network: internal_api
-         edpm_chrony_ntp_servers:
+         edpm_ntp_servers:
```
fultonj, line 44: My reading of https://github.com/openstack-k8s-operators/edpm-ansible/pull/373 makes
me think the correct way to override the default time server is this:
```
edpm_ntp_servers:
  - hostname: 'clock.redhat.com'
    max_poll: 10
    min_poll: 6
```
Will this list of items instead of a map work?

`clock.redhat.com vs {hostname: clock.redhat.com}`

jpodivin: Not really, the poll aregs have defaults set to 10/6 in timesync, just like edpm_chrony. We don't really need to set them unless we decide those don't work for some reason.

As for the `hostname`, this will be handled on the edpm-ansible side.

fultonj: I think you're saying that `edpm_ntp_servers:` ['clock.redhat.com']` will also work.

jpodivin: I'm not going to count on it. But I'll make the necessary changes in edpm-ansible, rather than here. That way we can have something that is compatible with both edpm_chrony and timesync, minimizing chances of breakage, later we can move toward more timesync-like form.

fultonj: I wouldn't expect `edpm_ntp_servers: ['clock.redhat.com']` to work so that's why I asked. IMO better to just force the new format rather than remain backwards compatible since none of this is marked stable yet. Else, we'll have to maintain the translation to support the old format but it won't be worth it IMO.

slagle: I don't understand the need for edpm_ntp_servers at all. Unless it is just an intermediate step? We shouldn't have EDPM variables wrapping variables from the system roles unless absolutely necessary for some reason. Just use the timesync_ntp_server variable directly.

slagle (changes requested):

tests/kuttl/tests/dataplane-create-test/00-dataplane-create.yaml
jpodivin: That can be done. But given that there may be other cases of the role being used I  thought it safer to denote clearly what this one var is supposed to be doing. If we just switch to the same name role uses, it won't be apparent at first glance.

jpodivin: Variable renamed.

fultonj (commented):
I'm +2 on this provided all concerns below are addressed.

https://github.com/openstack-k8s-operators/edpm-ansible/pull/373#pullrequestreview-1694636108

rebtoor:
recheck

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
```diff
@@ -48,7 +48,7 @@ spec:
            nova_api_network: internal_api
            nova_libvirt_network: internal_api
          edpm_chrony_ntp_servers:
```
fao89, line 50:
```suggestion
         timesync_ntp_servers:
```

config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
```diff
@@ -43,7 +43,7 @@ spec:
            nova_api_network: internal_api
            nova_libvirt_network: internal_api
          edpm_chrony_ntp_servers:
```
fao89, line 45:
```suggestion
         timesync_ntp_servers:
```

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
jpodivin: Right. The PR has been up for so long it has gone out of sync.

fao89 (approved):

fao89 (approved):

PR #464, opened 2023-10-12
Update HCI NodeSet with a pre/post Ceph Deployment example

fultonj:
Use servicesOverride feature of OpenStackDataPlaneDeployment to make HCI deployment examples more clear. Instead of having the human operator edit the services list before creating a second deployment, provide two deployment examples which each have the correct services list for before and after Ceph has been installed.

Jira: [OSP-29040](https://issues.redhat.com//browse/OSP-29040)

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanedeployment_post_ceph_hci.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanedeployment_post_ceph_hci.yaml
@@ -1,0 +1,16 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneDeployment
+metadata:
+  name: deployment-post-ceph
+spec:
+  nodeSets:
+    - openstack-edpm-hci
+  # Create this deployment after Ceph is deployed
+  # on EDPM nodes in an HCI scenario. Create a
+  # nova-custom-ceph service which uses a ConfigMap
+  # containing libvirt overrides for Ceph RBD.
+  servicesOverride:
+    - ceph-client
+    - ovn
+    - libvirt
+    - nova-custom-ceph
--- a/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml
@@ -1,0 +1,16 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneDeployment
+metadata:
+  name: deployment-pre-ceph
+spec:
+  nodeSets:
+    - openstack-edpm-hci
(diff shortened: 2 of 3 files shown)


config/samples/dataplane_v1beta1_openstackdataplanedeployment_post_ceph_hci.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - openstack-edpm-hci
+  # Create this deployment after Ceph is deployed
+  # on EDPM nodes in an HCI scenario. Create a
+  # nova-custom-ceph service which uses a ConfigMap
+  # containing libvirt overrides for Ceph RBD.
+  servicesOverride:
+    - ceph-client
+    - ovn
+    - libvirt
+    - nova-custom-ceph
```
abays, line 16: Just a heads-up -- there is already a proposal to rename `servicesOverride` to just `services`: https://github.com/openstack-k8s-operators/dataplane-operator/pull/465

We'll have to refactor the VA1 data plane CRs to accommodate the proposed changes as well

fultonj: > Just a heads-up -- there is already a proposal to rename `servicesOverride` to just `services`: #465

Thanks for pointing that out. I think the review of #465 might take some time to get resolved. I'd be happy to send in a follow up PR after #465 merges. I'd rather merge this to promote the new pattern for HCI with what works today and not block on if it will be called `services` or `servicesOverride`. If the name changes it's an easy fix I will take care of right away. Thus, I request this not be blocked by #465. 

> We'll have to refactor the VA1 data plane CRs to accommodate the proposed changes as well

I'll be sending a PR to https://github.com/openstack-k8s-operators/architecture to update VA1 today.

fultonj (commented):
/retest-required

config/samples/dataplane_v1beta1_openstackdataplanedeployment_post_ceph_hci.yaml
fultonj: > We'll have to refactor the VA1 data plane CRs to accommodate the proposed changes as well

https://github.com/openstack-k8s-operators/architecture/pull/12

fmount:
The job was previously killed during `make ceph` due to:

```
Error from server (InternalError): Internal error occurred: admission plugin "[security.openshift.io/SCCExecRestrictions](http://security.openshift.io/SCCExecRestrictions)" failed to complete validation in 13s
```

re-kicking to see if it's consistent.

fmount:
The job was previously killed during `make ceph` due to:

```
Error from server (InternalError): Internal error occurred: admission plugin "[security.openshift.io/SCCExecRestrictions](http://security.openshift.io/SCCExecRestrictions)" failed to complete validation in 13s
```

re-kicking to see if it's consistent.

PR #465, opened 2023-10-12
Refactor dataplane

rabi:
Consists of 2 commits:

- Drop service reconciliation from nodeset controller
    
    Services should be pre-req for the deployment and not
    nodeset. Also, they're currently overwritten after they
    are patched using `oc apply -f` which is not desirable.

- Refactor nodeset/deployment CRDs
    
    - Moves ansibleee related fields to deployment
    - Moves services to deployment

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/579


rabi:
recheck

api/v1beta1/openstackdataplanedeployment_types.go
```diff
@@ -41,10 +60,6 @@ type OpenStackDataPlaneDeploymentSpec struct {
 	// AnsibleSkipTags for ansible execution
 	// +kubebuilder:validation:Optional
 	AnsibleSkipTags string `json:"ansibleSkipTags,omitempty"`
-
-	// +kubebuilder:validation:Optional
-	// ServicesOverride list
-	ServicesOverride []string `json:"servicesOverride"`
```
slagle, line 47: Dropping the concept of a servicesOverride means it's on the user to remember the list of services for each NodeSet. The only place that would be stored is on a previous deployment. If the user wants to deploy again, they have to manually use the same list.

api/v1beta1/openstackdataplanedeployment_types.go
```diff
@@ -30,6 +32,23 @@ type OpenStackDataPlaneDeploymentSpec struct {
...
+
+	// +kubebuilder:validation:Optional
+	// NetworkAttachments is a list of NetworkAttachment resource names to pass to the ansibleee resource
+	// which allows to connect the ansibleee runner to the given network
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// ExtraMounts containing files which can be mounted into an Ansible Execution Pod
+	// +kubebuilder:validation:Optional
+	ExtraMounts []storage.VolMounts `json:"extraMounts,omitempty"`
+
```
slagle, line 51: Moving these fields here to Deployment means they can't be unique per NodeSet.

pkg/deployment/service.go
```diff
@@ -64,87 +58,3 @@ func GetService(ctx context.Context, helper *helper.Helper, service string) (dat
 	err := client.Get(ctx, types.NamespacedName{Name: service, Namespace: namespace}, foundService)
 	return *foundService, err
 }
-
-// EnsureServices - ensure the OpenStackDataPlaneServices exist
-func EnsureServices(ctx context.Context, helper *helper.Helper, instance *dataplanev1.OpenStackDataPlaneNodeSet) error {
```
slagle, line 69: If you drop this, then where are you creating the services? Are you expecting the user to do that, if so, where do they get the YAML files?

slagle (changes requested):
I don't think refactoring is necessarily the way to go. Personally, I think we need to focus on stabilizing what we enabled in Q3. We have not had time or feedback to really see if the changes we made in Q3 were effective.

Instead of moving fields, I'd see if we can enable new functionality in additive or enhancing ways. This is how I approached https://github.com/openstack-k8s-operators/dataplane-operator/pull/455.

I'd prefer to keep the NodeSet the declarative representation of the NodeSet, and the Deployment models the execution.  NodeSets are likely to be grouped by hardware, and in some cases there may be composable services to enable and configure that hardware. It makes sense to me to keep that information stored on the NodeSet services. That way if an OpenStackDataPlaneDeployment is triggered for that NodeSet, the right set of services is run for each unique NodeSet, where the services may be different.

With this proposed change, the set of services, mounts, networkAttachments are the same across all NodeSets for a Deployment. That really limits the flexibility of the Deployment. All nodes need to be the same, other than ansible vars.

As for the service reconciliation (EnsureServices), it seems you are expecting the user to do that going forward? That seems to be the case given the install_yamls PR (https://github.com/openstack-k8s-operators/install_yamls/pull/579). I'm also not in agreement with that approach. I don't see why dataplane-operator can't create the default services automatically. If we want to enable editing the default services after they're created, we could still allow that by dropping the part that patches the full service Spec if the service is already created. We could also have the NodeSet "own" the services (setControllerReference), so that each service could be customized per NodeSet if desired.

rabi:
> I don't think refactoring is necessarily the way to go. Personally, I think we need to focus on stabilizing what we enabled in Q3. We have not had time or feedback to really see if the changes we made in Q3 were effective.

I don't agree as this is not a big change that would make things unstable. PR is already in good shape and once https://github.com/openstack-k8s-operators/install_yamls/pull/579 merges the kuttl job would be green.


> I'd prefer to keep the NodeSet the declarative representation of the NodeSet, and the Deployment models the execution. 

So you want to store things like what networks would be attached to aee pod and what verbose level ansible would be run in the nodeset properties?


> It makes sense to me to keep that information stored on the NodeSet services. That way if an OpenStackDataPlaneDeployment is triggered for that NodeSet, the right set of services is run for each unique NodeSet, where the services may be different.

What set of services would be run should be deployment specific IMO. There is already the default set of services for a deployment as before.

> With this proposed change, the set of services, mounts, networkAttachments are the same across all NodeSets for a Deployment. That really limits the flexibility of the Deployment. All nodes need to be the same, other than ansible vars.

We would not use anything other than ctlplane as the network attached to aee pod. aee pod can't talk to openstack services using other networks with the current design using mcvlan. It has to use the openshift internal network. Also, I mentioned earlier services to be run are deployment specific and can change from time to time. We can still have the option to provide services per nodeset in deployment spec.

> As for the service reconciliation (EnsureServices), it seems you are expecting the user to do that going forward? That seems to be the case given the install_yamls PR (openstack-k8s-operators/install_yamls#579). I'm also not in agreement with that approach. I don't see why dataplane-operator can't create the default services automatically. If we want to enable editing the default services after they're created, we could still allow that by dropping the part that patches the full service Spec if the service is already created. We could also have the NodeSet "own" the services (setControllerReference), so that each service could be customized per NodeSet if desired.

It can't create those as it does not own them and still reconciles them from the image which is an anti-pattern and patching them externally has no impact. If nodeset owns the services with the full spec and then reconciles it, that would probably would ok. But I still say that what services has to be run should be deployment specific and users can run a subset of them when required.

> Dropping the concept of a servicesOverride means it's on the user to remember the list of services for each NodeSet

I don't understand the issue here. The default services in a nodeset and deployment (proposed) are the same, if they use a custom set services in the last run, then deployment CR would have that info in both cases.

> Moving these fields here to Deployment means they can't be unique per NodeSet.

These are mounts on the ansible pod right and what ansible expects? Why we want to tie them with nodeset? If we want to have different ones per nodeset we can still do that in deployment spec.

rabi:
> What "it" do you mean by "It can't"? The NodeSet controller is reconciling them today.

NodeSet controller does not own them neither watches them. If you patch the services it would reflect the modified CR in etcd. But when any nodeset CR is reconciled it would flip it back to the old definition in the dataplane-operator image. This is terrible design IMO.

> There are dataplane-operator defaults (The hardcoded list of defaults).  

NetworkAttachments are attached to aee and who owns the aee? I think it's the deployment controller[1] which also owns deployment.

> And the change makes the Deployment more similar to the old DataPlane that we just refactored away.

No I don't agree. Deployment does not have nodeset definition. It just have everything that a deployment run would need to deploy services.

> If you changed it so that all these fields were individual per NodeSet, then it is very similar to what we already have with NodeSet, except it's a list of NodeSets

Does not matter if a deployment has nodeset references, creating the deployment CR does not reconcile nodesets. So it's not the same .

slagle:
> > What "it" do you mean by "It can't"? The NodeSet controller is reconciling them today.
> 
> NodeSet controller does not own them neither watches them. If you patch the services it would reflect the modified CR in etcd. But when any nodeset CR is reconciled it would flip it back to the old definition in the dataplane-operator image. This is terrible design IMO.

I'm not convinced that a use case of editing our default OpenStackDataPlaneService definitions is needed outside of a dev/test context, for which there are various other easier methods. We did not support or encourage editing our rpm packaged tripleo-heat-templates services either. You had to copy and maintain that copy, and refer to it as a different service name or mapping. I don't recall that causing all that much consternation.

The current design works the same way. If an edit is needed, instead of overwriting what we ship as a default resource, you can customize it using the supported composable services interfaces, and use a different name.

rabi:
> I'm not convinced that a use case of editing our default OpenStackDataPlaneService definitions is needed outside of a dev/test context

How do we differentiate between standard services vs custom services in the nodeset spec? Customers will definitely write and use custom services and for me everything is a service (read playbook) with single interface. We can publish definition of some standard ones, which customers can use and change as per their requirement.  I also saw the note on the install_yamls PR that nova and other services won't be using defaults we have for them. 

The current way of using a different image for a so called `default` service is to patch the AEE default image (which will impact all default services) or build dataplane-operator with changed services https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/Dockerfile#L64 and then patch dataplane-operator CSV.

The fact that some services would be reconciled by nodeset controller and some not and then custom ones can be created/patched outside is a confusing/bad user experience and can lead to number of customer case. I've seen several questions on that behavior on slack till now. Anyway, I don't really understand the concerns  here exactly, but I can't argue more, if we want to block this. I rest my case.

slagle:
> > I'm not convinced that a use case of editing our default OpenStackDataPlaneService definitions is needed outside of a dev/test context
> 
> How do we differentiate between standard services vs custom services in the nodeset spec?

Other than by name, there is no differentiation in the NodeSet spec. The difference between a standard vs custom is:

- RH ships the standard services in the dataplane-operator image (I'm assuming this is what you mean by standard)
- The standard services are reconciled by the NodeSet controller
- Custom services are reconciled by the user (oc apply)

>Customers will definitely write and use custom services and for me everything is a service (read playbook) with single interface. We can publish definition of some standard ones, which customers can use and change as per their requirement.

I don't think the original intent is that customers would change the standard services directly on the same resource. They can however copy the standard service YAML, change the name, change the service spec in any way (including setting a different ansibleEE image), and then create it as a new service with the different name.

> I also saw the note on the install_yamls PR that nova and other services won't be using defaults we have for them.

Yes, some will be, some won't be depending on the customizations.

> 
> The current way of using a different image for a so called `default` service is to patch the AEE default runner image (which will impact all default services) or build dataplane-operator with changed services https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/Dockerfile#L64 and then patch openstack-operator CSV.

I think this is exclusive to dev/test context. There are also other ways, such as running dataplane-operator from git and editing under config/services. For prod, then the user creates a new service with a different name.

> The fact that some services would be reconciled by nodeset controller and some not and then custom ones can be created/patched outside is a confusing/bad user experience and can lead to number of customer case. I've seen several questions on that behavior on slack till now. Anyway, I don't really understand the concerns here exactly, but I can't argue more, if we want to block this. I rest my case.

My main objection is to the CRD changes. I don't have strong objections to changes to how the services are reconciled. But, there are some base lines I feel are needed, such as the set of standard services that can be created automatically (as we know all the values).

Service reconciliation and the user experience workflow around it could be separate from refactoring the CRDs. We can leave CRDs as-is, and either drop the service reconciliation entirely so it's all manual, or fine tune it so that it does not undo user edits, or something else. I can't say there will be broad agreement on that (I have not heard from others), but they could be separate proposals if you feel strongly about it.

rabi:
>  I can't say there will be broad agreement on that (I have not heard from others), but they could be separate proposals if you feel strongly about it.

I'm not sure there was a broad agreement on reconciling services in nodeset/role controller? I did point out my concerns in https://github.com/openstack-k8s-operators/dataplane-operator/pull/236 and the reasoning I  got was we've to fix the kuttl tests and similar stuff has been done elsewhere. The fact that the so called 'standard` services are not specific to nodeset (not owned by nodeset/nodeset controller) and live ever after till they're manually deleted, makes it a bad design.

> My main objection is to the CRD changes. 

For me, `NodeSet` is the definition of set of provisioned nodes that are ready for deployment or fully deployed. `Deployment` is the definition of what needs to be deployed and how. So all AEE related details should be in `Deployment`.If I understand correctly, you're saying  `Deployment` is just a trigger and nothing more than that and `NodeSet` contains all properties including what should be deployed in it and how, then the `ServicesOverride` in deployment spec does not really make sense. Any change in services to be deployed in a nodeset should be changed in the nodeset spec.

slagle:
> > My main objection is to the CRD changes.
> 
> For me, `NodeSet` is the definition of set of provisioned nodes that are ready for deployment or fully deployed. `Deployment` is the definition of what needs to be deployed and how. So all AEE related details should be in `Deployment`.If I understand correctly, you're saying `Deployment` is just a trigger and nothing more than that and `NodeSet` contains all properties including what should be deployed in it and how, then the `ServicesOverride` in deployment spec does not really make sense. Any change in services to be deployed in a nodeset should be changed in the nodeset spec.

Deployment is the trigger with optional overrides which are offered for convenience and specific use cases (upgrades, etc).

rabi:
> Deployment is the trigger with optional overrides which are offered for convenience and specific use cases (upgrades, etc).

That's where I think the design is broken. We're saying it's just a trigger, but then the deployment resource owns the aee resources and also can override what would be deployed (services). Looking at the nodeset spec no one can tell what has been deployed on it at any point as the services could be overridden with more (including new ones) could have been deployed with service override.

slagle:
I'm going to close this one, in the interest of closing stale PRs.

PR #468, opened 2023-10-13
Merge frr and ovn-bg-agent services into one

fernandoroyosanchez:
There is no need to split the services in install, configure and run.

This task is about unifying them so that we just have a frr service and an ovn-bgp-agent service.

Depends-on: https://github.com/openstack-k8s-operators/edpm-ansible/pull/441


luis5tb:
This now needs to also adapt https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml

PR #469, opened 2023-10-13
Add sample for BGP deployments

luis5tb:


bshephar (approved):
Nice one, glad you got this working in the end. Anything else we should consider documenting along with this sample file?

luis5tb:
> Nice one, glad you got this working in the end. Anything else we should consider documenting along with this sample file?

I'm working on the install_yaml support in here: https://github.com/openstack-k8s-operators/install_yamls/pull/584, perhaps I should add some steps in there?

PR #471, opened 2023-10-16
[DataplaneService/nova]Add migration ssh key

gibizer:
The nova role in edpm-ansible requires[1] an ssh key-pair that is then used as to authenticate the nova user when it copies data between compute nodes during migration.

A placeholder Secret name for this key-pair is now added to the nova DataplaneService definition. The deployer (human, ci_framework, or install_yamls) needs to make sure that a secret is created with an SSH key-pair in it.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/446


config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
```diff
@@ -6,4 +6,8 @@ spec:
   label: dataplane-deployment-nova
   secrets:
     - nova-cell1-compute-config
+    # NOTE: this Secret needs to be created before deploying the control plane
```
bshephar, line 9: Did you mean dataplane? ".. created before deploying the dataplane". 

Or is something in the Control Plane using this too?

gibizer:
recheck
baremetal job failure is relevant, updated the install_yamls patch to fix it

gibizer:
recheck
the previous recheck seems to be ignored

slagle (commented):
+2

fao89 (approved):

PR #473, opened 2023-10-17
Add hashes for deployment ConfigMaps and Secrets to NodeSet

slagle:
- Add hashes for deployment ConfigMaps and Secrets to NodeSet
- Update Secrets in kuttl test to match real values


controllers/openstackdataplanenodeset_controller.go
```diff
@@ -298,7 +299,9 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 			condition.DeploymentReadyInitMessage)
 	}
 
-	deploymentExists, isDeploymentReady, err := checkDeployment(helper, req)
+	configMapHashes := make(map[string]string)
+	secretHashes := make(map[string]string)
```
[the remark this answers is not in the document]
fao89: next update would be a new commit to this PR or a follow up PR?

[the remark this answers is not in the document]
slagle: same PR. should be addressed now.

docs/deploying.md
```diff
@@ -444,3 +444,36 @@ inventory for each resource.
...
+	<snip>
+  configMapHashes:
+    ovncontroller-config: n655h5...
+  secretHashes:
+    neutron-dhcp-agent-neutron-config: n9ch5...
+    neutron-ovn-metadata-agent-neutron-config: n588h...
+    neutron-sriov-agent-neutron-config: n648h...
+    nova-cell1-compute-config: n576h...
+    nova-metadata-neutron-config: n56fh...
+```
```
gibizer, line 479: I would add that:
These hashes can be compared against the hash of the CM / Secret to see if there is newer input data that has not been deployed to NodeSet. For example if the hash of `nova-cell1-compute-config` in the NodeSet Status is differnt from the hash of `nova-cell1-compute-config` in the NovaCell/cell1 status. Then there is nova-compute control plane configuration data the needs to be deployed to the EDPM compute nodes.

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -316,28 +328,117 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
 	if err != nil {
 		helper.GetLogger().Error(err, "Unable to retrieve OpenStackDataPlaneDeployment CRs %v")
 		return false, false, err
 	}
+
+	hashNodeSetServices := false
+	deploymentExists := false
+	deploymentReady := false
+
 	for _, deployment := range deployments.Items {
```
gibizer, line 354: If we have multiple Deployment CRs for the same NodeSet does the hash calculation below considers that newest Deployment?

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -316,28 +328,117 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
+			deployment.Spec.NodeSets, instance.Name) {
+
+			deploymentExists = true
+
 			if deployment.Status.Deployed {
-				return true, true, nil
+				deploymentReady = true
+				if len(deployment.Spec.ServicesOverride) > 0 {
+					for _, serviceName := range deployment.Spec.ServicesOverride {
+						err = getServiceHashes(ctx, helper, instance, serviceName, configMapHashes, secretHashes)
```
gibizer, line 364: Does it make sure that we record the hash of the CM/Secrets at the time when the Deployment was done? E.g. if the CM/Secret is updated after the Deployment became Ready then we should not put the new hash of the CM/Secret to the NodeSet.

tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml
gibizer, line 1: Can we add a test case that does to following scenario:
1) Do a deployment and assert that the hashes are stored (this step is already done by this PR)
2) Change one of the input CM / Secret and assert that the recorded hash in the NodeSet are not updated as the new input is not deployed yet.
3) Do another deployment and assert that now the new CM / Secret hash is stored in the NodeSet

controllers/openstackdataplanenodeset_controller.go
slagle: That is a good point. The time delta when such a change could occur would be from the CM/Secret is mounted into the AnsibleEE pod, and when the Deployment.Status.Deployed is set to true, which triggers the watch reconcile for the NodeSet. Given there could be ansible executions happening in between there, the delta could be large.

To mitigate that, I can update this PR to also save the hashes on the Deployment Status as well, as they are used. The NodeSet would then just read and copy the hash values from the Deployment. There is still a race condition there in that the CM/Secrets could change, but it's a much smaller window.

controllers/openstackdataplanenodeset_controller.go
slagle: All the hashes from all the Deployments are added. We're relying on the Watch to trigger this code. In theory the watch events ought to be ordered, as the Deployments are changed, so the latest would have been the most recent event. However, I assume there are numerous reasons why the Watch might be triggered, so probably best if i update this to sort the Deployments before looping over them.

controllers/openstackdataplanenodeset_controller.go
gibizer: I think we need to persist the hash of the CM/Secret when the Deployment first becomes Ready. If we do so then we have a snapshot of the hash from the time of the Deployment. Then the nodeset can copy the hash from the newest Deployment to show what is deployed on the nodes.  I think we are on the same page about this. 

As you noted there is still a race window but smaller. If we want to make this race free then we need to prevent CM/Secret changing while the deployment is running. One way to do that would be to use immutable CM/Secret objects https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable However that would need multiple changes:
* service operators would need to generate immutable CM/Secrets instead of updating them in place
* the user would need to update the DataplaneService definition to point to the new CM/Secret object after the service operator generated the new inputs.

How do you feel, do we want to go towards immutable CM/Secret usage?

controllers/openstackdataplanenodeset_controller.go
slagle: I added the sort.

controllers/openstackdataplanenodeset_controller.go
slagle: In the latest iteration, I've moved the calculation of the hashes to be done when the Deployment is completed and stored on the Deployment.Status. When the NodeSet Deployment watcher triggers a reconcile, they'll be copied to NodeSet.Status. I still need to add more tests and update the docs to account for this change. I'll mark this WIP until those are done.  However, in the meantime let me know if you think this approach is better.

I think we should defer the discussion around immutable ConfigMaps/Secrets to a separate effort, unless you think that doing that would make the hash calculation obsolete.

gibizer: > I think we should defer the discussion around immutable ConfigMaps/Secrets to a separate effort, unless you think that doing that would make the hash calculation obsolete.

Unfortunately that does not make the hash calculation obsolete as the DataPlaneService CRs are mutable and therefore can be changed to point to new CMs/Secrets. So the immutable CM/Secret only helps with the race window. So I guess we would not gain that much. Let's park this part of the discussion.

> In the latest iteration, I've moved the calculation of the hashes to be done when the Deployment is completed and stored on the Deployment.Status. When the NodeSet Deployment watcher triggers a reconcile, they'll be copied to NodeSet.Status. I still need to add more tests and update the docs to account for this change. I'll mark this WIP until those are done. However, in the meantime let me know if you think this approach is better.

This sounds good to me but I have to look at the code too.

controllers/openstackdataplanedeployment_controller.go
```diff
@@ -266,10 +272,56 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
...
+				instance.Status.ConfigMapHashes,
+				instance.Status.SecretHashes)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		for _, nodeSet := range nodeSets.Items {
+			for _, serviceName := range nodeSet.Spec.Services {
+				err = deployment.GetServiceHashes(
```
rabi, line 308: nit: GetServiceHashes() is little misleading here as it is actually updating the hashes in deployment status.

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -310,35 +319,54 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
 		helper.GetLogger().Error(err, "Unable to retrieve OpenStackDataPlaneDeployment CRs %v")
 		return false, false, err
 	}
 
 	isDeploymentReady := false
 	deploymentExists := false
+
+	// Sort deployments from oldest to newest
+	slices.SortFunc(deployments.Items, func(a, b dataplanev1.OpenStackDataPlaneDeployment) int {
+		if a.CreationTimestamp.String() < b.CreationTimestamp.String() {
```
rabi, line 344: Though we don't expect two deployments running at the same time, CreationTimestamp is when the CRD is created (not fully reconciled i.e services deployed).

rabi (approved):
Few minor comments, but LGTM.

controllers/openstackdataplanedeployment_controller.go
slagle: I renamed it to GetDeploymentHashesForService

controllers/openstackdataplanenodeset_controller.go
slagle: Good point. I've changed this to check LastTransitionTime on the DeploymentReadyCondition. Likely still not perfect, but I think it's slightly better.

tests/kuttl/tests/dataplane-deploy-no-nodes-test/00-dataplane-create.yaml
slagle: I've added these tests in the latest commit.

gibizer (approved):
looks good

slagle:
recheck

slagle:
recheck

baremetal job failing with:

`
TASK [osp.edpm.edpm_users : Ensure user is present on the host [ openvswitch ]] ***
Monday 22 January 2024  23:02:34 +0000 (0:00:01.967)       0:00:05.960 ******** 
[0;31mfatal: [edpm-compute-0]: FAILED! => {"changed": false, "msg": "usermod: user openvswitch is currently used by process 1192\n", "name": "openvswitch", "rc": 8}[0m
`

should be fixed by https://github.com/openstack-k8s-operators/edpm-ansible/pull/540/

slagle:
/retest

error: build error: Failed to push image: trying to reuse blob sha256:07a64a71e01156f8f99039bc246149925c6d1480d3957de78510bbec6ec68f7a at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": net/http: TLS handshake timeout

slagle:
recheck

`
"https://api.crc.testing:6443/api?timeout=32s\\\": dial tcp 192.168.122.10:6443: connect: connection refused\\nThe connection to the server api.crc.testing:6443 was refused -
`

slagle:
recheck

slagle:
Last rdo failure was:

`
{"msg": "'ctlplane_subnet_cidr' is undefined. 'ctlplane_subnet_cidr' is undefined"}
`

so perhaps there is a misconfiguration somewhere.

slagle:
recheck

looks like it just took too long

timed out at 15:33:06:

`
2024-01-23 15:33:06.146270 | controller | fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["oc", "wait", "--timeout=30m", "--for=condition=Ready", "openstackdataplanenodeset", "openstack-edpm-ipam", "-n", "openstack"], "delta": "0:30:00.143601", "end": "2024-01-23 15:33:06.113730", "msg": "non-zero return code", "rc": 1, "start": "2024-01-23 15:03:05.970129", "stderr": "error: timed out waiting for the condition on openstackdataplanenodesets/openstack-edpm-ipam", "stderr_lines": ["error: timed out waiting for the condition on openstackdataplanenodesets/openstack-edpm-ipam"], "stdout": "", "stdout_lines": []}
`

but by 20:33:50Z the NodeSet was ready:

`
    - lastTransitionTime: "2024-01-23T20:33:50Z"
      message: Deployment completed
      reason: Ready
      status: "True"
      type: Deployment ready for NodeSet
`

rebtoor:
recheck

putting node onhold for troubleshooting

rebtoor:
recheck

(putting node onhold for troubleshooting)

pablintino:
recheck

PR #474, opened 2023-10-17
Fix services list in HCI example

fultonj:
Any services include in the OpenStackDataPlaneNodeSet are automatically created by the data plane controller. Since we don't want to burden the user with manually creating these, restore the full list of services which are used. We still rely on the servicesOverride in each deployment however to identify which subset of services are deployed when.


fmount (approved):
/lgtm

PR #475, opened 2023-10-18
Don't use nodeName for dns records

rabi:
If the nodeName and hostName are different then host can't be resolved using hostName.

### DIFF
--- a/pkg/deployment/ipam.go
+++ b/pkg/deployment/ipam.go
@@ -18,6 +18,7 @@ package deployment
 
 import (
 	"context"
+	"regexp"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -72,7 +73,16 @@ func createOrPatchDNSData(ctx context.Context, helper *helper.Helper,
 	var ctlplaneSearchDomain string
 	// Build DNSData CR
 	for nodeName, node := range instance.Spec.Nodes {
+		var shortName string
 		nets := node.Networks
+		hostName := node.HostName
+
+		if isFQDN(hostName) {
+			shortName = strings.Split(hostName, ".")[0]
+		} else {
+			shortName = hostName
+		}
+
 		if len(nets) == 0 {
 			nets = instance.Spec.NodeTemplate.Networks
 		}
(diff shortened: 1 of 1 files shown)


pkg/deployment/ipam.go
```diff
@@ -81,11 +91,16 @@ func createOrPatchDNSData(ctx context.Context, helper *helper.Helper,
 			ipSet, ok := allIPSets[nodeName]
 			if ok {
 				for _, res := range ipSet.Status.Reservation {
+					var fqdnNames []string
 					dnsRecord := infranetworkv1.DNSHost{}
 					dnsRecord.IP = res.Address
-					var fqdnNames []string
-					fqdnName := strings.Join([]string{nodeName, res.DNSDomain}, ".")
-					fqdnNames = append(fqdnNames, fqdnName)
+					fqdnName := strings.Join([]string{shortName, res.DNSDomain}, ".")
```
bshephar, line 97: Just making sure my understanding is correct.
So, we strip out the `shortName` above. Then we turn it back into a FQDN here. 

This is just to verify that the FQDN we have been provided matches what we expect for the cluster yeah?

rabi: `hostName` could be `shortName` (edpm-compute-0) or `fqdn` (edpm-compute-0.example.com). We build the dns records in `dnsdata` by adding the `dnsDomain`[1] for every network to `shortName`. if the `hostName` is a `fqdn` we still add that to the dnsdata pointing to the ctlplane ip.

so if the `hostName` if `edpm-compute-0.example.com`. Following will be the dns records 

```
spec:
  hosts:
  - ip: 192.168.122.100
    hostnames:
    - edpm-compute-0.ctlplane.example.com
    - edpm-compute-0.example.com
  - ip: 172.17.0.100
    hostnames:
    - edpm-compute-0.internalapi.example.com
  - ip: 172.18.0.100
    hostnames:
    - edpm-compute-0.storage.example.com
  - ip: 172.19.0.100
    hostnames:
    - edpm-compute-0.tenant.example.com
```

Earlier we used the nodeName(key) as the `shortName`, but that causes issues as hosts can't be resolved if that key is not the same as hostName. It possible that nodeName key and hostName can be different and hostName can be fqdn. That has been addressed by this PR.
 
[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/config/samples/network_v1beta1_netconfig.yaml#L8

slagle (commented):
+2

pkg/deployment/ipam.go
bshephar: ack, understood. Thanks for the clarification. And yeah, the terminology can be confusing now. I was using cluster to refer to our edpm nodes and AEE pods configured DNS domains. :)

bshephar (approved):

PR #476, opened 2023-10-18
Added `bootstrap` service

rebtoor:
Since in the `configure-network` playbook there were a lot of roles unrelated to the network configuration, we moved those roles into another playbook called `bootstrap` that should be called as the first role (right after `download-cache`, if/when used) in the deployment process.

https://issues.redhat.com/browse/OSP-28971

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml
@@ -8,6 +8,7 @@ spec:
   # Create this deployment before Ceph is deployed
   # on EDPM nodes in an HCI scenario.
   servicesOverride:
+    - bootstrap
     - configure-network
     - validate-network
     - install-os
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -1862,6 +1862,7 @@ spec:
               services:
                 default:
                 - download-cache
+                - bootstrap
                 - configure-network
                 - validate-network
                 - install-os
--- a/api/v1beta1/openstackdataplanenodeset_types.go
+++ b/api/v1beta1/openstackdataplanenodeset_types.go
@@ -56,7 +56,7 @@ type OpenStackDataPlaneNodeSetSpec struct {
 	NetworkAttachments []string `json:"networkAttachments,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={download-cache,configure-network,validate-network,install-os,configure-os,run-os,ovn,neutron-metadata,libvirt,nova,telemetry}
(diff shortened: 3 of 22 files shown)


fultonj:
Could `bootstrap` be added here too?

https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml#L11

rebtoor:
> Could this PR be updated to add `bootstrap` here too?
> 
> https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml#L11
> 
> otherwise, this is fine by me

You were faster than me :)

rebtoor:
At usual, kuttl tests are failing because the image used for dataplane pods is:

```
quay.io/openstack-k8s-operators/openstack-ansibleee-runner@sha256:2b33d35abd0ed7137647b7a19e037dda84f37d09dadad54cf4348fd7f2200c10
```

which is the `latest` image, not the one generated by the edpm-ansible PR.

config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml
```diff
@@ -8,6 +8,7 @@ spec:
   # Create this deployment before Ceph is deployed
   # on EDPM nodes in an HCI scenario.
   servicesOverride:
+    - bootstrap
```
fultonj, line 11: thanks!

fultonj (commented):
+2

config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml
rebtoor: should i add `download-cache` here as well?

fultonj: No, thanks. That doesn't need to be covered in this PR.

fao89:
> At usual, kuttl tests are failing because the image used for dataplane pods is:
> 
> ```
> quay.io/openstack-k8s-operators/openstack-ansibleee-runner@sha256:2b33d35abd0ed7137647b7a19e037dda84f37d09dadad54cf4348fd7f2200c10
> ```
> 
> which is the `latest` image, not the one generated by the edpm-ansible PR.

here is what I did to force latest on adoption: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/dataplane_adoption/tasks/main.yaml#L1-L11
we might need to do the same for kuttl tests

rebtoor:
recheck

slagle (approved):

rebtoor:
recheck

fao89 (approved):

rebtoor:
recheck

rebtoor:
recheck

(added depends-on with ci platforms)

rebtoor:
recheck

PR #487, opened 2023-10-24
Block baremetal set template changes

bshephar:
This change ensures users can't change the `baremetalSetTemplate` items after the initial deployment. Since this would require the node to be re-provisioned and this is not something that would be desirable. Instead, the user should delete and re-create the nodeSet if they want to change `baremetalSetTemplate` items.

Related: https://github.com/openstack-k8s-operators/dataplane-operator/pull/486

### DIFF
--- a/api/v1beta1/openstackdataplanenodeset_webhook.go
+++ b/api/v1beta1/openstackdataplanenodeset_webhook.go
@@ -17,8 +17,15 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+	"reflect"
+
 	baremetalv1 "github.com/openstack-k8s-operators/openstack-baremetal-operator/api/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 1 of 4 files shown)


slagle (commented):
+2.

It would be good to document this as well. Although, I'm not sure we have much in the way of baremetal docs in this repo.

abays:
We do have checks in the `OpenStackBaremetalSet` controller that aim to prevent inadvertent re-provisioning, but this additional safety is prudent

bshephar:
> We do have checks in the `OpenStackBaremetalSet` controller that aim to prevent inadvertent re-provisioning, but this additional safety is prudent

Yeah, I did notice that. I thought it would be better if they got immediate feedback to let them know they can't do that. My intention is to allow for users to make changes to other aspects of the spec and have them rolled out to nodes. So before I embark on that journey, I thought protecting this specific parameter to be more explicit about its immutability with relevant errors would be nice.

bshephar:
> +2.
> 
> It would be good to document this as well. Although, I'm not sure we have much in the way of baremetal docs in this repo.

Yeah, the best place I thought of putting it would be in the description of the CRD for this field? But maybe we should have some more comprehensive baremetal docs for it.

api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -87,8 +93,30 @@ func (r *OpenStackDataPlaneNodeSet) ValidateCreate() error {
...
+		return apierrors.NewInternalError(
+			fmt.Errorf("Expected a OpenStackDataPlaneNodeSet object, but got %T", oldNodeSet))
+	}
+
+	var errors field.ErrorList
+	// Changes to the baremetalSetTemplate after the initial deployment would necessitate
+	// a redeploy of the node. Thus we should block these changes and require the user to
+	// delete and redeploy should they wish to make such changes after the initial deploy.
+	if !reflect.DeepEqual(oldNodeSet.Spec.BaremetalSetTemplate, r.Spec.BaremetalSetTemplate) {
+		errors = append(errors, field.Forbidden(
```
rabi, line 107: Well, this would not allow scale-out/scale-in as that's achieved updating `baremetalHosts` field.

rabi (dismissed):
There are already some checks in `openstack-baremetal-operator` webhook[1] which would not allow changing `hardwareReq` and `bmhLabelSelector` fields.  Rather than blocking all changes to the spec we probably need to filter out fields that can be changed.

Also, It would be better to use the same pattern we use in `openstack-operator`, where the validation/defaulting logic is `openstack-baremetal-operator` and called from `dataplane-operator` to avoid inconsistency. 

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/api/v1beta1/openstackbaremetalset_webhook.go#L99-L149

api/v1beta1/openstackdataplanenodeset_webhook.go
bshephar: This is a great point. So yeah, let's change it to match what is happening in the `openstack-baremetal-operator` so that we don't block valid operations like scale out/in.

bshephar: So, as of the latest update to this PR, we will just return the error we get from the `openstack-baremetal-operator` `ValidateUpdate()` webhook. Which from the functional tests I have here will be:

```
<string>: admission webhook "vopenstackdataplanenodeset.kb.io" denied the request: OpenStackDataPlaneNodeSet.dataplane.openstack.org "edpm-compute-nodeset" is invalid: spec.baremetalSetTemplate: Forbidden: cannot change "bmhLabelSelector" nor "hardwareReqs" when previous count of "baremetalHosts" > 0
```

bshephar:
> We do have checks in the `OpenStackBaremetalSet` controller that aim to prevent inadvertent re-provisioning, but this additional safety is prudent

The latest iteration of this just offloads the validation to the `OpenStackBaremetalSet` `ValidateUpdate()` function. So we avoid duplication there, and we get to warn end users directly rather than potentially passing the error back via the dataplane-operator logs.

bshephar:
recheck

bshephar:
recheck

slagle (approved):

api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -87,8 +94,54 @@ func (r *OpenStackDataPlaneNodeSet) ValidateCreate() error {
...
+		baremetalSetObject := &baremetalv1.OpenStackBaremetalSet{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      r.Name,
+				Namespace: r.Namespace,
+			},
+		}
+		r.Spec.BaremetalSetTemplate.DeepCopyInto(&baremetalSetObject.Spec)
+
+		// Call openstack-baremetal-operator ValidateUpdate() webhook to parse changes
+		err := baremetalSetObject.ValidateUpdate(oldBaremetalSetObject)
```
rabi, line 129: Does this work? We're creating a new OpenStackBaremetalSet object and calling ValidateUpdate() on it. So I think the ownerUIDLabelSelector won't match when selecting existing BMHs[1].

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/api/v1beta1/openstackbaremetalset_webhook.go#L136-L141

PR #490, opened 2023-10-26
Add mechanism to detect relevant spec changes

bshephar:
This change adds a Status field called ConfigChanged. We update this field in cases where the current config differs from the previously applied config. This is specifically used to detect changes between spec.nodes and spec.nodeTemplate. If either of those have a variance, then the ConfigChanged status field will be set to true.

### DIFF
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"time"
 
 	"golang.org/x/exp/slices"
@@ -94,6 +96,12 @@ type OpenStackDataPlaneNodeSetReconciler struct {
 	Kclient kubernetes.Interface
 	Scheme  *runtime.Scheme
 	Log     logr.Logger
+}
+
+// LastAppliedConfiguration holds the Spec section from the last-applied-configuration annotation
+// that is added by kubectl or oc.
+type LastAppliedConfiguration struct {
+	Spec *dataplanev1.OpenStackDataPlaneNodeSetSpec
 }
 
 //+kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplanenodesets,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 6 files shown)


controllers/openstackdataplanenodeset_controller.go
```diff
@@ -192,6 +198,21 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
 
+	// If we are reconciling an existing resource, compare last-applied-configuration with current spec to
+	// determine if any changes are requied. Set the status accordingly.
+	LastAppliedAnnotation := instance.GetAnnotations()["kubectl.kubernetes.io/last-applied-configuration"]
+	if LastAppliedAnnotation != "" {
+		var lastAppliedConfiguration LastAppliedConfiguration
+		if err := json.Unmarshal([]byte(LastAppliedAnnotation), &lastAppliedConfiguration); err != nil {
+			return ctrl.Result{}, err
+		}
+		if !reflect.DeepEqual(lastAppliedConfiguration.Spec.Nodes, instance.Spec.Nodes) || !reflect.DeepEqual(lastAppliedConfiguration.Spec.NodeTemplate, instance.Spec.NodeTemplate) {
```
bshephar, line 209: reflect is pretty expensive. It might be more efficient to just hash the values for each of these sections together and write that hash value instead of just a bool. Then just compare hashes each time we reconcile.

Happy to take feedback on this and implement whichever method we feel is more appropriate. The benefit of this way is that the user gets a clear indication that the configuration has been changed, whereas hash values might not be as intuitive.

bshephar:
recheck

bshephar:
CI Job Failed getting a ca-bundle secret.

controllers/openstackdataplanenodeset_controller.go
bshephar: Alternative proposal is here:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/491

bshephar:
Hash mechanism works better since we can use it to detect changes on the AEE Spec too.
https://github.com/openstack-k8s-operators/dataplane-operator/pull/491

PR #491, opened 2023-10-26
Generate config hash for nodeTemplate and nodes sections

bshephar:
This change generates a hash of the config contained within the nodeTemplate and nodes sections of the NodeSet spec. This can be used to determine whether or not it's necessary to re-run the Ansible tasks.

This is being proposed as a potential alternative to:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/490

### DIFF
--- a/api/v1beta1/openstackdataplanenodeset_types.go
+++ b/api/v1beta1/openstackdataplanenodeset_types.go
@@ -115,6 +115,17 @@ type OpenStackDataPlaneNodeSetStatus struct {
 
 	// SecretHashes
 	SecretHashes map[string]string `json:"secretHashes,omitempty" optional:"true"`
+
+	// ConfigHash - holds the curret hash of the NodeTemplate and Node sections of the struct.
+	// This hash is used to determine when new Ansible executions are required to roll
+	// out config changes.
+	ConfigHash string `json:"configHash,omitempty"`
+
+	// DeployedConfigHash - holds the hash of the NodeTemplate and Node sections of the struct
+	// that was last deployed.
+	// This hash is used to determine when new Ansible executions are required to roll
+	// out config changes.
+	DeployedConfigHash string `json:"deployedConfigHash,omitempty"`
 }
 
 //+kubebuilder:object:root=true
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -68,6 +68,13 @@ const (
 	OvnBgpAgentDefaultImage = "quay.io/podified-antelope-centos9/openstack-ovn-bgp-agent:current-podified"
 )
 
+// NodeConfigElements is a struct containing just the elements used for Ansible executions
+type NodeConfigElements struct {
+	BaremetalSetTempalte baremetalv1.OpenStackBaremetalSetSpec
+	NodeTemplate         dataplanev1.NodeTemplate
+	Nodes                map[string]dataplanev1.NodeSection
+}
+
 // SetupAnsibleImageDefaults -
 func SetupAnsibleImageDefaults() {
 	dataplaneAnsibleImageDefaults = dataplanev1.DataplaneAnsibleImageDefaults{
@@ -193,6 +200,16 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 
 	instance.Status.Conditions.MarkFalse(dataplanev1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
 
+	// Detect config changes and set Status ConfigHash
+	configHash, err := r.GetSpecConfigHash(instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if configHash != instance.Status.DeployedConfigHash {
+		instance.Status.ConfigHash = configHash
+	}
+
 	// Ensure Services
 	err = deployment.EnsureServices(ctx, helper, instance)
 	if err != nil {
@@ -304,6 +321,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		logger.Info("Set NodeSet DeploymentReadyCondition true")
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition,
 			condition.DeploymentReadyMessage)
+		instance.Status.DeployedConfigHash = configHash
 	} else if deploymentExists {
 		logger.Info("Set NodeSet DeploymentReadyCondition false")
 		instance.Status.Conditions.MarkFalse(condition.DeploymentReadyCondition,
@@ -321,7 +339,6 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 func checkDeployment(helper *helper.Helper,
 	instance *dataplanev1.OpenStackDataPlaneNodeSet,
 ) (bool, bool, error) {
-
 	// Get all completed deployments
 	deployments := &dataplanev1.OpenStackDataPlaneDeploymentList{}
 	opts := []client.ListOption{
@@ -450,4 +467,20 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(mgr ctrl.Manager)
 		Watches(&source.Kind{Type: &dataplanev1.OpenStackDataPlaneDeployment{}},
 			deploymentWatcher).
 		Complete(r)
+}
+
+// GetSpecConfigHash initialises a new struct with only the field we want to check for variances in.
+// We then hash the contents of the new struct using md5 and return the hashed string.
+func (r *OpenStackDataPlaneNodeSetReconciler) GetSpecConfigHash(instance *dataplanev1.OpenStackDataPlaneNodeSet) (string, error) {
+	nodeConfig := &NodeConfigElements{
+		BaremetalSetTempalte: instance.Spec.BaremetalSetTemplate,
+		NodeTemplate:         instance.Spec.NodeTemplate,
+		Nodes:                instance.Spec.Nodes,
+	}
+	configHash, err := util.ObjectHash(&nodeConfig)
+	if err != nil {
+		return "", err
+	}
+
+	return configHash, nil
 }
--- a/docs/assemblies/deploying.adoc
+++ b/docs/assemblies/deploying.adoc
@@ -245,7 +245,21 @@ NAME             STATUS   MESSAGE
 openstack-edpm   False    Deployment not started
 ----
 
-=== Understanding OpenStackDataPlaneServices
+=== NodeSet Config Changes
+We create a Hash of the inputs located in the OpenStackDataPlaneNodeSet Spec Nodes and NodeTemplate sections.
+This hash is then stored in the `status.configHash` field. If the current value of the configHash is different
+to the deployedConfigHash, then it is necessary to recreate the `OpenStackDataPlaneDeployment` to roll out
+the new changes:
+
+```
+$ oc get osdpns -o yaml | yq '.items[0].status.configHash'
+"n648hd6h88hc7h86hc7h568h585h79h5"
+
+```
+This field can be used to inform user decisions around when a new deploy is needed to reconclie the changes to the NodeSet.
+
+
+### Understanding OpenStackDataPlaneServices
 
 A dataplane is configured with a set of services that define the Ansible plays
 or playbooks that are executed to complete the deployment. The
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -1928,12 +1928,16 @@ spec:
                   - type
                   type: object
                 type: array
+              configHash:
+                type: string
               configMapHashes:
                 additionalProperties:
                   type: string
                 type: object
               deployed:
                 type: boolean
+              deployedConfigHash:
+                type: string
               deploymentStatuses:
                 additionalProperties:
                   items:
--- a/config/crd/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/config/crd/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -1928,12 +1928,16 @@ spec:
                   - type
                   type: object
                 type: array
+              configHash:
+                type: string
               configMapHashes:
                 additionalProperties:
                   type: string
                 type: object
               deployed:
                 type: boolean
+              deployedConfigHash:
+                type: string
               deploymentStatuses:
                 additionalProperties:
                   items:
--- a/docs/assemblies/custom_resources.adoc
+++ b/docs/assemblies/custom_resources.adoc
@@ -576,6 +576,16 @@ OpenStackDataPlaneNodeSetStatus defines the observed state of OpenStackDataPlane
 | SecretHashes
 | map[string]string
 | false
+
+| configHash
+| ConfigHash - holds the curret hash of the NodeTemplate and Node sections of the struct. This hash is used to determine when new Ansible executions are required to roll out config changes.
+| string
+| false
+
+| deployedConfigHash
+| DeployedConfigHash - holds the hash of the NodeTemplate and Node sections of the struct that was last deployed. This hash is used to determine when new Ansible executions are required to roll out config changes.
+| string
+| false
 |===
 
 <<custom-resources,Back to Custom Resources>>
--- a/tests/functional/openstackdataplanenodeset_controller_test.go
+++ b/tests/functional/openstackdataplanenodeset_controller_test.go
@@ -16,6 +16,7 @@ limitations under the License.
 package functional
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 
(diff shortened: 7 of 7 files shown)


bshephar:
recheck

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -389,3 +411,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(mgr ctrl.Manager)
...
+		return "", err
+	}
+
+	// Hash the values and convert to a slice of bytes required by hex.EncodeToString()
+	var hashSlice []byte
+	hash := md5.Sum(jsonBytes)
+	hashSlice = hash[:]
+
+	return hex.EncodeToString(hashSlice), nil
+}
```
slagle, line 434: Can we use ObjectHash() from lib-common instead?

slagle (commented):
I think we should use the hashing code from lib-common instead, otherwise LGTM.

controllers/openstackdataplanenodeset_controller.go
bshephar: I had no idea this function existed. That is much cleaner, I'll push a new revision.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -104,6 +104,16 @@ type OpenStackDataPlaneNodeSetStatus struct {
 
 	// AllIPs
 	AllIPs map[string]map[infranetworkv1.NetNameStr]string `json:"AllIPs,omitempty" optional:"true"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	// ConfigChanged - Informs us if there are any differences between the deployed spec vs what the most recent version is.
+	// We can use this to inform our decisions about when to re-execute Ansible tasks.
+	ConfigChanged bool `json:"configChanged,omitempty" optional:"true"`
```
rabi, line 111: Why do we need this flag in Status as this is only relevant to a "particular recnocile"? We can find out if it has changed just by comparing what's in the status already and what's generated,

bshephar: It's just a clear indicator to the user that there is a variance between the current spec sections for nodes and nodeTemplate. So the use case would be that a user deploys the nodeSet, later they change some random ansible variable. This would require the user to create a new `OpenStackDataPlaneDeployment` for those changes to actually be applied to the nodes. Because this isn't really intuitive for users familiar with Kubernetes, the intention behind this boolean flag is to provide users with a clear indication that they need to take additional actions for the changes to be applied.

rabi: >  later they change some random ansible variable.

Any changes to ansible variables and for that matter anything in NodeSet spec would require new deploy with a OpenStackDataPlaneDeployment resource. I don't think we should have a flag for that. Also, that flag would be changed from true/false when for any reason NodeSet is reconciled (again) other than changes to those sections.

bshephar: The hash I'm creating is only created from `/spec/nodes` and `/spec/nodeTemplate`. So changes to anything else aren't included in this hash. This should only be set to true if the user changes something specific to the configuration of the dataplane nodes themselves. It shouldn't be changed for each and every reconcile, because the contents of `/spec/nodes` and `/spec/nodeTemplate` won't be changed each reconcile. 

If they try to change the NTP server for example. This would then change to `configChanged: true` until the user creates a new `OpenStackDataplaneDeployment` to actually roll out the changes. Once those changes have been applied, then the `configHash` will be written to the status and this `configChanged` bool will be set back to `false`.

rabi: >  Once those changes have been applied, then the configHash will be written to the status

Oh! you're changing the `configHash` in `Status` only if `isDeploymentReady: true` (I missed that and forgot that we're now checking all existing deployments, which is not in your PR, it needs rebase). Then it's the deployed config hash (not the current one based on the spec changes), It would be be better to rename it as `DeployedConfigHash`. Also, why're we generating the configHash in every reconcile? Those lines from L196 should be moved there too.

rabi: If we keep the flag to show that config has changed, you probably have to leave code where you calculate the `ConfigHash`, but this has to clear to users that though the flag is true and ConfigHash in status is the old one without deployment. So yeah probably renaming it to `DeployedConfigHash` would make more sense.

rabi: Also I missed `isDeploymentReady` would be true without a new deployment (old deployments already deployed). So this probably won't work at all.

bshephar: I see what you mean, yeah. 

Maybe a better option would be to copy the configHash onto the deployment. Then if the associated deployment and the nodeSet have a variance in the configHash we could provide some indication for the user that the current config on the NodeSet has not been applied.

slagle: Why not just ConfigHash and DeployedConfigHash?
ConfigHash is the hash of the current spec.
DeployedConfigHash is the hash at the last time a deployment was done with the NodeSet services.

If they're equal, that's the signal that no new deployment is needed, and a deployment is needed if they are not equal.

controllers/openstackdataplanenodeset_controller.go
slagle: Can we leave this unresolved until it's updated to use lib-common?

slagle (commented):
This will be needed by https://issues.redhat.com/browse/OSPRH-3890, and I've added a sub-task to track it https://issues.redhat.com/browse/OSPRH-4306

controllers/openstackdataplanenodeset_controller.go
bshephar: Yeah, I marked this resolved when I changed it to lib-common last time, but after I realised the configHash wasn't actually changing, I reverted it. In any case, I have changed it again to use lib-common now.

bshephar:
Hmm functional tests pass fine locally. Retesting it to see if that's reproducible.

bshephar:
recheck

controllers/openstackdataplanenodeset_controller.go
slagle: thanks. it looks like it's working now based on the passing envtest

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -193,6 +199,20 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
 
+	// Detect config changes and set Status ConfigHash
+	configHash, err := r.GetSpecConfigHash(instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if configHash != instance.Status.DeployedConfigHash {
+		instance.Status.ConfigHash = configHash
+		err := helper.PatchInstance(ctx, instance)
```
slagle, line 210: isn't this a redundant call to PatchInstance given that happens in `defer func`? If you drop this call does instance.Status.ConfigHash still get updated?

bshephar:
Seems unrelated:

~~~
    - lastTransitionTime: "2024-01-31T06:30:39Z"
      message: libvirt Deployment not yet ready
      reason: Requested
      severity: Info
      status: "False"
      type: libvirt Deployment ready
~~~

bshephar:
recheck

controllers/openstackdataplanenodeset_controller.go
bshephar: Hmm, yeah, I think you're right. Because even if the reconcile fails for any other reason, it will run the defer func(). So we should still end up with the same result. Let's drop it and see how it goes.

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -304,6 +320,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		logger.Info("Set NodeSet DeploymentReadyCondition true")
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition,
 			condition.DeploymentReadyMessage)
+		instance.Status.DeployedConfigHash = configHash
```
ciecierski, line 323: I have deployed nodeset and ran this command to patch nodeset:
oc patch osdpns/openstack-edpm -p '{"spec": {"nodeTemplate": {"ansible": {"ansibleVars": { "dns_search_domains": ["bne-home.com"] }}}}}' --type=merge

Is it desired that DeployedConfigHash is equal to configHash right after patch is applied. There was no openstackdeployemnt started or anything.

docs/deploying.md
```diff
@@ -238,6 +238,16 @@ NAME             STATUS   MESSAGE
 openstack-edpm   False    Deployment not started
 ```
 
+### NodeSet Config Changes
+We create a Hash of the inputs located in the OpenStackDataPlaneNodeSet Spec Nodes and NodeTemplate sections. This hash is then stored in the `status.configHash` field. If the current value of the configHash is different to the deployed configHash, then we set `status.configChanged`:
+
+```
+❯ oc get osdpns -o yaml | yq '.items[].status.configChanged'
```
ciecierski, line 245: It seems to be leftover from attempt in https://github.com/openstack-k8s-operators/dataplane-operator/pull/490

controllers/openstackdataplanenodeset_controller.go
bshephar: So these two values should only be different if you already have a `OpenStackDataPlaneDeployment` that has completed, and then the config is changed.

If there is no deployment at all though, I guess we shouldn't have any value in the DeployedConfigHash at all. Not until a deployment is created and finished. 

I'll double check the code there.

docs/deploying.md
bshephar: Ahh true. Yes, I need to update the docs.

controllers/openstackdataplanenodeset_controller.go
bshephar: Looks like this part I'm describing works fine. I think the issue might be with the length of the hash that's being generated. I'll see if I can shorten that and hope that solves the issues I'm seeing with functional tests.

docs/openstack_dataplanenodeset.md
```diff
@@ -0,0 +1,154 @@
+
```
rabi, line 1: You've to drop this file.

docs/assemblies/deploying.adoc
```diff
@@ -245,7 +245,21 @@ NAME             STATUS   MESSAGE
 openstack-edpm   False    Deployment not started
 ----
 
-=== Understanding OpenStackDataPlaneServices
+### NodeSet Config Changes
```
rabi, line 248: We're using ascii doc now, so this format won't work I think.

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -193,6 +199,16 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 
 	instance.Status.Conditions.MarkFalse(dataplanev1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
 
+	// Detect config changes and set Status ConfigHash
+	configHash, err := r.GetSpecConfigHash(instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if configHash != instance.Status.DeployedConfigHash {
```
rabi, line 208: This is kind of redundant, though there could be reconciliation when services are updated in a `OpenStackDataPlaneNodeSet`. But I don't know why someone would do that when they can override it in `OpenStackDataPlaneDeployment`. That's the reason I had proposed to move services to `OpenStackDataPlaneDeployment`.

Also, the hash should include `baremetalSetTemplate` if exists.

rabi: Also, I noticed we're adding (extra) ansible vars to `OpenStackDataPlaneDeployment` with https://issues.redhat.com/browse/OSPRH-1141, so how are these hashes relevant in situations where users use it?

bshephar: `baremetalSetTemplate` will be reconciled regardless of the `OpenStackDataPlaneDeployment` though won't it? Like, we don't need a `OpenStackDataPlaneDeployment` to exist for it to provision the nodes.

The issue we ran into that led to this PR was that users might change some Ansible variable in the NodeSet and expect that it is reconciled. But in reality, they actually need to create a new `OpenStackDataPlaneDeployment` for those changes to be applied. So this `configHash` and `deployedConfigHash` are designed to give the user some visual indication that there is config drift, and the new config has not been applied.

I think @slagle has some use cases around updates and upgrades as well that he is considering.

rabi: > baremetalSetTemplate will be reconciled regardless of the OpenStackDataPlaneDeployment though won't it?

If we're looking for `OpenStackDataPlaneNodeSet` changes to run a new `OpenStackDataPlaneDeployment`, I think it could be anything including services in `OpenStackDataPlaneNodeSet` and update-able fields fields in  `baremetalSetTemplate`, no?

slagle: The point of this PR was to record the hashes as a visual indication if something had changed since the last Deployment, not to automatically start a new Deployment. I'd leave that as a future enhancement where we can discuss how that might work.

As for services and vars on the Deployment itself, I would only record Status.DeployConfigHash if the completed Deployment does not have services or vars set.

slagle: I'd probably add that logic to `checkDeployment`. `isDeploymentReady` and `deploymentExists` should probably only return True if the Deployment was for the services and vars on the NodeSet.

slagle (commented):
+2. If @rabi is happy, let's merge it.
I'm good to fix what I pointed out about checkDeployment in a followup.

rabi:
> `isDeploymentReady` and `deploymentExists` should probably only return True if the Deployment was for the services and vars on the NodeSet.

I don't think we should change that and it won't be a good design choice. A deployment is relevant for a NodeSet, if it's in the `nodeSets` section of that deployment and deployment should update NodeSets deployment status. 

Also, we would be merging the ansible vars in the NodeSet with the ones in deployment and therefore assuming that we can get to know `something has changed since last deployment` and we have to deploy again,  as the last deployment (with ansible vars in deployment) could have used the updated ansible vars from the NodeSet.

Anyway, this PR needs work to address my other comments for docs cleanup and proper rebase before we can merge it.

docs/assemblies/deploying.adoc
```diff
@@ -245,7 +245,21 @@ NAME             STATUS   MESSAGE
...
+
+```
+$ oc get osdpns -o yaml | yq '.items[0].status.configHash'
+"n648hd6h88hc7h86hc7h568h585h79h5"
+
+```
+This field can be used to inform user decisions around when a new deploy is needed to reconclie the changes to the NodeSet.
+
+
+### Understanding OpenStackDataPlaneServices
```
rabi, line 262: This is  different from others, though the docs seem ok https://logserver.rdoproject.org/91/491/d7fc2d89f1b518d2c05b67f83da4712e83fcd38f/github-check/dataplane-operator-docs-preview/113ede4/docs_build/docs_build/dataplane/index-upstream.html

rabi (approved):
/lgtm

rabi:
Merging this to not block other work...We'll discuss the concerns separately.

bshephar:
> > `isDeploymentReady` and `deploymentExists` should probably only return True if the Deployment was for the services and vars on the NodeSet.
> 
> I don't think we should change that and it won't be a good design choice. A deployment is relevant for a NodeSet, if it's in the `nodeSets` section of that deployment and deployment should update NodeSets deployment status.
> 
> Also, we would be merging the ansible vars in the NodeSet with the ones in deployment and therefore assuming that we can get to know `something has changed since last deployment` and we have to deploy again, as the last deployment (with ansible vars in deployment) could have used the updated ansible vars from the NodeSet.
> 
> Anyway, this PR needs work to address my other comments for docs cleanup and proper rebase before we can merge it.

Yeah, let's have a think about the problem a bit more and see how we can improve on this.

I'm not sure we need to hash the Deployment overrides though, since they will be rolled out via the deployment when it's created. If a user wanted to change them, they would need to create a new deployment anyway? The main objective was just for the NodeSet to inform users that something is different about the NodeSet compared to what was used last time the deployment was run. If they're being overridden on the Deployment instead, I think that's a different problem to what I set out to solve here. 

My initial intent with this change was to automatically roll out changes that are made to the NodeSet provided a valid Deployment for that NodeSet still exists. But there was some push back on that design, so instead we settled on just providing visual information to the user that the NodeSet has been changed.

You raised a good point about the services on Slack. If a user adds a service, we wouldn't currently capture that. I think one small change we can make immediately is to just hash the entire `spec`. Initially, I had avoided this because of the changes the controller makes to `baremetalSetTemplate` during reconcile:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/baremetal.go#L37

But we're hashing after that is done anyway, so it doesn't really matter. Hashing the full spec would simplify this implementation and capture changes to those services.

PR #499, opened 2023-10-30
Simply cutting down the label to size

jpodivin:


jpodivin:
/retest

` error: must build at directory: not a valid directory: evalsymlink failure on '/alabama/install_yamls/out/openstack/infra/cr' : lstat /alabama/install_yamls/out/openstack/infra: no such file or directory `

api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
```diff
@@ -35,6 +35,8 @@ spec:
                   type: string
                 type: array
               label:
+                maxLength: 53
```
fultonj, line 38: +2 This looks like the right idea.

Curious about 53. ISTR the limit is 64 chars (0-63). Did you arrive at 53 so that we have 10 extra chars for other things?

jpodivin: Pretty much. The original limit was even more strict, about 24 chars, then k8s switched to RFC1035[0], bumping the limit for service names. However the transition was a bit uneven, so it didn't happen all at once. Finally they have implemented the new hard limit and runtime validation for it [1]. Also, some other names follow RFC1123. But that is very similar, only slightly less lenient on the chars.

Unfortunately for us, we are not creating the label directly, but by getting values from CRD. And to make it a bit more complicated we are also using it as a basis for pod name, which is getting suffix. 

But that isn't the only reason I've decided to keep some buffer. We are still in relatively early development. If we decide that there is some other transformation we can do with labels, we'll have a space to work in. And if we never do, and it turns out we need all of the space, we can increase the limit without having to change anything else.

Otoh, if we find out we are out of space and need to cut down again ... well it's going to be more painful. 

[1] https://www.rfc-editor.org/rfc/rfc1035
[2] https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L236

fao89 (approved):

PR #500, opened 2023-10-31
Re-run AnsibleEE jobs when config changes.

bshephar:
This change leverages the configHash generated by:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/491

We then write the `configHash` to the `AnsibleEE` `EnvVars` and use this to determine which jobs have been executed using the latest config. If the config changes, then we detect this and update the `NodeSet.Status.ConfigChanged` bool to provide visual confirmation to the user that the change has been registered. We reference this field to trigger the workflow that will:

1. Check each service for the current Deployed Config Hash;
2. Delete `AnsibleEE` CR's that have the old ConfigHash;
3. Create new `AnsibleEE` objects with the new ConfigHash and subsequently roll out the new config to the dataplane nodes.

Currently, we are only checking the `NodeTemplate` and `Nodes` parts of the Spec for changes and ignoring all other fields including ENV. So this would do nothing if the ENV was updated on the nodeSet for example.

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -1910,6 +1910,10 @@ spec:
                   - type
                   type: object
                 type: array
+              configChanged:
+                type: boolean
+              configHash:
+                type: string
               deployed:
                 type: boolean
             type: object
--- a/api/v1beta1/openstackdataplanenodeset_types.go
+++ b/api/v1beta1/openstackdataplanenodeset_types.go
@@ -92,6 +92,16 @@ type OpenStackDataPlaneNodeSetStatus struct {
 
 	// CtlplaneSearchDomain
 	CtlplaneSearchDomain string `json:"CtlplaneSearchDomain,omitempty" optional:"true"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+ 	// ConfigChanged - Informs us if there are any differences between the deployed spec vs what the most recent version is.
+ 	// We can use this to inform our decisions about when to re-execute Ansible tasks.
+ 	ConfigChanged bool `json:"configChanged,omitempty" optional:"true"`
+
+	// ConfigHash - holds the hash of the NodeTemplate and Node sections of the struct.
+	// This hash is used to determine when new Ansible executions are required to roll
+	// out config changes.
+	ConfigHash string `json:"configHash,omitempty"`
 }
 
 //+kubebuilder:object:root=true
--- a/api/v1beta1/openstackdataplanenodeset_types.go
+++ b/api/v1beta1/openstackdataplanenodeset_types.go
@@ -146,10 +146,20 @@ func (instance *OpenStackDataPlaneNodeSet) InitConditions() {
 
 // GetAnsibleEESpec - get the fields that will be passed to AEE
 func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
+	// Build env to include DEPLOY_CONFIG_HASH at the front of the slice
+	var aeeEnv []corev1.EnvVar
+	aeeEnv = append(aeeEnv, corev1.EnvVar{
+		Name: "DEPLOY_CONFIG_HASH",
+		Value: instance.Status.ConfigHash,
+	})
+	for _, envItem := range instance.Spec.Env {
+		aeeEnv = append(aeeEnv, envItem)
+	}
+
 	return AnsibleEESpec{
 		NetworkAttachments: instance.Spec.NetworkAttachments,
 		ExtraMounts:        instance.Spec.NodeTemplate.ExtraMounts,
-		Env:                instance.Spec.Env,
+		Env:                aeeEnv,
 	}
 }
 
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -18,6 +18,9 @@ package controllers
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -66,6 +69,12 @@ const (
 	// OvnBgpAgentDefaultImage -
 	OvnBgpAgentDefaultImage = "quay.io/podified-antelope-centos9/openstack-ovn-bgp-agent:current-podified"
 )
+
+// NodeConfigElements is a struct containing just the elements used for Ansible executions
+type NodeConfigElements struct {
+	NodeTemplate dataplanev1.NodeTemplate
+	Nodes        map[string]dataplanev1.NodeSection
+}
 
 // SetupAnsibleImageDefaults -
 func SetupAnsibleImageDefaults() {
@@ -192,6 +201,18 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		instance.Status.Conditions.MarkFalse(dataplanev1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
 	}
 
+	// Detect config changes and set Status ConfigHash
+	configHash, err := r.GetSpecConfigHash(instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if configHash != instance.Status.ConfigHash {
+		instance.Status.ConfigChanged = true
+	} else {
+		instance.Status.ConfigChanged = false
+	}
+
 	// Ensure Services
 	err = deployment.EnsureServices(ctx, helper, instance)
 	if err != nil {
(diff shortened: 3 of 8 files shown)


bshephar:
Example of something that will re-trigger a new deployment:
```
❯ oc get osdpns/edpm-compute -ojson | jq .spec.nodeTemplate.ansible.ansibleVars.dns_search_domains
[
  "bne-home.net",
  "okd.bne-shift.net"
]
```

If I changed this:
```
❯ oc patch osdpns/edpm-compute -p '{"spec": {"nodeTemplate": {"ansible": {"ansibleVars": { "dns_search_domains": ["bne-home.net"] }}}}}' --type=merge
openstackdataplanenodeset.dataplane.openstack.org/edpm-compute patched
```

```
❯ oc get osdpns/edpm-compute -ojson | jq .spec.nodeTemplate.ansible.ansibleVars.dns_search_domains
[
  "bne-home.net"
]
```

We can see it's recreated the AnsibleEE jobs:
```
❯ oc get openstackansibleee
NAME                                                  NETWORKATTACHMENTS   STATUS   MESSAGE
dataplane-deployment-configure-network-edpm-compute                        True     AnsibleExecutionJob complete
dataplane-deployment-configure-os-edpm-compute                             False    AnsibleExecutionJob is running
dataplane-deployment-download-cache-edpm-compute                           False    AnsibleExecutionJob is running
dataplane-deployment-install-os-edpm-compute                               True     AnsibleExecutionJob complete
dataplane-deployment-validate-network-edpm-compute                         True     AnsibleExecutionJob complete
```

```
❯ oc get po -l app=openstackansibleee
NAME                                                        READY   STATUS      RESTARTS   AGE
dataplane-deployment-configure-network-edpm-compute-7nqgc   0/1     Completed   0          21m
dataplane-deployment-configure-os-edpm-compute-zttx8        1/1     Running     0          49s
dataplane-deployment-download-cache-edpm-compute-cb6fc      1/1     Running     0          48s
dataplane-deployment-install-os-edpm-compute-x6xtt          0/1     Completed   0          20m
dataplane-deployment-validate-network-edpm-compute-rv7mj    0/1     Completed   0          20m
```

pkg/deployment/deployment.go
```diff
@@ -141,9 +141,27 @@ func ConditionalDeploy(
 ) error {
 
 	var err error
+	var deployedConfigHash string
 	log := helper.GetLogger()
```
slagle, line 146: ConditionalDeploy is called by Deploy. Deploy is only called from the Deployment reconciler. So what triggers this to start executing again?

I can't follow how updating the NodeSet, which updates the new Status fields on the NodeSet leads to the Deployment reconciling again.

slagle (commented):
I'm probably overlooking it, but I don't follow how the OpenStackDataPlaneDeployment starts reconciling again after updating the NodeSet.

We also need to take care about which Deployments are re-triggered by a NodeSet change. There could be specialized Deployments to temporary set FFU configs, or apply a hotfix, or a one-off service. 
Should all of those be re-triggered if the NodeSet config changes?

Or should this be restricted to just the set of services listed on the NodeSet?

Also keep in mind the set of services on the NodeSet could change over time. Do we really want to delete all old AnsibleEEs?

pkg/util/ansible_execution.go
```diff
@@ -55,6 +55,21 @@ func AnsibleExecution(
 	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
+
+	// If the AnsibleEE already exists and the config has changed, then we will delete
+	// the existing AnsibleEE and re-create it below to roll out the new config.
+	if ansibleEE != nil && ansibleEE.Spec.Env[0].Value != aeeSpec.Env[0].Value {
+		helper.GetLogger().Info("Redeploying OpenStackAnsibleEE: %s", ansibleEE.Name)
+		err := helper.GetClient().Delete(ctx, ansibleEE)
```
slagle, line 63: This also would delete the Job, which deletes the ansible execution logs. How do we account for saving those logs somewhere for debuggability?

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -289,6 +305,15 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 	// all setup tasks complete, mark SetupReadyCondition True
 	instance.Status.Conditions.MarkTrue(dataplanev1.SetupReadyCondition, condition.ReadyMessage)
 
+	// If the instance.Status.ConfigChanged field is true, the we can mark the deployment condition
+	// false. This will signal that we need to re-execute the deployment tasks.
```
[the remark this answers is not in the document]
bshephar: It all happens here:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/500/files/0f7118ab671530f83e55a078e34ffdbc278d21fd#diff-b44566f9caa976421532eb049c77b7eb247418f138516fab277a1ec33d3a7992R151-R159

The Deployment reconciler loops through each `nodeSet`, and passes the `nodeSet` object to `deployment.Deploy()` function. So we have access to the `Status` fields for each individual `nodeSet` as we progress into `Deploy()` and `ConditionalDeploy()`.

In `ConditionalDeploy()`, we check for this `ConfigChanged` status field. If it's `true`, then we compare the deployed `AnsibleEE` CR for the service with the latest `ConfigHash`.  If they don't match, then we re-deploy it.

pkg/util/ansible_execution.go
bshephar: It does indeed delete the job. I guess the alternative would be to create a new instance of the AEE object with part of the configHash in the name. Although, that will break the rest of the logic above for checking and comparing the config hash in the existing AEE, but doable I think.

Could probably remove that logic above and just look for a AEE with part of the confighash in the name, if I get a `k8serrors.IsNotFound()` we know we need to create a new one.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -136,10 +146,20 @@ func (instance *OpenStackDataPlaneNodeSet) InitConditions() {
 
 // GetAnsibleEESpec - get the fields that will be passed to AEE
 func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
+	// Build env to include DEPLOY_CONFIG_HASH at the front of the slice
+	var aeeEnv []corev1.EnvVar
+	aeeEnv = append(aeeEnv, corev1.EnvVar{
+		Name: "DEPLOY_CONFIG_HASH",
+		Value: instance.Status.ConfigHash,
```
bshephar, line 153: It would probably be better to use a label rather than inserting into the `EnvVars` like this. Happy to discuss further.

pkg/deployment/deployment.go
bshephar: Ah yeah, I think the issue is that my env never set `instance.Status.Deployed = true` because it eventually got stuck on:
```
{"level":"error","ts":"2023-10-31T12:48:55+10:00","msg":"Reconciler error","controller":"openstackdataplanedeployment","controllerGroup":"dataplane.openstack.org","controllerKind":"OpenStackDataPlaneDeployment","OpenStackDataPlaneDeployment":{"name":"edpm-compute","namespace":"openstack-operators"},"namespace":"openstack-operators","name":"edpm-compute","reconcileID":"200e708b-b841-4eb1-9e0b-503bbbfce125","error":"ConfigMap \"ovncontroller-config\" not found","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/Users/brendanshephard/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/brendanshephard/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/Users/brendanshephard/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:235"}
```

So you're right, if it had finished then my updates to the `NodeSet` would never be registered because we would exit the reconcile here:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanedeployment_controller.go#L131-L134

So the only way this would work is if I were setting `instance.Status.Deployed` to `false` when the nodeSet changes. Which it currently doesn't do and requires more thought.

controllers/openstackdataplanenodeset_controller.go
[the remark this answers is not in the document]
bshephar: Realising all of this relies on the issue you raised in the other comment:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/500#discussion_r1377472531

api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
```diff
@@ -1910,6 +1910,10 @@ spec:
                   - type
                   type: object
                 type: array
+              configChanged:
```
ciecierski, line 1913: This is not implemented in https://github.com/openstack-k8s-operators/dataplane-operator/pull/491

ciecierski (commented):
For this pr, we should first decide if we choose #491 or #490 for detecting nodeset spec changes

slagle:
@bshephar can we close this one out until it's time to revisit it again?

PR #501, opened 2023-11-01
Refactor base templates

bshephar:
This change removes some unnecessary items from the base config samples.

Namely:
1. Removes unnecessary or unused AnsibleVars such as the selinux variable, or variables that are just setting default or unnecessary values from edpm-ansible.
2. Limit the number of parameters configured in the baremetal-ipam sample, since these are already demonstrated in the base nodeset templates. Such as the spec.env example.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/644
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/508
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/900


bshephar:
Right, I see the ci-framework has quite a number of jobs that depend on the /spec/services path being at least defined. So maybe let's leave the services in the base templates for now and we can update the other samples to show how we add to them.

bshephar:
And Env for the baremetal one apparently..

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -42,15 +42,6 @@ spec:
       ansibleUser: cloud-admin
```
slagle, line 42: Can we default this in common.go and then drop it from the samples?

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -139,5 +130,3 @@ spec:
          # edpm firewall, change the allowed CIDR if needed
          edpm_sshd_configure_firewall: true
```
slagle, line 131: can we change the default value in edpm-ansible and then drop this? I guess I don't see why the default is false in edpm-ansible, if we are just going to always set it to true in every sample.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -139,5 +130,3 @@ spec:
...
          edpm_sshd_allowed_ranges: ['192.168.122.0/24']
```
slagle, line 132: This is something that is only needed for ci/dev, so we should probably make sure it's set there, then drop it here.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
slagle, line 1: can just set the ANSIBLE_* env vars as the defaults ansibleee-operator and then drop them from here?

slagle, line 1: can we set the default value for networkAttachments to ['ctlplane'], and then drop it from here? Seems unnecessary to have to set it in every sample, that might as well just be the default value.

slagle, line 1: what about all the other other ansible vars set here? if those are the defaults in edpm-ansible, we can drop them from here. Also consider setting reasonable defaults in edpm-ansible so we can drop them here.

slagle (changes requested):

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
bshephar: Yeah, no need for any even vars here. Dropping.

bshephar: Sounds good. Done in the latest iteration.

bshephar: I'll double check, but I don't think the others are using non-default values. I thought maybe `dns_search_domains`. But it doesn't seem to have a default.

So I can add a default in edpm-ansible and then remove from here like you mention in other comments here about some other variables.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
bshephar: Sounds good. Done in the latest iteration.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -42,15 +42,6 @@ spec:
       ansibleUser: cloud-admin
       ansiblePort: 22
```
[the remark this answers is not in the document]
bshephar: We can just remove it, since it's the ansible default. Removed in the latest iteration.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
bshephar: Yeah, that sounds like a good plan. I have no idea either, but my assumption was that someone did it for a reason. If neither of us know though, then I'll go propose changing it there.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
bshephar: Sounds good. I'll send a ci-framework PR for this one.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -25,32 +20,20 @@ spec:
...
+    networks:
+      - name: CtlPlane
+        subnetName: subnet1
+        defaultRoute: true
+      - name: InternalApi
+        subnetName: subnet1
+      - name: Storage
+        subnetName: subnet1
+      - name: Tenant
+        subnetName: subnet1
```
bshephar, line 33: If we use IPAM, we can eliminate all of the unnecessary `_ip` `_cidr` type Ansible vars below. This is probably advantageous as each of those presents a new opportunity to make a mistake. Using IPAM ensures they are generated automagically for you.

IMO, we should use IPAM for all of the samples for that reason. But happy to discuss

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
```diff
@@ -9,8 +9,8 @@ spec:
     - name: ANSIBLE_ENABLE_TASK_DEBUGGER
       value: "True"
   services:
-    - download-cache
     - bootstrap
+    - download-cache
```
rebtoor, line 13: why this?

download-cache should be the first service to be executed.

slagle: I think we can actually drop the services list entirely from the samples that are just using the default list.

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
slagle: I agree, let's use IPAM for all the samples, other than a single sample that doesn't use IPAM and shows just that deviation.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
slagle: i'm fine with a couple of different PRs

slagle: Oh, I think @rebtoor's question is valid though. Can't we just leave the list unchanged for now? You said you hit a problem when you dropped the services list, but can we just leave it for now and clean it up in a later PR?

slagle (commented):
The Depends-On have merged, but there is still 1 unresolved conversation to address.

bshephar:
recheck

slagle (approved):
All depends-on have merged

config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
```diff
@@ -25,30 +25,20 @@ spec:
         hostName: edpm-compute-0
         ansible:
           ansibleHost: 192.168.122.100
-          ansibleVars:
-            ctlplane_ip: 192.168.122.100
-            internal_api_ip: 172.17.0.100
-            storage_ip: 172.18.0.100
-            tenant_ip: 172.19.0.100
-            fqdn_internal_api: edpm-compute-0.example.com
-  networkAttachments:
```
karelyatin, line 34: This got removed unintentionally ?

karelyatin: pushed https://github.com/openstack-k8s-operators/dataplane-operator/pull/606

PR #502, opened 2023-11-01
Adding a non-voting adoption test job

jpodivin:


jpodivin:
recheck - retry limit

slagle (changes requested):
Even with the job as non-voting, I don't think this should be in github-check. My opinion is that I don't want this job running on every PR, it ought to be on demand triggered, or periodic.

jpodivin:
> Even with the job as non-voting, I don't think this should be in github-check. My opinion is that I don't want this job running on every PR, it ought to be on demand triggered, or periodic.

Adoption process is going to be one of the first experiences our customers have with our new Openstack. Apart from obvious complexity involved, it's for this reason that we should expend effort to test this process soon, and often.

The job does take time, but not substantially more than those we already have. The resources to run it are available, and even if it turned out to be unstable, which shouldn't happen, the non-voting nature will prevent it from blocking us.

zuul.d/projects.yaml
```diff
@@ -4,5 +4,7 @@
     github-check:
       jobs:
         - dataplane-operator-docs-preview
+        - data-plane-adoption-github-rdo-centos-9-crc-single-node:
```
fao89, line 7: I wonder if it is testing the dataplane-operator from the PR or if it is testing the dataplane that is pinned on openstack-operator

slagle:
> > Even with the job as non-voting, I don't think this should be in github-check. My opinion is that I don't want this job running on every PR, it ought to be on demand triggered, or periodic.
> 
> Adoption process is going to be one of the first experiences our customers have with our new Openstack. Apart from obvious complexity involved, it's for this reason that we should expend effort to test this process soon, and often.

Agreed, but it doesn't need to be part of testing on all PRs.

> The job does take time, but not substantially more than those we already have. The resources to run it are available, and even if it turned out to be unstable, which shouldn't happen, the non-voting nature will prevent it from blocking us.

We should be pushing more of our CI testing to be post-merge, not pre-merge. Are there specific gaps in kuttl or EnvTest that we can close that would provide similar benefit for pre-merge testing?

zuul.d/projects.yaml
jpodivin: afaik the job should be getting a new build from the CI framework.  https://review.rdoproject.org/r/plugins/gitiles/rdo-jobs/+/refs/heads/master/playbooks/data_plane_adoption/setup_crc_use_install_yamls.yaml#26 I suppose @jistr would have more info about that.

jistr: I don't know the answer to that but maybe @marios or @cescgina could.

One more thing i'd say is we should be adding the extracted adoption job, not the single node one. I think the extracted job is meant to supersede the single node one eventually, as it runs CRC without nested virt.

cescgina: @jistr is right, that in case of adding an adoption job, it should be the extracted crc one and not the single node. I have a PR open in install_yamls to setup the job there: https://github.com/openstack-k8s-operators/install_yamls/pull/608
right now, the job running is not consuming the output of the content-provider, since it's not listed as a dependency

jpodivin: Ok. Switching to `data-plane-adoption-github-rdo-centos-9-extracted-crc` .

PR #506, opened 2023-11-08
Cleanup bgp sample

bshephar:
This change removes unncessary variables from the BGP sample NodeSet
    config file. We remove the spec.env section, along with variables that
    are either unnecessary for this example, or variables that are just set
    to their defaults.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
@@ -33,24 +33,10 @@ spec:
             storage_ip: 172.18.0.100
             tenant_ip: 172.19.0.100
             fqdn_internal_api: edpm-compute-0.example.com
-  networkAttachments:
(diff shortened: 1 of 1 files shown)


config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
```diff
@@ -33,24 +33,10 @@ spec:
             storage_ip: 172.18.0.100
             tenant_ip: 172.19.0.100
             fqdn_internal_api: edpm-compute-0.example.com
-  networkAttachments:
-    - ctlplane
```
slagle, line 37: The kubebuilder marker that sets this default wasn't actually having any effect since it was on AnsibleEESpec. I've attempted to clean that up in https://github.com/openstack-k8s-operators/dataplane-operator/pull/591

slagle (approved):

slagle:
@bshephar Can we close this one out?

PR #508, opened 2023-11-08
Validating inventory variable names to comply with ansible reqs

jpodivin:
Not all variable names are accepted by ansible.[0] If we want to prevent issues during ansible runs, we need to have some sort of checks on them. This is the earliest point when we can detect issues with var names. Normally it would be something for kubebuilder, but that checks only values.  

I've used keyword list from python docs[1].

[0] https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#valid-variable-names 
[1] https://docs.python.org/3/reference/lexical_analysis.html#keywords

### DIFF
--- a/pkg/deployment/inventory.go
+++ b/pkg/deployment/inventory.go
@@ -27,6 +27,7 @@ import (
 
 	yaml "gopkg.in/yaml.v3"
 
+	validator "github.com/go-playground/validator/v10"
 	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/ansible"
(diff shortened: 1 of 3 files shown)


pkg/deployment/inventory.go
```diff
@@ -250,3 +271,55 @@ func buildNetworkVars(networks []infranetworkv1.IPSetNetwork) ([]string, map[str
...
+		"del",
+		"global",
+		"not",
+		"with",
+		"async",
+		"elif",
+		"if",
+		"or",
+		"yield":
+		return false
```
slagle, line 316: I really don't think we need to be keeping a list of python keywords in dataplane-operator. ansible variable validation is for ansible to do...not dataplane-operator. I understand the utility in trying to validate everything as early as possible, but not at the cost of having to write and maintain validation logic that is specific to an entirely separate tool (ansible in this case).

slagle (changes requested):
I don't think this belongs in dataplane-operator.  Our other operators wouldn't validate OpenStack configuration, that would be done by oslo_config instead. I don't see why we would add ansible validations to dataplane-operator. This should be done by ansible.

jpodivin:
> I don't think this belongs in dataplane-operator. Our other operators wouldn't validate OpenStack configuration, that would be done by oslo_config instead. I don't see why we would add ansible validations to dataplane-operator. This should be done by ansible.

Ansible already does this. by failing. Normally the ansible inventory is made by people/tools that know and respect conventions of ansible. So there is no chance of them adding invalid var name, or messing up the general datastructure. But here we are not using them. We're running our own code, constructing inventory from values supplied in CRD. Customer doesn't know that we are going to be using them for this, so they can't watch out for ansible stuff. And why would they?

This is OC after all. Why should a completely different tool matter. But in our case it does matter. We can cover the issue here, or we can move it somewhere else, like to lib-common. But if we do nothing we can open our customers to some ugly errors whenever we run something with this constructed inventory. Which can be often.

slagle:
> > I don't think this belongs in dataplane-operator. Our other operators wouldn't validate OpenStack configuration, that would be done by oslo_config instead. I don't see why we would add ansible validations to dataplane-operator. This should be done by ansible.
> 
> Ansible already does this. by failing. Normally the ansible inventory is made by people/tools that know and respect conventions of ansible. So there is no chance of them adding invalid var name, or messing up the general datastructure. But here we are not using them. We're running our own code, constructing inventory from values supplied in CRD. Customer doesn't know that we are going to be using them for this, so they can't watch out for ansible stuff. And why would they?
> 
> This is OC after all. Why should a completely different tool matter. But in our case it does matter. We can cover the issue here, or we can move it somewhere else, like to lib-common. But if we do nothing we can open our customers to some ugly errors whenever we run something with this constructed inventory. Which can be often.

I guess we disagree on the "ugly error" part. I'd rather expose the actual error from ansible to the user. We don't need another layer there. Let ansible do it's job, and we report the output to the user. Where does it stop? Do we keep adding things like this such as ansible syntax checkers, checking for modules, ansible configs, etc? In my opinion, no, and we don't need this either.

We are only generating an incredibly small number of ansible variables in dataplane-operator. I think the chance of using a python keyword there is very small.

PR #513, opened 2023-11-11
Cleanup hci sample file

bshephar:
This change removes unnecessary parameters in the Ceph HCI file. This change is part of a wider effort to eliminate complexity and technical debt in each of the sample files. They should only consist of parameters necessary for the service they represent while still remaining functional.

This change is removing the following parameters:
1. SELinux parameters that are not functional, along with the `service_net_map`.
2. All of the Network AnsibleVars that will be automatically generated since this sample file is using IPAM.
3. Parameters that are just setting default values.

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/501

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
@@ -5,28 +5,15 @@ metadata:
   namespace: openstack
 spec:
   env:
-  - name: ANSIBLE_CALLBACKS_ENABLED
-    value: profile_tasks
-  - name: ANSIBLE_FORCE_COLOR
-    value: "True"
+    - name: ANSIBLE_CALLBACKS_ENABLED
+      value: profile_tasks
+    - name: ANSIBLE_FORCE_COLOR
+      value: "True"
   networkAttachments:
   - ctlplane
   nodeTemplate:
     ansible:
-      ansiblePort: 22
-      ansibleUser: cloud-admin
       ansibleVars:
-        ctlplane_dns_nameservers:
-        - 192.168.122.1
-        ctlplane_gateway_ip: 192.168.122.1
-        ctlplane_host_routes:
-        - ip_netmask: 0.0.0.0/0
-          next_hop: 192.168.122.1
-        ctlplane_mtu: 1500
-        ctlplane_subnet_cidr: 24
-        dns_search_domains: []
-        timesync_ntp_servers:
-        - hostname: pool.ntp.org
         edpm_network_config_hide_sensitive_logs: false
         edpm_network_config_template: |
           ---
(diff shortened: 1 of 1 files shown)


config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
```diff
@@ -9,24 +9,13 @@ spec:
...
-        ctlplane_host_routes:
-        - ip_netmask: 0.0.0.0/0
-          next_hop: 192.168.122.1
-        ctlplane_mtu: 1500
-        ctlplane_subnet_cidr: 24
-        dns_search_domains: []
         timesync_ntp_servers:
         - hostname: pool.ntp.org
+        edpm_chrony_ntp_servers:
+        - pool.ntp.org
```
slagle, line 18: this was removed in https://github.com/openstack-k8s-operators/edpm-ansible/commit/9d8a7b2497cd26de25a61dfa4c6965a5f4b0a007, so the var can be dropped from here.

bshephar: ack, done.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
```diff
@@ -4,18 +4,10 @@ metadata:
   name: openstack-edpm-hci
   namespace: openstack
 spec:
-  env:
-  - name: ANSIBLE_CALLBACKS_ENABLED
-    value: profile_tasks
-  - name: ANSIBLE_FORCE_COLOR
-    value: "True"
-  - name: ANSIBLE_ENABLE_TASK_DEBUGGER
-    value: "True"
```
bshephar, line 13: This is going to fail actually. I need to address this in ci-framework and install_yamls before it will work here.

Tracking via:
https://issues.redhat.com/browse/OSPRH-2284

bshephar: I might just leave it in for now and then address the env: params in a seperate PR to reduce complexity and the number of Depends-On required.

slagle (commented):
+2. We can merge this once https://github.com/openstack-k8s-operators/dataplane-operator/pull/501 merges

bshephar:
recheck

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
```diff
@@ -25,8 +23,6 @@ spec:
         ctlplane_mtu: 1500
         ctlplane_subnet_cidr: 24
         dns_search_domains: []
-        timesync_ntp_servers:
```
rabi, line 28: Why're we removing this? Isn't it a valid default?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
```diff
@@ -5,16 +5,14 @@ metadata:
...
+    - name: ANSIBLE_FORCE_COLOR
+      value: "True"
   networkAttachments:
   - ctlplane
   nodeTemplate:
     ansible:
-      ansiblePort: 22
-      ansibleUser: cloud-admin
       ansibleVars:
         ctlplane_dns_nameservers:
```
rabi, line 17: These all are not required as this sample is using IPAM.

ctlplane_dns_nameservers
ctlplane_mtu
ctlplane_subnet_cidr
dns_search_domains

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
bshephar: Just removed it because it is the default, so doesn't need to be explicitly set here:
https://review.rdoproject.org/zuul/build/4583a2fda8314470917efb3e048bd2d8/log/ci-framework-data/logs/192.168.122.100/etc/chrony.conf#7

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph_hci.yaml
bshephar: Very true. I'll remove all of these now as well.

rabi (approved):
/lgtm

PR #519, opened 2023-11-14
Expose error handling services

bshephar:
Add handling for max_fail_percentage and any_errors_fatal

This change adds handling for the max_fail_percentage and
any_errors_fatal arguments. These are provided via --extra-vars as
edpm_max_fail_percentage and edpm_any_errors_fatal when used on the
OpenStackDataPlaneService CR's. Else, they can be provided globally to
all services like any other Ansible variable under ansibleVars on the
OpenStackDataPlaneNodeSet.

If values are not provided, or values are removed. Then we will default
to Ansible default values.


bshephar:
Unrelated Kuttl failure:

```
Error running must-gather collection:
    gather did not start for pod must-gather-6z4f5: unable to pull image: ImagePullBackOff: Back-off pulling image "quay.io/openstack-k8s-operators/openstack-must-gather:latest"
```

bshephar:
Confirmed the image can be pulled now:
```
❯ podman pull quay.io/openstack-k8s-operators/openstack-must-gather:latest
Trying to pull quay.io/openstack-k8s-operators/openstack-must-gather:latest...
Getting image source signatures
Copying blob sha256:9cba3a43324a1e54c728b7fb727b733f180dfcf18f24545e07d73b3fb67ff4d2
Copying blob sha256:42f5b36203c412784f366842a5e0a1b9be98741e6544df06da7c9e01d78e911f
Copying config sha256:f8f8b16929798f4173941fd8f3a676bebb00a43d222a159448186edf9ba10fa9
Writing manifest to image destination
f8f8b16929798f4173941fd8f3a676bebb00a43d222a159448186edf9ba10fa9
```

pkg/deployment/service.go
```diff
@@ -45,7 +45,7 @@ type ServiceYAML struct {
 // DeployService service deployment
 func DeployService(ctx context.Context, helper *helper.Helper, obj client.Object, sshKeySecret string, inventorySecret string, aeeSpec dataplanev1.AnsibleEESpec, foundService dataplanev1.OpenStackDataPlaneService) error {
 
-	err := dataplaneutil.AnsibleExecution(ctx, helper, obj, foundService.Spec.Label, sshKeySecret, inventorySecret, foundService.Spec.Play, foundService.Spec.Playbook, aeeSpec)
+	err := dataplaneutil.AnsibleExecution(ctx, helper, obj, &foundService, sshKeySecret, inventorySecret, aeeSpec)
```
fao89, line 48: sweet!

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -76,6 +76,15 @@ type OpenStackDataPlaneServiceSpec struct {
...
 	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty" yaml:"openStackAnsibleEERunnerImage,omitempty"`
+
+	// AnsibleMaxFailPercentage is used to tune service specific, allowable failure percentages during the Ansible execution
+	// https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#setting-a-maximum-failure-percentage
+	// +kubebuilder:validation:Optional
+	AnsibleMaxFailPercentage int64 `json:"ansibleMaxFailPercentage,omitempty"`
+
+	// AnsibleAnyErrorsFatal is used to tune service specific, any_errors_fatal
+	// +kubebuilder:validation:Optional
+	AnsibleAnyErrorsFatal bool `json:"ansibleAnyErrorsFatal,omitempty"`
```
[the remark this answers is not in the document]
bshephar: This would apply to all services though right? I guess it depends what we're targeting. My initial understanding of the requirement was that we want the ability to set these on a per-service basis. So that we could have different settings for bootstrap compared to telemetry for example.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -76,6 +76,15 @@ type OpenStackDataPlaneServiceSpec struct {
 	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
 	// +kubebuilder:validation:Optional
 	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty" yaml:"openStackAnsibleEERunnerImage,omitempty"`
+
+	// AnsibleMaxFailPercentage is used to tune service specific, allowable failure percentages during the Ansible execution
+	// https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#setting-a-maximum-failure-percentage
+	// +kubebuilder:validation:Optional
+	AnsibleMaxFailPercentage int64 `json:"ansibleMaxFailPercentage,omitempty"`
```
jpodivin, line 83: Please add kubebuilder constraint on range of values. We should only accept <0, 100> range.

jpodivin (changes requested):
min and max value from here should do the trick
https://book.kubebuilder.io/reference/markers/crd-validation.html

api/v1beta1/openstackdataplaneservice_types.go
bshephar: Done

jpodivin (approved):
It just needs a rebase to avoid errors.

api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
```diff
@@ -30,6 +30,13 @@ spec:
             type: object
           spec:
             properties:
+              ansibleAnyErrorsFatal:
+                type: boolean
+              ansibleMaxFailPercentage:
+                format: int64
```
rebtoor, line 36: can we just use `int`? I mean, i don't think there's a particular reason to use `int64` (and should be the default anyway on 64bit arch IIRC). If i add to pick a "custom" integer type I'd go with `uint`. WDYT?

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -76,6 +76,17 @@ type OpenStackDataPlaneServiceSpec struct {
...
+	// AnsibleMaxFailPercentage is used to tune service specific, allowable failure percentages during the Ansible execution
+	// https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#setting-a-maximum-failure-percentage
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=100
+	AnsibleMaxFailPercentage int64 `json:"ansibleMaxFailPercentage,omitempty"`
+
+	// AnsibleAnyErrorsFatal is used to tune service specific, any_errors_fatal
+	// +kubebuilder:validation:Optional
+	AnsibleAnyErrorsFatal bool `json:"ansibleAnyErrorsFatal,omitempty"`
```
slagle, line 89: I was thinking we would just set variables for these in ansibleVars on the NodeSet instead of adding new CRD fields. There's also an issue (https://issues.redhat.com/browse/OSPRH-1141) to add ansibleVars to OpenStackDataPlaneDeployment, so they could be set per deployment there.

What would be missing is setting them per-service, which is possible with the CRD fields as you've done here.

However, there is another issue with the CRD fields in that you can not modify the Service CRs of our standard Services since they will be reconciled by the NodeSet controller. So, I don't see how you could set these fields differently per execution, or even from what we ship without writing a custom service.

Or is the intent that you can use either the fields or set ansible variables?

bshephar: Given the current proposal, it would work in both ways right. You could just adds the vars to ansibleVars and all services would receive them. Or, if you are adding your own service and you don't want it to stop the overall deployment you could configure the individual service.

We could configure our default services though, for example everything needs to pass bootstrap, configure-network, etc. But telemetry could block the entire deployment with a single failure. So tuning that service to allow a percentage of failures, or not having any_errors_fatal could be advantageous for those types of services. In that case, we could update our default service definition to tune it appropriately.

api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
bshephar: I changed it to int, I think I just tab completed the suggestion from my editor tbh. I didn't give it too much thought, but int sounds better to me.

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
fao89: yes, it would apply to all services (I think it would be useful in some cases),  but now I agree with https://github.com/openstack-k8s-operators/dataplane-operator/pull/519#discussion_r1396042110

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -76,6 +76,17 @@ type OpenStackDataPlaneServiceSpec struct {
...
+	// AnsibleMaxFailPercentage is used to tune service specific, allowable failure percentages during the Ansible execution
+	// https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#setting-a-maximum-failure-percentage
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=100
+	AnsibleMaxFailPercentage int `json:"ansibleMaxFailPercentage,omitempty"`
+
+	// AnsibleAnyErrorsFatal is used to tune service specific, any_errors_fatal
+	// +kubebuilder:validation:Optional
+	AnsibleAnyErrorsFatal bool `json:"ansibleAnyErrorsFatal,omitempty"`
```
gibizer, line 89: Having a bool with omitempty means the False value will always be ignored. Therefore this field once set to True cannot be explicitly set to False, as when the client sends False the json serializer after the webhook code will omit the False value from the json body as False is the empty value. You probably need a *bool to allow differentiation between False value and case when value is not provided.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -76,6 +76,17 @@ type OpenStackDataPlaneServiceSpec struct {
 	// OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image
 	// +kubebuilder:validation:Optional
 	OpenStackAnsibleEERunnerImage string `json:"openStackAnsibleEERunnerImage,omitempty" yaml:"openStackAnsibleEERunnerImage,omitempty"`
+
+	// AnsibleMaxFailPercentage is used to tune service specific, allowable failure percentages during the Ansible execution
+	// https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#setting-a-maximum-failure-percentage
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=100
+	AnsibleMaxFailPercentage int `json:"ansibleMaxFailPercentage,omitempty"`
```
gibizer, line 85: You mentioned safe defaults in the commit message but I don't see those defaults here. Where are they set?

api/v1beta1/openstackdataplaneservice_types.go
bshephar: Ah nice. Thanks!

api/v1beta1/openstackdataplaneservice_types.go
bshephar: Yeah, I think I need to update my wording there. We'll just fallback to what Ansible would do without these arguments provided.

api/v1beta1/openstackdataplaneservice_types.go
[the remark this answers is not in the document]
bshephar: Ok, it will work that way anyway. So if you just want to do it globally to all services, you could just provide the values under the `ansibleVars`. If you want to be more granular about it on a per service basis, then you could set these values on the service itself.

api/v1beta1/openstackdataplaneservice_types.go
slagle: In one case it's a var under ansibleVars, but when it's on the service, it's a custom field. The custom field takes precedence since it's implemented as using extra vars, but that's not clear by looking at the CRD API (that's something users would have to remember).

I'd rather see it consistently set if we're going to have it in both places. We could even call the variables field ansibleExtraVars on the Service, which makes it more clear that these variables would be set using the ansible extra vars interface, which would be familiar to ansible users.

bshephar: Ok, that sounds reasonable. So I don't think we want that interface to be extensible enough to set any arbitrary variable though right? Like, we want to limit this interface on the service to specifically these two error handling variables? So, if I implement this with a `ansibleExtraVars` struct, then move the two currently defined parameters under that struct. Would that satisfy your concern on this one?

gibizer (commented):
I'm wondering about the high level use case of this.
By default the ansible run succeeds when every task succeeds. So if something fails it the AEE and Deployment failure will be clearly visible. However when the fail_percentage is defined there can be a success recorded AEE and Deployment CRs even if a given percentage of tasks failed. How will the human deployer know that it was a truly successful run or a partially successful run? Do we need to give some guidance (doc) to the human about it? 

Personally I would be very grumpy troubleshooting a customer case where some compute behaves strangely even though the deployment seems to be succeeded just to later figure out that one particular task on that one compute failed to execute causing issue a lot later. As a random concrete example: The task putting the ssh config for the nova user fails on that particular compute. Everything seems to work until the first VM is about to cold migrated away from the given compute, which will fail due to the missing ssh config.

pkg/util/ansible_execution.go
```diff
@@ -90,15 +87,23 @@ func AnsibleExecution(
 		if len(aeeSpec.AnsibleSkipTags) > 0 {
 			fmt.Fprintf(&cmdLineArguments, "--skip-tags %s ", aeeSpec.AnsibleSkipTags)
 		}
+		if service.Spec.AnsibleExtraVars != (dataplanev1.AnsibleExtraVars{}) {
+			if service.Spec.AnsibleExtraVars.AnsibleMaxFailPercentage != 0 {
+				fmt.Fprintf(&cmdLineArguments, "--extra-vars edpm_max_fail_percentage=%d ", service.Spec.AnsibleExtraVars.AnsibleMaxFailPercentage)
+			}
+			if !*service.Spec.AnsibleExtraVars.AnsibleAnyErrorsFatal {
```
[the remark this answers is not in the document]
bshephar: Right, yeah. I think we need to validate that this isn't nil before trying to use it. I have updated that now.

bshephar:
recheck

jpodivin:
recheck

Failed on tempest.

fao89:
tempest is broken for all PRs, seems to be an issue with neutron: https://logserver.rdoproject.org/29/529/f8257dd139519f166655747656659877c32c77cc/github-check/podified-multinode-edpm-deployment-crc/1651826/controller/ci-framework-data/logs/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-0812f031e363406238d47a3ba3cfb33412e9a2d143d2b4b9365c9796b80bb8aa/namespaces/openstack/pods/neutron-846f77ffc7-59wxz/logs/neutron-api.log

api/v1beta1/openstackdataplaneservice_types.go
slagle: I think we need to think about where we are setting ansible variables and how they are used.

We set them on the nodes and nodeSet, and those end up as host and group vars in the inventory. This makes sense.

We'll likely vars to the Deployment, and those should use the ExtraVars interface and override all other vars from the inventory (https://issues.redhat.com/browse/OSPRH-1141).

Adding variables to the Service CRD as well might make sense for this use case, but how is the precedence handled and how are they implemented? If they also use ExtraVars, that could conflict with setting them on the Deployment. I don't want to create new precedence rules between dataplane-operator CRDs that users have to remember.

That being said, I'm not sure we need to add variables for these to the Service CRD. Can't we just keep the default values in the playbooks as you've done in https://github.com/openstack-k8s-operators/edpm-ansible/pull/494, use custom variables per playbook, and the user can override those per service by setting the right variable in either the node/nodeSet/Deployment?

For instance playbooks/bootstrap.yml uses edpm_bootstrap_ any_errors_fatal and edpm_bootstrap_max_fail_percentage. And we can fine tune those defaults for each service in the playbooks directly.

bshephar: Ok, we can abandon this one and just push the edpm-ansible one through

jpodivin (approved):

gibizer:
/hold 
The below comment from me was missed:

> I'm wondering about the high level use case of this. By default the ansible run succeeds when every task succeeds. So if something fails it the AEE and Deployment failure will be clearly visible. However when the fail_percentage is defined there can be a success recorded AEE and Deployment CRs even if a given percentage of tasks failed. How will the human deployer know that it was a truly successful run or a partially successful run? Do we need to give some guidance (doc) to the human about it?
> 
> Personally I would be very grumpy troubleshooting a customer case where some compute behaves strangely even though the deployment seems to be succeeded just to later figure out that one particular task on that one compute failed to execute causing issue a lot later. As a random concrete example: The task putting the ssh config for the nova user fails on that particular compute. Everything seems to work until the first VM is about to cold migrated away from the given compute, which will fail due to the missing ssh

PR #520, opened 2023-11-15
Update samples for max_fail_percentage and any_errors_fatal

bshephar:
This change adds examples for how to define max_fail_percentage and any_errors_fatal. These were added by:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/519


bshephar:
recheck

PR #521, opened 2023-11-15
Remove unnecessary nilness check

bshephar:
This change removes an unnecessary nilness check in each of the controllers.


controllers/openstackdataplanedeployment_controller.go
```diff
@@ -102,9 +101,6 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
 		r.Scheme,
 		logger,
 	)
-	if err != nil {
```
bshephar, line 105: Either this, or we should replace the null identifier with err when we initialise the helper.

bshephar:
recheck

slagle (approved):

PR #525, opened 2023-11-16
Remove check for BaremetalHosts in defaulting webhook

rabi:
This creates a problem during update to scale-up as the spec passed to defaulting webhook contains earlier BaremetalHosts.


abays (approved):
/lgtm

PR #531, opened 2023-11-21
Removing ANSIBLE_ENABLE_TASK_DEBUGGER

fao89:
ref: https://github.com/ansible/ansible-runner/issues/1327


rebtoor (approved):
we can't use the ansible task debugger as is since ansible-runner is supposed to run in a non-interactive way, so it's safe to remove this var.

bshephar (approved):

fao89:

CI failure is related to variable names from https://github.com/openstack-k8s-operators/ovn-operator/blob/main/config/default/manager_default_images.yaml x https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/default/manager_default_images.yaml

PR #540, opened 2023-11-24
Add a sample for deploying with BGP and OVN routing

luis5tb:
Kernel routing is used by default when deploying with BGP, but there is a new option to perform that routing directly in OVN, so that datapath acceleration is supported. This is a new template with a sample the configuration required for that.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
@@ -1,0 +1,166 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNodeSet
+metadata:
+  name: openstack-edpm
+spec:
+  services:
+    - download-cache
+    - bootstrap
+    - configure-network
+    - validate-network
+    - frr
+    - install-os
+    - configure-os
+    - run-os
+    - ovn
+    - neutron-metadata
+    - ovn-bgp-agent
+    - libvirt
+    - nova
+    - telemetry
+  preProvisioned: true
+  nodes:
+      edpm-compute-0:
+        hostName: edpm-compute-0
+        ansible:
+          ansibleHost: 192.168.122.100
+          ansibleVars:
+            ctlplane_ip: 192.168.122.100
+            internal_api_ip: 172.17.0.100
+            storage_ip: 172.18.0.100
+            tenant_ip: 172.19.0.100
+            fqdn_internal_api: edpm-compute-0.example.com
+  networkAttachments:
+    - ctlplane
+  nodeTemplate:
+    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+    managementNetwork: ctlplane
+    ansible:
+      ansibleUser: cloud-admin
+      ansiblePort: 22
+      ansibleVars:
+         timesync_ntp_servers:
+           - hostname: pool.ntp.org
+         # edpm_network_config
+         # Default nic config template for a EDPM compute node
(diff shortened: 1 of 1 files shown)


config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
```diff
@@ -0,0 +1,174 @@
...
+    - install-os
+    - configure-os
+    - run-os
+    - ovn
+    - neutron-metadata
+    - ovn-bgp-agent
+    - libvirt
+    - nova
+    - telemetry
+  preProvisioned: true
```
bshephar, line 24: Do we only support BGP with pre-provisioned nodes? I recall we had issues with IPAM in the last iteration of this work because we couldn't use the same IP address multiple times? 

Is this still the case?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
```diff
@@ -0,0 +1,174 @@
...
+  nodeTemplate:
+    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+    managementNetwork: ctlplane
+    ansible:
+      ansibleUser: cloud-admin
+      ansiblePort: 22
+      ansibleVars:
+         service_net_map:
+           nova_api_network: internal_api
+           nova_libvirt_network: internal_api
```
bshephar, line 47: I know other samples have this, but we need to clean all of them up too. There is nothing that uses `service_net_map` though. So we can drop this.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
```diff
@@ -0,0 +1,174 @@
...
+         ctlplane_dns_nameservers:
+         - 192.168.122.1
+         dns_search_domains: []
+         gather_facts: false
+         enable_debug: false
+         # edpm firewall, change the allowed CIDR if needed
+         edpm_sshd_configure_firewall: true
+         edpm_sshd_allowed_ranges: ['192.168.122.0/24']
+         # SELinux module
+         edpm_selinux_mode: enforcing
```
bshephar, line 162: Nothing uses this in edpm-ansible

bshephar (changes requested):
I know you have kept this in-line with the existing samples. But all of them need cleaning up too. Ideally, I would like to reduce the amount of stuff that needs cleaning up as we move forward. So ideally, my comments are focused on that objective.

If we can just try to keep only the necessary parts for what your sample is demonstrating, along with those required for it to be actually functional, that would be ideal.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
luis5tb: I think I never tried the non-preProvisioned option but I don't see why it should not work

slagle (commented):
We've got a growing number of samples, and I think we need to get that under control before adding more. That is happening here:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/501

That will greatly simplify what you'd need to maintain here, in that it would be just the pieces that are specific to BGP and OVN routing.

bshephar (approved):

bshephar:
> #501

We want the bare minimum required to show how to deploy BGP with OVN Routing, but also remain functional right?

slagle:
> > #501
> 
> We want the bare minimum required to show how to deploy BGP with OVN Routing, but also remain functional right?

Sure, but this adds more than the minimum.

PR #544, opened 2023-11-28
Add reboot-os service to dataplane-operator

ciecierski:
Add reboot-os service to sample services and add it to default services list.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/503

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -1869,6 +1869,7 @@ spec:
                 - install-os
                 - configure-os
                 - run-os
+                - reboot-os
                 - ovn
                 - neutron-metadata
                 - libvirt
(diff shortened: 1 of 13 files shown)


fao89 (commented):
please update the nodeset samples here:
https://github.com/openstack-k8s-operators/dataplane-operator/tree/main/config/samples
So the CI can test the new service

rebtoor:
if @fao89 request is addressed, it's a +2 for me.

slagle:
> please update the nodeset samples here: https://github.com/openstack-k8s-operators/dataplane-operator/tree/main/config/samples So the CI can test the new service

@bshephar Is in the process of simplifying these samples, and dropping the services (especially if it's the default list) is part of that. That is happening here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/501
So for me, I wouldn't feel the need to block this one on updating those samples, since it will just get removed in the near term anyway.

slagle (commented):
per my previous comment, this LGTM

fao89:
but we still need [these samples ](https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L328-L330)to be updated so the CI can test it

slagle (commented):
This LGTM now, just pending the depends-on, https://github.com/openstack-k8s-operators/edpm-ansible/pull/503

fao89:
recheck

ciecierski:
> but we still need [these samples ](https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L328-L330)to be updated so the CI can test it

Reboot service add to samples

fao89 (approved):
Thank you!

PR #546, opened 2023-11-28
[doc]Change ssh migration key to ECDSA

gibizer:
Even though the recent FIPS standard[1] allows ED25519 ssh keys, the current FIPS profile in centos9 stream does not allow it. So the current migration ssh key isn't usable. This PR changes the key type to ECDSA until the profile is updated.

[1] https://csrc.nist.gov/pubs/fips/186-5/final


gibizer:
Related change https://github.com/openstack-k8s-operators/install_yamls/pull/641

fao89 (approved):

PR #549, opened 2023-11-30
DNS Server config issue fix

Jaganathancse:


config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
```diff
@@ -55,6 +55,9 @@ spec:
          # Default nic config template for a EDPM compute node
          # These vars are edpm_network_config role vars
          edpm_network_config_hide_sensitive_logs: false
+         ctlplane_dns_nameservers:
```
rabi, line 58: You don't have  to add it here, should be populated with IPAM[1]

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/inventory.go#L113

rabi (changes requested):

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
```diff
@@ -55,6 +55,9 @@ spec:
...
+         - 192.168.122.80
+         dns_search_domains: []
```
bshephar, line 60: The variable is `edpm_dns_search_domains` now:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_network_config/defaults/main.yml#L53

But this is just setting the default value anyway, so we can just leave it out of the sample file to avoid unnecessary clutter.

bshephar (changes requested):
I'm not entirely sure this change is necessary. But if we want to pursue it, I have left a comment in addition to what Rabi has already pointed out.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
slagle: Agree with Rabi and Brendan in that I'm not sure this is needed. @Jaganathancse The commit message should explain what is being fixed, and describe the issue, and how it's fixed. Can you add some more detail about why this PR might be needed?

slagle (changes requested):

Jaganathancse:
This change is not required.

PR #550, opened 2023-11-30
Always reset SetupReadyCondition to False

rabi:
We use this condition to trigger deployments for a nodeset. Without this being reset to false, deployments would tigger before nodeset is reconciled for update/patch as the condition would be true.

### DIFF
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -183,9 +183,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, nil
 	}
 
-	if instance.Status.Conditions.IsUnknown(dataplanev1.SetupReadyCondition) {
-		instance.Status.Conditions.MarkFalse(dataplanev1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
-	}
+	instance.Status.Conditions.MarkFalse(dataplanev1.SetupReadyCondition, condition.RequestedReason, condition.SeverityInfo, condition.ReadyInitMessage)
 
 	// Ensure Services
 	err = deployment.EnsureServices(ctx, helper, instance)


rabi:
recheck

slagle:
I'd like to understand this a little better, and the scenario that's being solved.

> deployments would tigger
> before nodeset is reconciled for update/patch as the condition
> would be true.

Only if the NodeSet is already SetupReady=True, and the deployment was already reconciling right? Specifically, are you trying to address the case where someone patches a NodeSet during an already reconciling Deployment?

It's likely an inconsistent pattern, but I don't see other conditions in other operators working the same as you're proposing here. Example, InputReady getting marked to False on every reconcile loop.

Forcing SetupReady=False on every reconcile seems unnecessary IMO, and may be inaccurate in many cases (we don't actually know that SetupReady is indeed False, granted we don't actually know that it's True either on every reconcile until we've checked).

rabi:
> Only if the NodeSet is already SetupReady=True, and the deployment was already reconciling right? Specifically, are you trying to address the case where someone patches a NodeSet during an already reconciling Deployment?

Nah, this is the scale-out case where a nodeset and deployment are both already reconciled. Then you patch the nodeset to add more nodes to the nodeset and create a (new) deployment. Because the nodset status is already SetupReady=True[1], the new deployment moves ahead and uses the old inventory and  runs the aee jobs and the scale-out nodes are not deployed.

> we don't actually know that SetupReady is indeed False, granted we don't actually know that it's True either on every reconcile until we've checked

During every reconcile we don't not know what is a specific status till reconciliation is gone through that step and updated the status. The assumption is that there won't be any reconcile if there is no spec change and if there is a spec change we don't know any of the statuses.

The other option is to evaluate SetupReady based on all other statuses that should be ready before that and in that case also we've to reset it back at the beginning of the reconciliation.




[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanedeployment_controller.go#L148

bshephar (approved):

rabi:
error: resource mapping not found for name: "openstack-galera" namespace: "openstack" from "STDIN": no matches for kind "OpenStackControlPlane" in version "core.openstack.org/v1beta1"
ensure CRDs are installed first

PR #553, opened 2023-12-01
Add infrastructure for edpm certs

vakwetu:
This PR adds the ability to generate and mount TLS certs in the ansibleEE
    pod for each service.
    
    The basic idea is as follows:

    There is a top level parameter on the nodeset (TLSEnabled) that determines whether
    TLS-E is enabled.  This is analogous to a similar parameter on the control plane.
    
    Each service has a boolean parameter HasTLSCerts which determines
    whether or not certs are expected for this service.  The certs will be issued for each
    node in the nodeset and then concatenated into a secret that contains the tls certs and keys
    for all the nodes in the node set.  They are referenced as <nodeName>-tls.key, nodeName-tls.crt.
    The secret is named <nodeSet.Name>-<service.Name>-certs . 
    
    This secret will be mounted in the openstackAnsibleEE pod at /var/lib/openstack/certs.  It will be up to the ansible  
    playbook to move the appropriate certs, keys etc. for each node to the node, and do
    any reconfiguration of the service.


slagle:
This is the same PR as https://github.com/openstack-k8s-operators/dataplane-operator/pull/442. For some reason, github won't allow re-opening the old PR, even though there are new commits.

slagle (approved):

PR #561, opened 2023-12-06
Move TLS logic to deployment controller

vakwetu:
As noted in review, a user might select an entirely different set of
override services in a deployment.  To handle this (and the default
case), we move the cert handling logic to the OSDPDeployment controller.


rabi (approved):
/lgtm

PR #562, opened 2023-12-07
Check for all deployments for a NodeSet

rabi:
It's possible that there are more than one deployment with different names for the nodeset (earlier ones already deploted). Check all deployments before returning the deployment status.


rabi:
recheck

slagle (approved):

PR #563, opened 2023-12-07
Remove unused parameters from cert.go functions

bshephar:
This change removes the unused nodeName parameter from the GetTLSNodeCert function.


rabi (approved):
/lgtm

PR #566, opened 2023-12-08
Bump controller-runtime to v0.16.0

bshephar:


fao89 (commented):
I think controller-runtime bump should come from upgrading the operator-sdk

PR #567, opened 2023-12-09
Don't use nodeName in inventory

rabi:
We use nodeName to generate number of ansible vars in inventory and that causes issues when using pre-provisioned nodes with a different hostname.

### DIFF
--- a/pkg/deployment/inventory.go
+++ b/pkg/deployment/inventory.go
@@ -53,12 +53,12 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 	}
 
 	for nodeName, node := range instance.Spec.Nodes {
-		host := nodeSetGroup.AddHost(nodeName)
+		host := nodeSetGroup.AddHost(strings.Split(node.HostName, ".")[0])
 		// Use ansible_host if provided else use hostname. Fall back to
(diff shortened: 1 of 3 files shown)


rabi:
recheck bm job kernel panic

rabi:
recheck

pkg/deployment/inventory.go
```diff
@@ -53,12 +53,12 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 	}
 
 	for nodeName, node := range instance.Spec.Nodes {
-		host := nodeSetGroup.AddHost(nodeName)
+		host := nodeSetGroup.AddHost(strings.Split(node.HostName, ".")[0])
```
fao89, line 56: should we make Hostname required?
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/common.go#L52-L53

rabi: It's defaulted to nodeName in webhook[1].

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanenodeset_webhook.go#L59

fao89 (approved):
thank you!

PR #580, opened 2023-12-14
Don't add ansibleHost when using IPAM

rabi:
Unless using predictable ip, don't add ansibleHost as it won't be the IP always reserved by IPAM.


slagle (approved):

rabi:
recheck

rabi:
recheck

rabi:
recheck kernel panic

PR #584, opened 2023-12-18
standardize Makefile for operator-sdk

yazug:
copy operator-sdk related stuff from skeleton
generated by operator-sdk init

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,10 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 	BUNDLE_GEN_FLAGS += --use-image-digests
 endif
 
+# Set the Operator SDK version to use. By default, what is installed on the system is used.
+# This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
+OPERATOR_SDK_VERSION ?= v1.31.0
+
 # Image URL to use all building/pushing image targets
 DEFAULT_IMG ?= quay.io/openstack-k8s-operators/dataplane-operator:latest
 IMG ?= $(DEFAULT_IMG)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -248,13 +250,25 @@ $(KUTTL): $(LOCALBIN)
 	chmod +x $(LOCALBIN)/kubectl-kuttl
 
 .PHONY: operator-sdk
-operator-sdk: $(OPERATOR_SDK) ## Download operator-sdk locally if necessary.
-$(OPERATOR_SDK): $(LOCALBIN)
-	test -s $(OPERATOR_SDK) || curl -o $(LOCALBIN)/operator-sdk -L https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_linux_amd64 && chmod +x $(LOCALBIN)/operator-sdk
+OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
+operator-sdk: ## Download operator-sdk locally if necessary.
+ifeq (,$(wildcard $(OPERATOR_SDK)))
```
rabi, line 255: Why're we using wildcard here? Isn't the binary always named as operator-sdk or whatever name in $OPERATOR_SDK?

yazug: This is the current Makefile from operator-sdk skeleton.  We are bringing all of the operators up to the same template as there have been some divergence.

As to why the wildcard I am not sure.  we would have to look into the operator-sdk changes that introduced that for the skeleton Makefile

rabi: which skeleton Makefile we' are talking about here?  I guess it's something we've written.

yazug: the skeleton Makefile that is generated if you run "operator-sdk init ... "  as if you were creating a new operator from scratch.  It generates a skeleton operator including a Makefile for building/testing and what I synced in with this patch brings our current Makefile closer to the latest generation as well as bring in some automation standardization improvements in the Makefile.  It allows the Makefile to better pick up or use provided OPERATOR_SDK var as well as several others but also just work for local dev workflow as well.

rabi: > It allows the Makefile to better pick up or use provided OPERATOR_SDK var

Though I don't see any reason why we should look for sdk binary with wildcard function, if this what is generated by operator-sdk we can go with it.

rabi (approved):
/lgtm

Makefile
yazug: This construct allows the Makefile to check to see if first $(OPERATOR_SDK) is present locally in the filesystem, then it checks to see if operator-sdk is present (update OPERATPR_SDK to which operator-sdk),  if it is not present then download the cli and use that.

rabi: I understand that but you don't use a wildcard to look if a file `$(OPERATOR_SDK) ` exists :smile:

PR #585, opened 2023-12-20
Add dedicated certs service

vakwetu:
As discussed, this will add a dedicated certs service that will create and copy in certs before the other services run.
This will allow the subsequent services to use all the certs that were created beforehand.

One weirdness I found is that the NodeSetConditions map was being set to nil by the defer() function, causing a panic.
I put in a workaround to keep things going, but this needs investigation. 

Depends on: https://github.com/openstack-k8s-operators/edpm-ansible/pull/525


vakwetu:
weirdness about NodeSetConditions fixed - I was running new code with old crds.  We likely need an update in CRD as in https://github.com/openstack-k8s-operators/openstack-operator/pull/604

tests/kuttl/tests/dataplane-deploy-no-nodes-test/01-assert.yaml
```diff
@@ -398,6 +399,56 @@ status:
...
+  ownerReferences:
+  - apiVersion: dataplane.openstack.org/v1beta1
+    blockOwnerDeletion: true
+    controller: true
+    kind: OpenStackDataPlaneDeployment
+    name: edpm-compute-no-nodes
+spec:
+  backoffLimit: 6
+  extraMounts:
+  - mounts:
```
jpodivin, line 414: Won't the service add some volumes of it's own?

jpodivin (commented):
I have some questions, but all things taken into account I think we can move forward.

pkg/deployment/deployment.go
```diff
@@ -196,6 +207,78 @@ func (d *Deployer) ConditionalDeploy(
...
+	services []string,
+) (*dataplanev1.AnsibleEESpec, error) {
+	log := d.Helper.GetLogger()
+	client := d.Helper.GetClient()
+	for _, service := range services {
+		foundService, err := GetService(d.Ctx, d.Helper, service)
+		if err != nil {
+			return nil, err
+		}
+		if foundService.Spec.TLSCertsEnabled != nil && *foundService.Spec.TLSCertsEnabled {
```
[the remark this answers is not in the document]
vakwetu: done

tests/kuttl/tests/dataplane-deploy-no-nodes-test/01-assert.yaml
vakwetu: Only when we do the test with tlsEnabled: true

That's coming -- and you've already been involved in that patch.

pkg/deployment/deployment.go
```diff
@@ -84,10 +84,20 @@ func (d *Deployer) Deploy(services []string) (*ctrl.Result, error) {
...
 		copy(d.AeeSpec.ExtraMounts, aeeSpecMounts)
 		d.AeeSpec, err = d.addServiceExtraMounts(foundService)
-
 		if err != nil {
 			return &ctrl.Result{}, err
 		}
+
+		// Add certMounts to the install-certs service if TLS enabled
+		if *d.NodeSet.Spec.TLSEnabled {
+			if foundService.Name == InstallCertsLabel {
```
slagle, line 93: This is more service specific logic in the code, which I really don't think is needed. I left this comment on the initial PR, but agreed to allow it to move forward for progress sake, ... but nova/libvirt is still hardcoded. Now this hardcodes something for install-certs. We need to find a different way.

Off the top of my head, an AddCertMounts bool on the Service type, could be a bool used to trigger this logic. This would also keep this logic functional when using non-default services. We don't want restrictions like "you must have a service called install-certs if you have any other services that use certs". Instead, the bool provides a more usable interface to enable/disable this functionality.

Otherwise, this looks ok to me as a general direction. Although, it would be nice if we could eliminate the extra loop of fetching all the services somehow.

slagle (dismissed):

pkg/deployment/deployment.go
vakwetu: ACK.  Your initial comments about the service specific code hasn't been forgotten.  The issue has been captured in https://issues.redhat.com/browse/OSPRH-2389 which is on-deck to be tackled shortly.

Will add a boolean for this, as suggested.

docs/openstack_dataplaneservice.md
```diff
@@ -125,9 +125,10 @@ OpenStackDataPlaneServiceSpec defines the desired state of OpenStackDataPlaneSer
 | configMaps | ConfigMaps list of ConfigMap names to mount as ExtraMounts for the OpenStackAnsibleEE | []string | false |
 | secrets | Secrets list of Secret names to mount as ExtraMounts for the OpenStackAnsibleEE | []string | false |
 | openStackAnsibleEERunnerImage | OpenStackAnsibleEERunnerImage image to use as the ansibleEE runner image | string | false |
-| tlsCertsEnabled | TLSCertsEnabled - Whether the nodes have TLS certs | *bool | false |
+| tlsCertsEnabled | TLSCertsEnabled - Whether the nodes have TLS certs | bool | true |
```
slagle, line 128: Why does this show the default as true? I do see you have default as false where the field is actually defined? Do the docs need to be regenerated?

Same for tlsEnabled.

vakwetu: The final column here is not the default value, but rather whether or not the field is required.

You'll notice I changed from a pointer to a boolean to plain boolean with a default value (which is set to false)

pkg/deployment/deployment.go
vakwetu: See https://github.com/openstack-k8s-operators/dataplane-operator/pull/610 for the changes to resolve your concerns with cert contents

docs/openstack_dataplaneservice.md
slagle: ok, that makes sense.

vakwetu:
recheck

slagle (commented):
+2

A kuttl test that covers the functionality you added would be a good follow up.

vakwetu:
kuttl test already in progress. thanks

vakwetu:
The changes in this pull request were already merged in #610

PR #590, opened 2024-01-03
Add sample config for the networker deployment

slawqo:
This patch adds config sample for the dataplane deployment with networker node.
It also renames sample nodeset with networker nodes from "openstack-edpm" to "openstack-edpm-networker" and renames host names in that sample file to be "edpm-networker-..." instead of "edpm-compute-...".

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
@@ -1,7 +1,7 @@
 apiVersion: dataplane.openstack.org/v1beta1
 kind: OpenStackDataPlaneNodeSet
 metadata:
-  name: openstack-edpm
+  name: openstack-edpm-networker
 spec:
   env:
     - name: ANSIBLE_FORCE_COLOR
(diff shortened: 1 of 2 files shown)


slagle (commented):
This looks ok, but I don't see the sample deployment, only the change to the NodeSet. Did you intend to add something else?

config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
```diff
@@ -18,16 +18,16 @@ spec:
     - telemetry
```
luis5tb, line 18: now that you are reshaping this... I think the telemetry is related to nova/libvirt and can be removed too

luis5tb:
lgtm, just a nit in there, and the same comment as James... it something else missing?

slawqo:
> This looks ok, but I don't see the sample deployment, only the change to the NodeSet. Did you intend to add something else?

Sorry but I'm not sure I understand Your question. Isn't change in config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml actually change in the sample deployment? Should I add it somewhere else too?

slawqo:
> This looks ok, but I don't see the sample deployment, only the change to the NodeSet. Did you intend to add something else?

Sorry. I had new deployment sample prepared locally but forgot to add it to commit as it was new file :) Now it should be there

fao89 (approved):

jpodivin (approved):

PR #591, opened 2024-01-03
Drop kubebuilder markers on AnsibleEESpec

slagle:
The AnsibleEESpec struct is not part the CRD API. The struct is created
by the code later for use. Therefore, kubebuilder markers have no effect
on its fields. The only marker that was not already set on other fields
was the default for NetworkAttachments, which is moved to the
NetworkAttachments field on OpenStackDataPlaneNodeSetSpec struct.

### DIFF
--- a/api/v1beta1/common.go
+++ b/api/v1beta1/common.go
@@ -116,25 +116,18 @@ type NodeTemplate struct {
 
 // AnsibleEESpec is a specification of the ansible EE attributes
 type AnsibleEESpec struct {
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={ctlplane}
 	// NetworkAttachments is a list of NetworkAttachment resource names to pass to the ansibleee resource
 	// which allows to connect the ansibleee runner to the given network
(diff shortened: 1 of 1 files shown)


fao89 (approved):

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -52,6 +52,7 @@ type OpenStackDataPlaneNodeSetSpec struct {
 	Env []corev1.EnvVar `json:"env,omitempty"`
 
 	// +kubebuilder:validation:Optional
```
fao89, line 54: you have to drop this
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_dataplane-operator/591/pull-ci-openstack-k8s-operators-dataplane-operator-main-precommit-check/1742579977779417088

slagle: I actually dropped the default value. I think that's the right thing to do here.

fao89 (approved):

fao89:
recheck

PR #592, opened 2024-01-03
Adding AnsibleOpts default values

fao89:
If we set the default with kubebuilder markers, it will force us to set the value for each node. 
Instead, we should set the default values using the webhook to ensure the default values are only set on the nodetemplate

### DIFF
--- a/api/v1beta1/openstackdataplanenodeset_webhook.go
+++ b/api/v1beta1/openstackdataplanenodeset_webhook.go
@@ -61,6 +61,10 @@ func (spec *OpenStackDataPlaneNodeSetSpec) Default() {
 		spec.Nodes[nodeName] = *node.DeepCopy()
 	}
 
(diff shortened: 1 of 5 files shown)


api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -61,6 +61,13 @@ func (spec *OpenStackDataPlaneNodeSetSpec) Default() {
 		spec.Nodes[nodeName] = *node.DeepCopy()
 	}
 
+	if spec.NodeTemplate.Ansible.AnsiblePort == 0 {
+		spec.NodeTemplate.Ansible.AnsiblePort = 22
```
slagle, line 65: why do we need to set the default port to 22? That's the ansible default already. We should just leave it unset.

PR #600, opened 2024-01-10
Generate certificates for ceph_hci_pre service

katarimanojk:
By using the feature [1] to generate certs, this PR will enable tlsCertsEnabled for ceph_hci_pre service which should generate certs for the service on each node in the nodeset.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/553

### DIFF
--- a/config/services/dataplane_v1beta1_openstackdataplaneservice_ceph_hci_pre.yaml
+++ b/config/services/dataplane_v1beta1_openstackdataplaneservice_ceph_hci_pre.yaml
@@ -5,3 +5,6 @@ metadata:
 spec:
   label: ceph-hci-pre
   playbook: osp.edpm.ceph_hci_pre
+  hasTLSCerts: True
+  issuers:
+    default: osp-rootca-issuer-internal


danpawlik:
recheck

config/services/dataplane_v1beta1_openstackdataplaneservice_ceph_hci_pre.yaml
```diff
@@ -5,3 +5,6 @@ metadata:
 spec:
   label: ceph-hci-pre
   playbook: osp.edpm.ceph_hci_pre
+  hasTLSCerts: True
+  issuers:
+    default: osp-rootca-issuer-internal
```
fmount, line 10: do we have evidence in CI of this part? Also, I'm wondering if the plan is to generate certificates per service or we can just reuse the existing certs created for the edpm nodes.

fultonj:
The `ceph_hci_pre` EDPM service prepares EDPM nodes to host Ceph. Right now that only consists of opening firewall ports though the space is there for it to do more.

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ceph_hci_pre/tasks/firewall.yml

My understanding from the following PR is that `hasTLSCerts: True` results in certs being generated and mounted to the AEE pod so that the playbook could copy those certs to EDPM nodes if needed.

https://github.com/openstack-k8s-operators/dataplane-operator/pull/553

I agree that we will want certs on EDPM nodes for any service which is deployed that needs those certs. I don't see what the `ceph_hci_pre` EDPM service would do with certificates however. It doesn't deploy a service listening on a port which needs to be encrypted, unless the intent is for Ceph to somehow use those certificates.

katarimanojk:
> I agree that we will want certs on EDPM nodes for any service which is deployed that needs those certs. I don't see what the >`ceph_hci_pre` EDPM service would do with certificates however. It doesn't deploy a service listening on a port which needs to >be encrypted, unless the intent is for Ceph to somehow use those certificates.
Yes the intent here is just to generate a certificate (via some ceph related dataplane service) and use it for ceph dashboard/RGW if needed.

fmount:
> The `ceph_hci_pre` EDPM service prepares EDPM nodes to host Ceph. Right now that only consists of opening firewall ports though the space is there for it to do more.
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ceph_hci_pre/tasks/firewall.yml
> 
> My understanding from the following PR is that `hasTLSCerts: True` results in certs being generated and mounted to the AEE pod so that the playbook could copy those certs to EDPM nodes if needed.
> 
> #553
> 
> I agree that we will want certs on EDPM nodes for any service which is deployed that needs those certs. I don't see what the `ceph_hci_pre` EDPM service would do with certificates however. It doesn't deploy a service listening on a port which needs to be encrypted, unless the intent is for Ceph to somehow use those certificates.

So (just to make sure I understand this part correctly) what the `hasTLSCerts: True`  does is basically creating crt/key and get them mounted to `AEE`: if you don't have the logic to use them those files are not synced anywhere in the `EDPM` nodes, right?
I agree that `ceph_hci_pre` is not the right place to manage this part, and I was hoping to use crt/key already present in the EDPM nodes (w/o  regenerating them explicitly), but I suspect it depends on the strategy.
If they're "per service" generated, then we might want to have a little playbook that put them in the right place (e.g. dataplane-operator generates them and once they get mounted to AEE we can invoke a ceph related playbook to sync them in the EDPM nodes).
Not sure we can simplify this model further: the `ci-framework` ceph playbook assumes both crt/key exist in a given path, but I'm wondering if we can simply request a new crt/key for the Ceph services.

config/services/dataplane_v1beta1_openstackdataplaneservice_ceph_hci_pre.yaml
vakwetu: The parameter name has changed.  The new name is tlsCertsEnabled  as seen in https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml  or https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_libvirt.yaml

vakwetu: Other than that.  This should be fine.

vakwetu:
There are  a couple of reviews currently in flight that will affect how all of this will be put together.
- https://github.com/openstack-k8s-operators/edpm-ansible/pull/525  (just merged) 
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/585

Once these land, there will be a new dataplane service ("install-certs") which will take any certs generated for a particular service (per node) and copy them to specific locations on that node.  Specifically,  
- cert at /var/lib/openstack/certs/<service>/tls.crt
- key at /var/lib/openstack/certs/<service>/tls.key
- cacert bundle at /var/lib/openstack/cacerts/<service>/tls-ca-bundle.pem

The <service> that is there corresponds to services which has set tlsCertsEnabled and the issuer as you've done in this patch.  At this point, we have certs generated for nova and libvirt.

These certs/keys/cacerts will be available for any of the edpm_* roles to use as they see fit.  In the nova role, for instance, they use kolla to mount the certs to the locations they want in the container.  In libvirt, they copy to /etc/pki/... directly on the host because that's where libvirt is running.  

Because the install-certs service runs first, nova also has access to and can mount libvirt certs as needed.

Ping me here on or on slack, and we can discuss what your requirements are and how to meet them.

fultonj:
> > I agree that we will want certs on EDPM nodes for any service which is deployed that needs those certs. I don't see what the >`ceph_hci_pre` EDPM service would do with certificates however. It doesn't deploy a service listening on a port which needs to >be encrypted, unless the intent is for Ceph to somehow use those certificates.
> 
> Yes the intent here is just to generate a certificate (via some ceph related dataplane service) and use it for ceph dashboard/RGW if needed.

I'm not sure that EDPM should be doing that though.

slagle:
> > > I agree that we will want certs on EDPM nodes for any service which is deployed that needs those certs. I don't see what the >`ceph_hci_pre` EDPM service would do with certificates however. It doesn't deploy a service listening on a port which needs to >be encrypted, unless the intent is for Ceph to somehow use those certificates.
> > 
> > 
> > Yes the intent here is just to generate a certificate (via some ceph related dataplane service) and use it for ceph dashboard/RGW if needed.
> 
> I'm not sure that EDPM should be doing that though.

Me neither. EDPM isn't installing ceph dashboard right? Shouldn't cephadm be used for that? Is that your concern as well?

fultonj:
> > > > I agree that we will want certs on EDPM nodes for any service which is deployed that needs those certs. I don't see what the >`ceph_hci_pre` EDPM service would do with certificates however. It doesn't deploy a service listening on a port which needs to >be encrypted, unless the intent is for Ceph to somehow use those certificates.
> > > 
> > > 
> > > Yes the intent here is just to generate a certificate (via some ceph related dataplane service) and use it for ceph dashboard/RGW if needed.
> > 
> > 
> > I'm not sure that EDPM should be doing that though.
> 
> Me neither. EDPM isn't installing ceph dashboard right? Shouldn't cephadm be used for that? Is that your concern as well?

yes, same concerns.

katarimanojk:
As per the discussions , as Ceph is not part of the deployment, EDPM is not supposed to generate certificates .

Ceph is deployed outside the product,  cifmw itself should generate it (may be self-signed) or use existing certs in EDPM node

This change can be abandoned.

PR #604, opened 2024-01-15
Removed KubeService creation

jlarriba:
The creation of Kubernetes Services that front EDPM nodes ports was intended to be used primarily to configure Prometheus scraping endpoints. To take in account EDPM deployments with node counts in potentially in the thousands, EndpointSlice Kubernetes object was used, which allows the definition of 100 endoints and, when it is full, it just create a new EndpointSlice, as many as are needed.

Sadly, the cluster-observability-operator that we will be using to provide and configure Prometheus does not yet support reading endpoints from EndpointSlices, only Endpoint, which is limited to 1000 endpoints and then it just stops scaling. After a conversation with the cluster-observability-operator developers, we end up ditching the usage of Services and instead make the telemetry-operator read directly the endpoints from multiple EDPM configuration sources using the same code used here and configure an ScrapeConfig with all of them.

This renders the Kubernetes Service useless for Telemetry purposes and, as we were the only known use case for this feature, I am then submitting this PR to remove it.


slagle:
I didn't hear any objections to removing this, so it's safe to proceed. I did fix the merge conflict by pushing to your private branch. When CI passes, it's good to go.

slagle (approved):

jlarriba:
recheck

PR #605, opened 2024-01-15
Add missing binaryData fields for ConfigMaps

cschwede:
ConfigMaps can use binaryData instead of data fields. In that case the OpenStackAnsibleEE spec failed due to missing keys. This patch iterates over these fields as well and adds the missing keys.


cschwede:
recheck

PR #608, opened 2024-01-16
Expose the enabled services to the ansible-runner

fao89:
edpm-ansible can be aware of the enabled services from nodeset inventory

### DIFF
--- a/pkg/deployment/inventory.go
+++ b/pkg/deployment/inventory.go
@@ -47,6 +47,8 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 		return "", err
 	}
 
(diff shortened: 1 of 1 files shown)


fao89:
it could help make download cache smarter, the issue is with servicesOverride

pkg/deployment/inventory.go
```diff
@@ -47,6 +47,8 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 		return "", err
 	}
 
+	nodeSetGroup.Vars["edpm_enabled_services"] = instance.Spec.Services
```
rabi, line 50: Services can be overridden in the deployment and this ansible var can have incorrect data, right?

fao89: yeah, I'll close it for now

PR #609, opened 2024-01-17
Add initial documentation for TLS-E

vakwetu:


fao89 (commented):
you need to add
```yaml
- TLS: tls.md
```
here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/mkdocs.yml#L27

fao89 (approved):

PR #614, opened 2024-01-22
Make the network names all lowercase

rabi:
Make the network names all lowercase (not snakecase) to align them with the network names used in ctlplane services.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1050


bshephar (approved):
I like it, but this will probably have implications for people's test envs and CI environments. So we'll probably need to communicate this change with folks.

rabi:
recheck

rabi:
recheck

rabi:
recheck kernel panic bm job

PR #616, opened 2024-01-23
Add nic-config samples folder

bshephar:
This change copies some of the examples from EDPM Ansible into the dataplane-operator samples. This allows users to better reference the examples that can be in-lined in the NodeSet without having to jump back and forward between edpm-ansible and dataplane-operator. This should reduce confusion and improve user experience while trying to find examples for what they are trying to do.

### DIFF
--- a/docs/common_configurations.md
+++ b/docs/common_configurations.md
@@ -90,17 +90,16 @@ The
 [edpm_network_config](https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_network_config)
 ansible role is responsible for configuring networking on dataplane nodes.
 
-The `edpm_network_config_template` variable specifies the path to an ansible
+The `edpm_network_config_template` variable specifies the contents of a jinja2
 template that describes the networking configuration to be applied. The
 template itself also contains variables that can be used to customize the
 networking configuration for a specific node (IP addresses, interface names,
-routes, etc). Templates provided with the edpm_network_config role are at
-<https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_network_config/templates>.
+routes, etc). See template examples provided in the nic-config-samples directory:
+<https://github.com/openstack-k8s-operators/dataplane-operator/config/samples/nic-config-samples>.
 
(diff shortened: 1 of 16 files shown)


docs/common_configurations.md
```diff
@@ -94,13 +94,8 @@ The `edpm_network_config_template` variable specifies the path to an ansible
 template that describes the networking configuration to be applied. The
```
rabi, line 94: We need to change the text as it's not a path to the template any more. We should also mention here that the samples should be used inline in the CRs.

bshephar:
> I don't totally think this is needed, as they need to be maintained in 2 places now. Unless you intend to remove them from edpm-ansible as well?
> 
> Adding them here in dataplane-operator, doesn't make them any easier to consume for the product users. If this is just for developers, I don't see why we can't just have them one place in edpm-ansible.

Plan was to remove them from edpm-ansible once they were here. Since they're only being used in-line within the nodeset CR's, I thought it might be less confusing for people to find the examples here, rather than having to go to a different repo. Seemed like less mental hurdles for people to jump.

docs/common_configurations.md
bshephar: Done

rabi:
> Adding them here in dataplane-operator, doesn't make them any easier to consume for the product users.

Isn't it the same for CR samples? I agree that it's not easy for end users to consume them unless we add these samples in the docs as an appendix. The reason I suggested to move them here is because they're used inline in the CR samples and documented in dataplane-operator.

rabi:
+2

slagle:
> > Adding them here in dataplane-operator, doesn't make them any easier to consume for the product users.
> 
> Isn't it the same for CR samples? I agree that it's not easy for end users to consume them unless we add these samples in the docs as an appendix. The reason I suggested to move them here is because they're used inline in the CR samples and documented in dataplane-operator.

Yes, it's the same for CR  samples, that's why I didn't understand the point of moving the nic samples here. I'm ok with it, although I'd like to see them in only one place, so I think removing them from edpm-ansible would be good to do.

slagle (approved):

PR #617, opened 2024-01-24
Change the network names to lower case in samples

rabi:
This aligns with the network names used in control plane network attachments and  metallb IPAddressPool etc.


rabi:
recheck

rabi:
recheck crc could not start, no logs

PR #619, opened 2024-01-25
Update ctlplane_cidr variable to ctlplane_subnet_cidr

bshephar:
This change updates the ctlplane_cidr variable to be ctlplane_subnet_cidr in line with the changes to variable names made: https://github.com/openstack-k8s-operators/dataplane-operator/pull/614


rabi:
We don't need this. That change was in https://github.com/openstack-k8s-operators/dataplane-operator/pull/614 was intentional.

PR #621, opened 2024-01-25
All sample files can use IPAM

bshephar:
Using IPAM reduces the complexity of required Ansible variables and minimises confusion.


config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
```diff
@@ -25,15 +25,19 @@ spec:
...
-            storage_ip: 172.18.0.100
-            tenant_ip: 172.19.0.100
-            fqdn_internalapi: edpm-compute-0.example.com
   networkAttachments:
     - ctlplane
   nodeTemplate:
+    networks:
+      - name: CtlPlane
+        subnetName: subnet1
+        defaultRoute: true
```
rabi, line 34: You need to add the `fixedIP` here and other files changd here (same what's used for `ansibleHost`) as this is pre-provisioned. Else we would have issues with reserved ip vs `ansibleHost` var. Also, we should make the same change in https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml (which is bug I mentioned earlier) and drop https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml

rabi (changes requested):

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
bshephar: Ahh obviously, thanks for catching that.

ack on the second point. That sounds good.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
```diff
@@ -26,15 +26,20 @@ spec:
...
-            ctlplane_ip: 192.168.122.100
-            internalapi_ip: 172.17.0.100
-            storage_ip: 172.18.0.100
-            tenant_ip: 172.19.0.100
-            fqdn_internalapi: edpm-compute-0.example.com
   networkAttachments:
     - ctlplane
   nodeTemplate:
+    networks:
+      - name: CtlPlane
```
rabi, line 33: Let's change these network names to lowercase to align with what's in `networkAttachments`.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph.yaml
```diff
@@ -62,34 +66,11 @@ spec:
...
-        - tenant
-        storage_cidr: "24"
-        storage_host_routes: []
-        storage_vlan_id: 21
-        storage_mtu: 9000
-        tenant_cidr: "24"
-        tenant_host_routes: []
-        tenant_mtu: 1500
-        tenant_vlan_id: 22
+        service_net_map:
```
rabi, line 71: Why adding these again, not required? Probably rebase issue.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
bshephar: ack, true. Updated them all now

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
rabi: Sorry, I missed this earlier. `networks` in `nodeTemplate` is meant for situations where you don't want predictable IP at all (even for ctlplane). So using fixedIP here is incorrect. The correct thing would be to add it on the node section.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
```diff
@@ -25,15 +25,20 @@ spec:
...
-            tenant_ip: 172.19.0.100
-            fqdn_internalapi: edpm-compute-0.example.com
   networkAttachments:
     - ctlplane
   nodeTemplate:
+    networks:
+      - name: ctlplane
+        subnetName: subnet1
+        defaultRoute: true
+        fixedIP: 192.168.122.100
```
rabi, line 35: same here.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_ceph.yaml
```diff
@@ -7,18 +7,22 @@ spec:
   networkAttachments:
   - ctlplane
   nodeTemplate:
+    networks:
```
rabi, line 10: This is not required as it's using IPAM, see the `nodes` section. Only drop the ansible vars as you've done below.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_customnetworks.yaml
```diff
@@ -24,17 +24,20 @@ spec:
...
-          internalapi_ip: 172.17.0.101
-          storage_ip: 172.18.0.101
-          tenant_ip: 172.19.0.101
-          external_ip: 172.20.12.76
-          fqdn_internalapi: edpm-compute-1.example.com
-          ansible_ssh_transfer_method: scp
   networkAttachments:
     - ctlplane
   nodeTemplate:
+    networks:
```
rabi, line 30: This also has to go to `nodes` section. Also it seems `preProvisioned: true` is missing in this one.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
```diff
@@ -20,15 +20,20 @@ spec:
...
-          ansibleVars:
-            ctlplane_ip: 192.168.122.100
-            internalapi_ip: 172.17.0.100
-            storage_ip: 172.18.0.100
-            tenant_ip: 172.19.0.100
-            fqdn_internalapi: edpm-networker-0.example.com
   networkAttachments:
     - ctlplane
   nodeTemplate:
+    networks:
```
rabi, line 26: same here move to 'nodes` section as you can't have `fixedIP` in a template for all nodes.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_nmstate.yaml
```diff
@@ -25,15 +25,20 @@ spec:
...
-        ansibleVars:
-          ctlplane_ip: 192.168.122.100
-          internalapi_ip: 172.17.0.100
-          storage_ip: 172.18.0.100
-          tenant_ip: 172.19.0.100
-          fqdn_internalapi: edpm-compute-0.example.com
   networkAttachments:
   - ctlplane
   nodeTemplate:
+    networks:
```
rabi, line 31: same here.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_sriov.yaml
```diff
@@ -22,15 +22,20 @@ spec:
...
-          ansibleVars:
-            ctlplane_ip: 192.168.122.100
-            internalapi_ip: 172.17.0.100
-            storage_ip: 172.18.0.100
-            tenant_ip: 172.19.0.100
-            fqdn_internalapi: edpm-compute-0.example.com
   networkAttachments:
     - ctlplane
   nodeTemplate:
+    networks:
```
rabi, line 28: same.

rabi (changes requested):
Sorry I should have looked at it properly earlier. `nodeTemplate` section applies to all nodes, and you can't use it in samples with `preProvisioned: true` , where you've to provide fixedIP for ctlplane (that's different for each node).

config/samples/dataplane_v1beta1_openstackdataplanenodeset_customnetworks.yaml
bshephar: As long as I remove the `fixedIP` from the `nodeTemplate` section, this would still be valid though right?

And I added `preProvisioned: true` in the latest iteration as well.

rabi: No it won't work as with `preProvisioned: true` the IPAM provided `fixedIP` can be different from what you use in `ansibleHost` and aee can't reach edpm nodes using `ansibleHost`. For `preProvisoned: true` you can't have networks in `nodeTemplate` section.

bshephar:
recheck

config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
```diff
@@ -22,16 +22,21 @@ spec:
...
-            ctlplane_ip: 192.168.122.100
-            internalapi_ip: 172.17.0.100
-            storage_ip: 172.18.0.100
-            tenant_ip: 172.19.0.100
-            fqdn_internalapi: edpm-compute-0.example.com
+    edpm-compute-0:
+      hostName: edpm-compute-0
+      ansible:
+        ansibleHost: 192.168.122.100
+    networks:
```
rabi, line 29: Wrong indent it seems. `networks` should be a child of `edpm-compute-0`. Same in other samples.

bshephar: Ah true, and I made the same mistake in all of them because I just did 'y' 'p' everywhere.

Fixed them all up now with the extra indent level.

config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
```diff
@@ -26,10 +26,6 @@ spec:
   nodes:
     edpm-compute-0:
       hostName: edpm-compute-0
-  networkAttachments:
-    - ctlplane
-  nodeTemplate:
-    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
     networks:
```
rabi, line 29: Hmm.. we should not move this from nodeTemplate section. Though it works, you don't have to repeat it for every node when there are no `fixedIP`s.

rabi (approved):
Anyway, let's merge this and revert the commented change later.

PR #622, opened 2024-01-25
Remove variables that just set default values

bshephar:
This changes removes any variables from samples that just set default values


lewisdenny:
recheck

slagle (commented):
what about the other samples that also set some of these default values? such as config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml. shouldn't that be cleaned up as well?

rabi:
This has the same kind of changes as in https://github.com/openstack-k8s-operators/dataplane-operator/pull/621, so let's wait for that to merge and the rebase this.

slagle:
@bshephar Can we close this one out?

PR #624, opened 2024-01-25
Adding option to deploy service across nodesets

jpodivin:
This weakly depends on patch in edpm-ansible. Without it the variable won't be used for anything but shouldn't cause any damage.

Depends-on: https://github.com/openstack-k8s-operators/edpm-ansible/pull/548


api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -85,6 +85,10 @@ type OpenStackDataPlaneServiceSpec struct {
 	// CACerts - Secret containing the CA certificate chain
 	// +kubebuilder:validation:Optional
 	CACerts string `json:"caCerts,omitempty" yaml:"caCerts,omitempty"`
+
+	// TransNodeset - should the service be deploy across all nodesets
+	// +kubebuilder:validation:optional
+	TransNodeset bool `json:"transNodeset,omitempty" yaml:"transNodeset,omitempty"`
```
slagle, line 91: I'd prefer a different name, that is also more descriptive. Consider:

AllNodeSets
DeployAllNodeSets
DeployAcrossAllNodeSets

slagle: Should we also make the type `*bool` so that it matches TLSCertsEnabled? Although, we haven't been consistent with that in other Spec structs. Perhaps we should be though.

slagle (changes requested):

api/v1beta1/openstackdataplaneservice_types.go
jpodivin: I don't know about the type. I'm personally more on the side of avoiding pointers when it comes to things like `bool`.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -85,6 +85,18 @@ type OpenStackDataPlaneServiceSpec struct {
...
+	// This will override default target of a service play, setting it to 'all'.
+	// If both DeoployOnAllNodesets and PlayTargetOverride are set, default to 'all'.
+	// +kubebuilder:validation:Optional
+	DeployOnAllNodesets bool `json:"deployOnAllNodesets,omitempty" yaml:"deployOnAllNodesets,omitempty"`
+
+	// PlayTargetOverride - alternative target for plays
+	// Target host of the play will be overriden with provided string.
+	// If both DeoployOnAllNodesets and PlayTargetOverride are set, default to 'all'.
+	// +kubebuilder:validation:Optional
+	PlayTargetOverride string `json:"playTargetOverride,omitempty" yaml:"playTargetOverride,omitempty"`
```
slagle, line 99: I don't see why this is needed. Why add a field we'd need to support when we don't really have a use case for it?

jpodivin: Playbooks we have in edpm-ansible now are targeting all hosts by default. We are going to allow overrides with the https://github.com/openstack-k8s-operators/edpm-ansible/pull/548 . But the only override we can provide would be, again `all`, to deploy everywhere.

That is fine in the narrow sense of the epic. We could stick with that and call it a day. But what if we wanted to deploy some service on nodes that were not in the original playbook target, but not on `all` either?

To me it seems like relatively likely scenario, and since we are already implementing most of this anyway with the `all` override, I figured we may want to anticipate the need.

api/v1beta1/openstackdataplaneservice_types.go
jpodivin: From what I can tell the ratio between bool and *bool is roughly 50/50. In my opinion the `bool` is better, because you don't have to handle the reference, and since bool is 1 byte it's actually smaller than a pointer would be.

jpodivin:
recheck - `error: You must be logged in to the server (Unauthorized)`

api/v1beta1/openstackdataplaneservice_types.go
jpodivin: Type changed to `*bool` after consultation.

jpodivin:
recheck - `ERROR! Error when getting collection version metadata for community.general:3.0.0 from default (https://galaxy.ansible.com/api/) (HTTP Code: 504, Message: Gateway Time-out Code: Unknown)`

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -102,6 +102,11 @@ type OpenStackDataPlaneServiceSpec struct {
 	// +kubebuilder:default=false
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
 	AddCertMounts bool `json:"addCertMounts" yaml:"addCertMounts"`
+
+	// DeployOnAllNodesets - should the service be deploy across all nodesets
+	// This will override default target of a service play, setting it to 'all'.
+	// +kubebuilder:validation:Optional
+	DeployOnAllNodesets *bool `json:"deployOnAllNodesets,omitempty" yaml:"deployOnAllNodesets,omitempty"`
```
slagle, line 109: this needs to be `DeployOnAllNodeSets` (NodeSet with capital S) for consistency with how we use NodeSets everywhere else.

slagle (commented):
This is still just the start of the implementation right? It still needs to mount all the inventories for all the NodeSets in the deployment, and execute the service for all NodeSets just a single time. Do you plan to add further commits to this PR?

api/v1beta1/openstackdataplaneservice_types.go
jpodivin: Thanks. done.

jpodivin:
recheck

slagle (commented):
This is looking good, but I don't think we need to loop separately for the global services, and we don't need the extra field, AllNodeSetsServices. You can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.

When that's true, the Deployer then needs to set the var for the play hosts, and also set a var (via ansibleEE.Spec.ExtraVars) for the ssh key for each NodeSet.

jpodivin:
> u can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.

You are right. Except in this specific case we only want to deploy global (all nodeset) service, after we are done with everything else. Really the name of var `AllNodeSetsServices` is very misleading, I guess the `PostDeployAllNodeSetsServices` would be more appropriate.

Now if we wanted to deploy something everywhere, without having to wait for normal services, then it would make sense to do it  in one loop. Anyway, I'll factor out the inventory gathering. One extra loop isn't going to cost us much.

slagle:
> > u can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.
> 
> You are right. Except in this specific case we only want to deploy global (all nodeset) service, after we are done with everything else. Really the name of var `AllNodeSetsServices` is very misleading, I guess the `PostDeployAllNodeSetsServices` would be more appropriate.
> 
> Now if we wanted to deploy something everywhere, without having to wait for normal services, then it would make sense to do it in one loop. 

This ^^ *is* what we want to do. Let's save the complexity around ordering and dependencies between the services for when we actually want to solve for that. The specific use case (configure host key identity everywhere) does not depend on any of that.

It makes sense to start with the simplest design first, instead of a design we'd need to change later as we identify other use cases.

jpodivin:
> > > u can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.
> > 
> > 
> > You are right. Except in this specific case we only want to deploy global (all nodeset) service, after we are done with everything else. Really the name of var `AllNodeSetsServices` is very misleading, I guess the `PostDeployAllNodeSetsServices` would be more appropriate.
> > Now if we wanted to deploy something everywhere, without having to wait for normal services, then it would make sense to do it in one loop.
> 
> This ^^ _is_ what we want to do. Let's save the complexity around ordering and dependencies between the services for when we actually want to solve for that. The specific use case (configure host key identity everywhere) does not depend on any of that.
> 
> It makes sense to start with the simplest design first, instead of a design we'd need to change later as we identify other use cases.

In that case I'm going to rewrite this.

jpodivin:
> > > u can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.
> > 
> > 
> > You are right. Except in this specific case we only want to deploy global (all nodeset) service, after we are done with everything else. Really the name of var `AllNodeSetsServices` is very misleading, I guess the `PostDeployAllNodeSetsServices` would be more appropriate.
> > Now if we wanted to deploy something everywhere, without having to wait for normal services, then it would make sense to do it in one loop.
> 
> This ^^ _is_ what we want to do. Let's save the complexity around ordering and dependencies between the services for when we actually want to solve for that. The specific use case (configure host key identity everywhere) does not depend on any of that.
> 
> It makes sense to start with the simplest design first, instead of a design we'd need to change later as we identify other use cases.

I suppose the PR could be rewritten to work like that. But almost, if not all, of our playbooks have hosts default to `all` unless overridden, which is not something we can do right now. This means that if we supply ansibleee with all inventories, it will try to run them everywhere, turning every service into 'global' service.

slagle:
> > > > u can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.
> > > 
> > > 
> > > You are right. Except in this specific case we only want to deploy global (all nodeset) service, after we are done with everything else. Really the name of var `AllNodeSetsServices` is very misleading, I guess the `PostDeployAllNodeSetsServices` would be more appropriate.
> > > Now if we wanted to deploy something everywhere, without having to wait for normal services, then it would make sense to do it in one loop.
> > 
> > 
> > This ^^ _is_ what we want to do. Let's save the complexity around ordering and dependencies between the services for when we actually want to solve for that. The specific use case (configure host key identity everywhere) does not depend on any of that.
> > It makes sense to start with the simplest design first, instead of a design we'd need to change later as we identify other use cases.
> 
> I suppose the PR could be rewritten to work like that. But almost, if not all, of our playbooks have hosts default to `all` unless overridden, which is not something we can do right now. This means that if we supply ansibleee with all inventories, it will try to run them everywhere, turning every service into 'global' service.

This really isn't that complicated. Just set the var for the play hosts to the nodeset name when it's not a global service or "all" when it is a global service.

jpodivin:
> > > > > u can just pass all secrets to the Deployer every time, and pass DeployAcrossAllNodeSets to the Deployer. You've already added a field to the Deployer struct, so i would just keep that named the same as DeployAcrossAllNodeSets.
> > > > 
> > > > 
> > > > You are right. Except in this specific case we only want to deploy global (all nodeset) service, after we are done with everything else. Really the name of var `AllNodeSetsServices` is very misleading, I guess the `PostDeployAllNodeSetsServices` would be more appropriate.
> > > > Now if we wanted to deploy something everywhere, without having to wait for normal services, then it would make sense to do it in one loop.
> > > 
> > > 
> > > This ^^ _is_ what we want to do. Let's save the complexity around ordering and dependencies between the services for when we actually want to solve for that. The specific use case (configure host key identity everywhere) does not depend on any of that.
> > > It makes sense to start with the simplest design first, instead of a design we'd need to change later as we identify other use cases.
> > 
> > 
> > I suppose the PR could be rewritten to work like that. But almost, if not all, of our playbooks have hosts default to `all` unless overridden, which is not something we can do right now. This means that if we supply ansibleee with all inventories, it will try to run them everywhere, turning every service into 'global' service.
> 
> This really isn't that complicated. Just set the var for the play hosts to the nodeset name when it's not a global service or "all" when it is a global service.

I believe this PR had variable override for the target two weeks ago, although derived from service field. If you want I can include it and fill it with nodeset name.

jpodivin:
recheck

controllers/openstackdataplanedeployment_controller.go
```diff
@@ -195,13 +195,21 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
...
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.ReadyMessage)
+	shouldRequeue := false
+	haveError := false
+
+	globalInventorySecrets := []string{}
+
+	// Gathering individual inventory secrets for later use
+	for _, nodeSet := range nodeSets.Items {
+		// Add inventory secret to list of inventories for global services
+		globalInventorySecrets = append(globalInventorySecrets, fmt.Sprintf("dataplanenodeset-%s", nodeSet.Name))
```
slagle, line 206: not related to your PR, but as a follow up, we should likely factor out the fmt.SPrintf logic into a GetNodeSetInventoryName func in nodeset_types.go, as this same logic is also in inventory.go.

pkg/util/ansible_execution.go
```diff
@@ -99,7 +105,20 @@ func AnsibleExecution(
 			ansibleEE.Spec.Playbook = service.Spec.Playbook
 		}
 
-		ansibleEEMounts := storage.VolMounts{}
+		// If we have a service that ought to be deployed everywhere
+		// substitute the existing play target with 'all'
+		if service.Spec.DeployOnAllNodeSets != nil && *service.Spec.DeployOnAllNodeSets {
+			ansibleEE.Spec.ExtraVars = map[string]json.RawMessage{
+				"edpm_override_hosts": json.RawMessage([]byte("\"all\"")),
+			}
```
slagle, line 113: shouldn't this add to ansibleEE.Spec.ExtraVars and not overwrite the whole value?

pkg/util/ansible_execution.go
```diff
@@ -99,7 +105,20 @@ func AnsibleExecution(
...
+			util.LogForObject(helper, fmt.Sprintf("for service %s, substituting existing ansible play host with 'all'.", service.Name), ansibleEE)
+		} else {
+			ansibleEE.Spec.ExtraVars = map[string]json.RawMessage{
```
slagle, line 116: ditto

controllers/openstackdataplanedeployment_controller.go
slagle: also, I don't know that globalInventorySecrets is the right var name here, as it's not global for all NodeSets, just the NodeSets in the Deployment

pkg/util/ansible_execution.go
```diff
@@ -42,11 +43,16 @@ func AnsibleExecution(
 	obj client.Object,
 	service *dataplanev1.OpenStackDataPlaneService,
 	sshKeySecret string,
```
slagle, line 45: sshKeySecret needs to be updated to handled the same way as inventorySecrets. As each NodeSet could have a different sshKeySecret, we need to pass those in as a list here, and then mount each sshKeySecret into the pod.

you'll likely have to set an ansible group var somehow to make sure each nodeset uses the right SSH key.

pkg/util/ansible_execution.go
```diff
@@ -119,32 +138,44 @@ func AnsibleExecution(
 			MountPath: "/runner/env/ssh_key",
```
slagle, line 138: each key will need to be mounted at a unique path (based on the secret name). They will likely need to be mounted under the runner project dir, since /runner/env/ssh_key only has support for a single key (https://github.com/ansible/ansible-runner/issues/51). So something like `/runner/env/project/ssh_keys/<secret_name>`

Then when we generate the NodeSet inventory in inventory.go, I'd go ahead and set a group var there for ansible_ssh_private_key_file set to that same path.

slagle (commented):
Thanks, this is looking better. I think the last missing chunk of functionality is the handling for different SSH keys for each NodeSet.

controllers/openstackdataplanedeployment_controller.go
jpodivin: I don't know if naming it `deploymentNodesetsInventorySecrets` would be much better. These are as global as it can get within deployment.

pkg/util/ansible_execution.go
jpodivin: We are only passing one ssh key ansible runner though. It's specific to a node we are trying to reach. So this will have to be changed in deployment reconciliation first.

slagle: you'll have to build a list of the ssh key secrets just like you build a list of the inventory secrets, and pass them in to the ansible execution.

ansible-runner won't use all the keys automatically, that's why I said you'll need to set a group var in the inventory with a path to the key. ansible-runner doesn't need to have any before hand knowledge of the path, it's just a variable set to the ssh key path.

jpodivin:
recheck

controllers/openstackdataplanedeployment_controller.go
```diff
@@ -226,13 +234,14 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
...
-			AeeSpec:         &ansibleEESpec,
-			InventorySecret: nodeSetSecretInv,
+			Ctx:                         ctx,
+			Helper:                      helper,
+			NodeSet:                     &nodeSet,
+			Deployment:                  instance,
+			Status:                      &instance.Status,
+			AeeSpec:                     &ansibleEESpec,
+			InventorySecrets:            globalInventorySecrets,
+			AnsibleSSHPrivateKeySecrets: globalSshKeySecrets,
```
slagle, line 244: can you use consistent camelCase? it should be SSH everywhere.

pkg/deployment/inventory.go
```diff
@@ -51,6 +51,8 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 		nodeSetGroup.Vars["edpm_tls_certs_enabled"] = "true"
 	}
 
+	nodeSetGroup.Vars["ansible_ssh_private_key_file"] = fmt.Sprintf("/runner/env/ssh_key/ssh_key_%s", instance.Name)
```
slagle, line 54: could we have this fmt.Sprintf in a func in nodeset_types, since it needs to be done twice

pkg/util/ansible_execution.go
```diff
@@ -99,51 +109,83 @@ func AnsibleExecution(
...
+			util.LogForObject(helper, fmt.Sprintf("for service %s, substituting existing ansible play host with 'all'.", service.Name), ansibleEE)
+		} else {
+			ansibleEE.Spec.ExtraVars["edpm_override_hosts"] = json.RawMessage([]byte(fmt.Sprintf("\"%s\"", targetNodeset)))
+			util.LogForObject(helper, fmt.Sprintf("for service %s, substituting existing ansible play host with 'all'.", service.Name), ansibleEE)
+		}
+
+		for sshKeyNodeName, sshKeySecret := range sshKeySecrets {
+			if service.Spec.DeployOnAllNodeSets != nil && *service.Spec.DeployOnAllNodeSets {
+				sshKeyName = fmt.Sprintf("ssh-key-%s", sshKeyNodeName)
+				sshKeyMountSubPath = fmt.Sprintf("/ssh_key_%s", targetNodeset)
```
slagle, line 125: is the leading `/` needed here?

jpodivin: no it isn't.

controllers/openstackdataplanedeployment_controller.go
jpodivin: yes.

fao89:
recheck

(last error didn't reach dataplane code)

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
```diff
@@ -0,0 +1,174 @@
...
+  - neutron-metadata
+  - neutron-ovn
+  - neutron-sriov
+  - neutron-dhcp
+  - libvirt
+  - nova
+  env:
+  - name: ANSIBLE_FORCE_COLOR
+    value: "True"
+  nodes:
```
slagle, line 87: You can't actually have nodes defined here. There are no nodes in kuttl tests, the ansible jobs fail with SSH connection errors.

tests/kuttl/tests/dataplane-deploy-global-service-test/01-dataplane-deploy.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: custom-global-service
```
slagle, line 5: this service isn't actually added anywhere, so the new code isn't tested. Did you mean to add this to servicesOverride on the Deployment?

tests/kuttl/tests/dataplane-deploy-global-service-test/00-assert.yaml
```diff
@@ -0,0 +1,113 @@
...
+  - configure-os
+  - run-os
+  - install-certs
+  - ovn
+  - neutron-metadata
+  - neutron-ovn
+  - neutron-sriov
+  - neutron-dhcp
+  - libvirt
+  - nova
```
slagle, line 23: you're missing `custom-global-service` here

slagle: I went ahead and added it and pushed to your branch.

jpodivin: Squashed in original, thank you James.

pkg/util/ansible_execution.go
```diff
@@ -98,51 +108,87 @@ func AnsibleExecution(
...
-							Path: "ssh_key",
+		// If we have a service that ought to be deployed everywhere
+		// substitute the existing play target with 'all'
+		// Check if we have ExtraVars before accessing it
+		if ansibleEE.Spec.ExtraVars == nil {
+			ansibleEE.Spec.ExtraVars = make(map[string]json.RawMessage)
+		}
+		if service.Spec.DeployOnAllNodeSets != nil && *service.Spec.DeployOnAllNodeSets {
+			ansibleEE.Spec.ExtraVars["edpm_override_hosts"] = json.RawMessage([]byte("\"all\""))
+			util.LogForObject(helper, fmt.Sprintf("for service %s, substituting existing ansible play host with 'all'.", service.Name), ansibleEE)
```
gibizer, line 119: so if we have two nodesets and a global service then what will happen?
* A) run the ansible playbook of the service once with hosts=all
* B) run the ansible playbook of the service twice both with hosts=all

what will happen if both of the two nodesets have the global service added?
what will happen if only one of the two nodesets have the global service added?

tests/functional/openstackdataplanenodeset_controller_test.go
```diff
@@ -211,6 +211,140 @@ var _ = Describe("Dataplane NodeSet Test", func() {
 			})
 		})
 
+		When("A Dataplane resorce is created with PreProvisioned nodes, no deployment and global service", func() {
```
gibizer, line 214: nit:resource

tests/functional/openstackdataplanenodeset_controller_test.go
```diff
@@ -211,6 +211,140 @@ var _ = Describe("Dataplane NodeSet Test", func() {
...
+					ConditionGetterFunc(DataplaneConditionGetter),
+					dataplanev1.SetupReadyCondition,
+					corev1.ConditionFalse,
+				)
+			})
+
+			It("Should not have created a Secret", func() {
+				th.AssertSecretDoesNotExist(dataplaneSecretName)
+			})
+		})
```
gibizer, line 346: could we have test that proves that AEE resource is reacted for the global service with the proper config?

pkg/util/ansible_execution.go
jpodivin: Good point. Services don't have any concept of dependencies, yet, so B is our answer.

That isn't necessarily catastrophic, because our services use ansible to make changes. If the roles are written properly, utilizing idempotent operations, or at least operations that don't cause damage if run twice, we should be fine.

Furthermore, just like with standard services, we rely on user to make the right call when setting up nodesets. Global services add a new aspect to this. But it's not that different from the current situation, when we expect users not to include one service twice in a nodeset, or to run services which are mutually incompatible.

That being said. I'll add documentation about this.

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
```diff
@@ -0,0 +1,169 @@
...
+      gather_facts: no
+      name: global kuttl play
+      tasks:
+        - name: Sleep
+          command: sleep 1
+          delegate_to: localhost
+  deployOnAllNodeSets: true
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNodeSet
```
slagle, line 79: could you add a 2nd NodeSet here and then actually assert that both inventories and secrets are mounted into the custom-global-service-edpm-compute-global-service OpenStackAnsibleEE?

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
```diff
@@ -0,0 +1,169 @@
...
+  - run-os
+  - install-certs
+  - ovn
+  - neutron-metadata
+  - neutron-ovn
+  - neutron-sriov
+  - neutron-dhcp
+  - libvirt
+  - nova
+  - custom-global-service
```
slagle, line 100: since we already have a different kuttl test that uses the whole services list, you could reduce this list to just custom-global-service and one other non global service, and I think the coverage would be the same. It would make the test run faster.

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
slagle: could be done in a follow up

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
slagle: could be done in a follow up

pkg/util/ansible_execution.go
slagle: I guess the concern here is about conflicts if there were 2 global executions simultaneously. It would be 2 different ansible processes trying to configure known_hosts at the same time, which could cause an issue. We could document this as a limitation, and add a check to make sure it doesn't occur, as we are already iterating over all the services, we'd need to check that each global service is only specified once. If we wanted to go beyond a check, we could just make sure the service is only execute one time, by making sure it only runs once by modifying the list of services we sent to the deploy function.

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
jpodivin: I think the speedup would be marginal. And furthermore, given the fact that global services have special approach to secrets, including inventories, I would rather keep normal services for testing as well.

tests/functional/openstackdataplanenodeset_controller_test.go
jpodivin: I'll put that among deployment functional tests. But I'm going to make an assertion here on the service itself.

pkg/util/ansible_execution.go
jpodivin: We can check that the global service is specified only once. But I don't think we should resolve wrong specification for the user by modifying the list. There should either be an error, preventing deployment completely, or a warning.

gibizer: Yeah the main concern is about simultaneously executing both instance of the global service against the same node. Do we deploy nodesets in a deployment sequentially or in parallel? If the latter, then we need to do some logic in the operator for mutual exclusiveness of the two run of the same service at the same time, if the former then doc is enough.

gibizer: Btw, what will be our doc strategy here?
* ask the user to add the global service to each nodeset in the deployment
* ask the user to add the global service to one of the nodesets and ensure that if that nodeset is removed later then the global service is added to another nodeset in the deployment

gibizer:
@jpodivin do you plan to create separate PRs to move the edpm_ssh_known_hosts role to its own playbook and then into its own, global service?

controllers/openstackdataplanedeployment_controller.go
```diff
@@ -213,13 +213,23 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
...
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.ReadyMessage)
+	shouldRequeue := false
+	haveError := false
+
+	globalInventorySecrets := []string{}
+	globalSSHKeySecrets := map[string]string{}
+
+	// Gathering individual inventory and ssh secrets for later use
+	for _, nodeSet := range nodeSets.Items {
+		// Add inventory secret to list of inventories for global services
```
rabi, line 224: As we're doing this in every deployment reconciliation, can we make this conditional by checking for nodesets to be deployed or service override in the deployment, containing a service with `deployOnAllNodeSets: true`?

rabi: Also I was thinking should we make global services deployed only part of service override, as them being in nodeset spec when deployed across nodesets misleading. That brings back the discussion whether service list should even be part of nodeset spec.

jpodivin: If we want to move services outside of nodeset spec, I think this goes beyond objective of this feature.
Keeping global services in separate list, as we do with Overrides, was part of earlier iteration of this feature.

jpodivin:
> @jpodivin do you plan to create separate PRs to move the edpm_ssh_known_hosts role to its own playbook and then into its own, global service?

Not really no.

pkg/util/ansible_execution.go
jpodivin: Have global service in one deployment. It's going to get executed against all.

gibizer: but the service is in nodeset not in deployment.

controllers/openstackdataplanedeployment_controller.go
rabi: > If we want to move services outside of nodeset spec....

Neither I'm not asking to move the services out of nodeset spec in the PR (however, that's a design discussion we should have without blocking this PR), nor I want the earlier iteration of having global service list separately in nodeset spec.  The point I was making is, with service list in deployments, one can create a deployment with the whatever services to be deployed across all nodesets (without any nodesets in deployment spec or providing something like `all` ) and we deploy the services across all nodesets by collecting all nodeset inventories.

However, we should not loop over all nodesets to create list of inventories in all reconciliations as done in this PR unnecessarily. We should check if we've anything to be deployed across all nodesets before doing that.

jpodivin: ServicesOverride is part of deployment spec. So effectively, your proposal would mean implementing another list in deployment spec, with list of global services. Which is what I'm talking about.

pkg/deployment/inventory.go
jpodivin: Sorry, I've just noticed this comment. Where is the other one?

controllers/openstackdataplanedeployment_controller.go
rabi: > your proposal would mean implementing another list in deployment spec,

No that's not what I'm saying, if you read my comments properly. I'm suggesting we just use ServiceOverrides for deploy all services to start with and then move service list to deployment  spec and remove ServiceOverride.

gibizer: @rabi that would prevent to pass different config maps to a service per nodeset, e.g. nova cell config

rabi: @gibizer If I understand correctly, you're saying we need different set of services per nodeset, as it's a different service (instance) with a different service config map. Now  nodesets in a deployment are deployed in parallel, so one can create multiple deployments, one each per cell, no? Anyway, I find the current design quite confusing where we say nodeset should have service list of services to be deployed and also have services to be deployed in all nodesets. Then we deploy all nodeset deployments multiple times in a deployment (as we do it per nodeset). Keep aside the fact that what's to be deployed in a nodeset can be overridden in a deployment.

tests/functional/openstackdataplanenodeset_controller_test.go
jpodivin: Test is now included in dataplanedeploymentcontroller suite.

slagle (commented):
+2

I see the earlier comment from @rabi to optimize the looping. Is that something we could do in a follow up in the interest of getting this merged since we have green CI now?

controllers/openstackdataplanedeployment_controller.go
slagle: Could we collect the feedback about the design, with a proposal of how we'd implement that along with the impact? Part of the reason why I push back on that scope of change is that it is quite disruptive. If there was a holistic plan on how we'd implement it and mitigate the impact, perhaps it's something we could execute on.

I tend to think the NodeSet services list as the defaults for the NodeSet, which can be customized at Deployment time. I'd be in favor of building out the customization interface on the Deployment, as that is additive at this stage.

jpodivin:
> +2
> 
> I see the earlier comment from @rabi to optimize the looping. Is that something we could do in a follow up in the interest of getting this merged since we have green CI now?

If we are to do that, it should be included in further changes to how we handle services, like dependencies. We need to properly define the architecture and make sure that we don't have to come back to this again.

controllers/openstackdataplanedeployment_controller.go
gibizer: >  Now nodesets in a deployment are deployed in parallel, so one can create multiple deployments, one each per cell, no?

gibizer: >  Now nodesets in a deployment are deployed in parallel, so one can create multiple deployments, one each per cell, no? 

then we are back to the original request to run some service against all nodes in multiple nodesets and now in multiple deplotments too.

gibizer: I agree to have a separate design discussion about this to get us on the same page.

tests/functional/openstackdataplanedeployment_controller_test.go
```diff
@@ -186,4 +200,230 @@ var _ = Describe("Dataplane Deployment Test", func() {
...
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      ansibleeeName.Name,
+							Namespace: ansibleeeName.Namespace,
+						}}
+					g.Expect(th.K8sClient.Get(th.Ctx, ansibleeeName, ansibleEE)).To(Succeed())
+					ansibleEE.Status.JobStatus = ansibleeev1.JobStatusSucceeded
+
+					g.Expect(th.K8sClient.Status().Update(th.Ctx, ansibleEE)).To(Succeed())
+				}, th.Timeout, th.Interval).Should(Succeed())
+			}
```
gibizer, line 413: could you assert that the AEE for the global service has the host override variable set to all while the rest of the services has that variable set to the nodeset name? I'm OK to do that in a follow up

gibizer:
If there will be follow ups for the testing request form @slagle and me then I'm OK to land this as is

jpodivin:
> If there will be follow ups for the testing request form @slagle and me then I'm OK to land this as is

There will be.

jpodivin:
recheck `retry limit`

gibizer (commented):
thanks for the additional assert. looks good to me

slagle:
recheck

slagle (approved):

jpodivin:
recheck
```

The connection to the server oauth-openshift.apps-crc.testing was refused - did you specify the right host or port?
Can not login to the CRC cluster. Sleeping...
Checking login to the cluster 19...
WARNING: Using insecure TLS client config. Setting this option is not supported!
```

jpodivin:
recheck

final attempt
```

The connection to the server oauth-openshift.apps-crc.testing was refused - did you specify the right host or port?
Can not login to the CRC cluster. Sleeping...
Checking login to the cluster 19...
WARNING: Using insecure TLS client config. Setting this option is not supported!
```

jpodivin:
recheck

PR #626, opened 2024-01-26
Drop OpenStackDataPlaneService.Spec.Label field

slagle:
The field was proving to be redundant, since it always matched the name
of the Service. It was only really being used to enforce a length, that
logic is moved to a function to just use the first 53 chars of the name,
then the field can be dropped.

### DIFF
--- a/api/v1beta1/openstackdataplaneservice_types.go
+++ b/api/v1beta1/openstackdataplaneservice_types.go
@@ -42,15 +42,6 @@ type OpenstackDataPlaneServiceCert struct {
 
 // OpenStackDataPlaneServiceSpec defines the desired state of OpenStackDataPlaneService
 type OpenStackDataPlaneServiceSpec struct {
-	// Label to use for service.
-	// Must follow DNS952 subdomain conventions.
-	// Since we are using it to generate the pod name,
-	// we need to keep it short.
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Pattern="[a-z]([-a-z0-9]*[a-z0-9])"
-	// +kubebuilder:validation:MaxLength=53
-	Label string `json:"label,omitempty"`
-
 	// Play is an inline playbook contents that ansible will run on execution.
 	Play string `json:"play,omitempty"`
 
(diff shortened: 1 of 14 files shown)


pkg/deployment/const.go
```diff
@@ -53,4 +53,7 @@ const (
 
 	// CACertPaths base path for CA cert volume mount in OpenStackAnsibleEE pod
 	CACertPaths = "/var/lib/openstack/cacerts"
+
+	// AnsibleExecutionServiceNameLen max length for the ansibleEE service name prefix
+	AnsibleExecutionServiceNameLen = 53
```
fao89, line 58: I think we can move the validation to name instead

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -45,15 +45,6 @@ type KubeService struct {
 
 // OpenStackDataPlaneServiceSpec defines the desired state of OpenStackDataPlaneService
 type OpenStackDataPlaneServiceSpec struct {
-	// Label to use for service.
-	// Must follow DNS952 subdomain conventions.
-	// Since we are using it to generate the pod name,
-	// we need to keep it short.
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Pattern="[a-z]([-a-z0-9]*[a-z0-9])"
-	// +kubebuilder:validation:MaxLength=53
```
fao89, line 54: we can move it to Name

slagle: can we? Name is part of metav1.ObjectMeta, and already has its own Kubernetes validations. https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

fao89: my bad, the Name I saw in `api/v1beta1/openstackdataplaneservice_types.go` was from KubeService

fao89 (approved):

rabi:
recheck

slagle:
recheck

fao89 (approved):

pkg/deployment/const.go
```diff
@@ -59,4 +59,7 @@ const (
 
 	// IPValuesStr value for setting ip addresses in a cert
 	IPValuesStr = "ips"
+
+	// AnsibleExecutionServiceNameLen max length for the ansibleEE service name prefix
+	AnsibleExecutionServiceNameLen = 53
```
rabi, line 64: This seems redundant as we've it in util package and referred there.

PR #627, opened 2024-01-26
Add test-all target

slagle:
The test-all target includes test, golint golangci golangci-lint so that
there is a single target to run all automated tests and checks (not
including kuttl).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,9 @@ PROC_CMD = --procs ${PROCS}
 test: manifests generate fmt vet envtest ginkgo ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) --trace --cover --coverpkg=../../pkg/...,../../controllers,../../api/v1beta1 --coverprofile cover.out --covermode=atomic ${PROC_CMD} $(GINKGO_ARGS) ./tests/...
 
+.PHONY: test-all
+test-all: test golint golangci golangci-lint ## Run all tests.
+
 ##@ Build
 
 .PHONY: build


pablintino:
recheck

Makefile
```diff
@@ -123,6 +123,9 @@ PROC_CMD = --procs ${PROCS}
 test: manifests generate fmt vet envtest ginkgo ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) --trace --cover --coverpkg=../../pkg/...,../../controllers,../../api/v1beta1 --coverprofile cover.out --covermode=atomic ${PROC_CMD} $(GINKGO_ARGS) ./tests/...
 
+.PHONY: test-all
+test-all: test golint golangci golangci-lint ## Run all tests.
```
rabi, line 127: Maybe also add `govet`  which is run in CI?

slagle: The `test` target already runs the `vet` target. CI uses the `test` target

rabi: > The test target already runs the vet target. CI uses the test target

CI also calls `govet`[1] target which uses the scripts[2] from `openstack-k8s-operators-ci` . I thought the intent was to run everything that CI runs in this target before submitting a PR, though `vet` target would kind of check the same thing.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/actions/runs/7673553024/job/20916393507?pr=627
[2] https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/blob/main/test-runner/govet.sh

rabi (approved):
/lgtm

Makefile
slagle: What change are you asking me to make? The new `test-all` target I add calls the `govet` target indirectly by way of the `test` target.

rabi: Isn't `test` calling `vet` target instead [1]? There are two targets `vet` and `govet` in the `Makefile`  and CI job I linked earlier calls `govet`. I thought it would be better to call `govet` as well, but it's fine. I have approved and we can merge the change.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/Makefile#L123

PR #629, opened 2024-01-28
No dup nodes

bshephar:
This changes adds a webhook validation to ensure that nodes are only present in a single NodeSet. If a user tries to create a new NodeSet with a node that already exists in the cluster, it will be blocked by the webhook.

### DIFF
--- a/api/v1beta1/openstackdataplanenodeset_webhook.go
+++ b/api/v1beta1/openstackdataplanenodeset_webhook.go
@@ -17,8 +17,11 @@ limitations under the License.
 package v1beta1
 
 import (
+	"context"
 	"fmt"
 	"reflect"
+
+	"golang.org/x/exp/slices"
 
 	baremetalv1 "github.com/openstack-k8s-operators/openstack-baremetal-operator/api/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -90,10 +102,68 @@ var _ webhook.Validator = &OpenStackDataPlaneNodeSet{}
...
+			existingNodeNames = append(existingNodeNames, node.HostName)
+			if node.Ansible.AnsibleHost != "" {
+				existingNodeNames = append(existingNodeNames, node.Ansible.AnsibleHost)
+			}
+		}
+	}
+
+	for _, newNodeName := range r.Spec.Nodes {
+		if slices.Contains(existingNodeNames, newNodeName.HostName) || slices.Contains(existingNodeNames, newNodeName.Ansible.AnsibleHost) {
+			errors = *field.Invalid(
```
gibizer, line 157: It would be useful for the user to list all the duplication instead of showing only one of them. So I would use an `field.ErrorList` instead of a `field.Error`. E.g. https://github.com/openstack-k8s-operators/nova-operator/blob/ccdb40976d65fdbc7c401ac345433ab6d357dbd0/api/v1beta1/nova_webhook.go#L143

bshephar: Probably, but this is already going to be really inefficient if we have 100's of nodes. So my intention was to just exit early as soon as we hit a duplicate, rather than continuing to iterate over any beyond the duplicate.

bshephar: Just realised now that I'm still iterating over the rest of them anyway, and this would just end up with the last error, not the first error...

bshephar: Done in the latest iteration.

gibizer (approved):

slagle (approved):

slagle:
> I'm wondering how much API response delay this webhook will add as it needs to query the k8s API for NodeSets. The general recommendation is https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#availability

It's something to keep in mind. If it proves to be an issue, we could move the check to the NodeSet reconcile.

PR #630, opened 2024-01-29
Don't use nodeName for IPSets

rabi:
hostName would be unique and let's use it instead. Also, we use nodeNames as hostNames when not provided.

Also changes certs to use hostName.


rabi:
2024-01-29T05:31:59Z	ERROR	controller-runtime.source	if kind is a CRD, it should be installed before calling Start	{"kind": "RabbitmqCluster.rabbitmq.com", "error": "no matches for kind \"RabbitmqCluster\" in version \"rabbitmq.com/v1beta1\""}
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1.1

fao89 (approved):

slagle (approved):

bshephar (approved):

PR #633, opened 2024-01-30
This change ensures the `dataplane-operator` is built with FIPS compliance

rebtoor:
- Changed the build image to `ubi9/go-toolkit`
- Changed the base image to `ubi9/minimal`
- Added the default `GO_BUILD_EXTRA_ARGS="-tags strictfipsruntime"`
- Added the `GO_BUILD_EXTRA_ENV_ARGS` build argument to allow custom build arguments at build time. It defaults to `"CGO_ENABLED=1 GO111MODULE=on"`
- Those default parameters have been added to enable FIPS compliance
- Fixed indentation
- Removed `TARGETOS` and `TARGETARCH` env vars.
- Added `DOCKER_BUILD_ARGS` variable in Makefile to pass custom parameters during `podman build`
- Added export `FAIL_FIPS_CHECK=true` in `.prow_ci.env` file

Closes: https://issues.redhat.com/browse/OSPRH-3121

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-ARG GOLANG_BUILDER=golang:1.19
-ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
+ARG GOLANG_BUILDER=registry.access.redhat.com/ubi9/go-toolset:1.19
+ARG OPERATOR_BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-minimal:latest
 
 # Build the manager binary
 FROM $GOLANG_BUILDER AS builder
@@ -12,13 +12,13 @@ ARG REMOTE_SOURCE_DIR=/remote-source
 ARG REMOTE_SOURCE_SUBDIR=
 ARG DEST_ROOT=/dest-root
 
-ARG TARGETOS
-ARG TARGETARCH
-ARG GO_BUILD_EXTRA_ARGS=
+ARG GO_BUILD_EXTRA_ARGS="-tags strictfipsruntime"
+ARG GO_BUILD_EXTRA_ENV_ARGS="CGO_ENABLED=1 GO111MODULE=on"
 
 COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR
 WORKDIR $REMOTE_SOURCE_DIR/$REMOTE_SOURCE_SUBDIR
 
+USER root
 RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 
 # cache deps before building and copying source so that we don't need to re-download as much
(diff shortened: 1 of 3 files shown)


slawqo (commented):
/lgtm for me

rebtoor:
recheck

rebtoor:
recheck

Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG GOLANG_BUILDER=golang:1.19
-ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
+ARG GOLANG_BUILDER=registry.access.redhat.com/ubi9/go-toolset:1.19
+ARG OPERATOR_BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-minimal:latest
```
rabi, line 2: `ubi-micro` would be better choice I guess. So we plan to use the same imagesd both upstream/downsteam?

Dockerfile
```diff
@@ -12,21 +12,21 @@ ARG REMOTE_SOURCE_DIR=/remote-source
 ARG REMOTE_SOURCE_SUBDIR=
 ARG DEST_ROOT=/dest-root
 
-ARG TARGETOS
-ARG TARGETARCH
-ARG GO_BUILD_EXTRA_ARGS=
+ARG GO_BUILD_EXTRA_ARGS="-tags strictfipsruntime"
+ARG GO_BUILD_EXTRA_ENV_ARGS="CGO_ENABLED=1 GO111MODULE=on"
```
rabi, line 16: CGO_ENABLED=1 leads to large binaries (though it enables dynamic linking of C libraries).  Why do we have to enable it when we're using ubi images?

Dockerfile
rebtoor: Ah, didn't know about that, thanks @rabi !

BTW no, i don't think so. But this let us to avoid using different build flags between upstream and downstream.

Dockerfile
rebtoor: Yeah, i get your point about binary size but it seems it's required to be FIPS compliant.

[Here](https://github.com/openshift/check-payload/blob/main/internal/validations/validations.go#L103) you can find the check done in the `check-payload` tool used for analyzing the images.

Dockerfile
rebtoor: @rabi i had to rollback to `minimal` image. It seems that it fails FIPS test because `openssl` package is needed and isn't installed in it, and since a package manager isn't present in that image, we can't bypass the issue. Too bad.

rabi (approved):
/lgtm

slagle (approved):

PR #644, opened 2024-02-05
For baremetal nodes use CloudUserName as AnsibleUser

rabi:
AnsibleUser being different from CloudUserName would create issues as user won't exist.


rabi:
recheck

abays:
recheck

```
Exception '403\nReason: Forbidden\n
```

fao89 (approved):

PR #648, opened 2024-02-07
Improve user docs

fao89:
https://logserver.rdoproject.org/48/648/a36b23a90e9c40e53bde42a9706bc6ed9450f0ad/github-check/dataplane-operator-docs-preview/a9d9959/docs_build/docs_build/dataplane/index-upstream.html


docs/assemblies/con_provisioning-bare-metal-data-plane-nodes.adoc
```diff
@@ -0,0 +1,17 @@
+[id="con_provisioning-bare-metal-data-plane-nodes_{context}"]
```
rabi, line 1: This repeats lot of stuff already have in https://openstack-k8s-operators.github.io/dataplane-operator/#_provisioning_baremetal_nodes

docs/assemblies/proc_creating-a-set-of-data-plane-nodes.adoc
```diff
@@ -0,0 +1,272 @@
...
+                    mtu: {{ lookup('vars', networks_lower[network] ~ '_mtu') }}
+                    use_dhcp: false
+                    primary: true
+                {% endif %}
+                {% endfor %}
+----
++
+You can copy a sample template from https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_network_config/templates. For information about customizing the template, see link:https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/17.1/html/installing_and_managing_red_hat_openstack_platform_with_director/assembly_configuring-overcloud-networking_installing-director-on-the-undercloud#ref_network-interface-configuration-options_custom-network-interface-templates[Network interface configuration options].
+
+. If your nodes are not pre-provisioned, you must configure the bare-metal template:
```
rabi, line 145: Same here...

fao89: it is just a quick recap, right below it points the link with the full explanation

rabi: "If your nodes are not pre-provisioned, you must configure the bare-metal template" could be reworded as it' confusing. It could say  "Set the `baremetalSetTemplate` section in `OpenStackDataPlaneNodeSet` spec as required and just point to the detailed doc "Provisioning bare-metal data plane nodes". We probably should not repeat the sample here. A few more things..

1. I would prefer we write it as `bare metal` and not `bare-metal` in the text.
2. Also it would be good to have all NodeSet specific sections including bare metal provisioning before talking about deployment (i.e proc_deploying-the-data-plane.adoc)

jpodivin: I don't think repeating the sample is necessary either, in fact it could make things more painful for us down the line when we'll have to update docs for some reason. Regarding terminology, the primary criterion would be what are we calling things elsewhere. 

If we are using wrong terms everywhere, we need to file an issue. If we do not, we shouldn't introduce more cases.

fao89: I've addressed the comments: https://logserver.rdoproject.org/48/648/a36b23a90e9c40e53bde42a9706bc6ed9450f0ad/github-check/dataplane-operator-docs-preview/a9d9959/docs_build/docs_build/dataplane/index-upstream.html

jpodivin (approved):

PR #651, opened 2024-02-07
Add kuttl tests for TLS certs

vakwetu:
This adds kuttl tests for the tls cert functionality in the dataplane operator.

This depends on https://github.com/openstack-k8s-operators/install_yamls/pull/717 to make sure that the certmanager operator is running as well so we can issue certs.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -54,6 +54,26 @@ rules:
   - update
   - watch
 - apiGroups:
+  - cert-manager.io
+  resources:
+  - certificates
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - cert-manager.io
+  resources:
+  - issuers
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - ""
   resources:
   - configmaps
--- a/controllers/openstackdataplanedeployment_controller.go
+++ b/controllers/openstackdataplanedeployment_controller.go
@@ -54,6 +54,8 @@ type OpenStackDataPlaneDeploymentReconciler struct {
 //+kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplaneservices,verbs=get;list;watch
 //+kubebuilder:rbac:groups=ansibleee.openstack.org,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch;create;update;patch;delete;
+//+kubebuilder:rbac:groups=cert-manager.io,resources=issuers,verbs=get;list;watch;
+//+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete;
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
--- a/tests/kuttl/tests/dataplane-deploy-tls-test/00-assert.yaml
+++ b/tests/kuttl/tests/dataplane-deploy-tls-test/00-assert.yaml
@@ -1,0 +1,199 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: default-service-tls-dnsnames
+spec:
+  label: default-service-tls-dnsnames
+  caCerts: combined-ca-bundle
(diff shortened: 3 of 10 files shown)


vakwetu:
recheck

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -127,6 +127,8 @@ type OpenStackDataPlaneNodeSetReconciler struct {
 //+kubebuilder:rbac:groups=network.openstack.org,resources=dnsdata/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups=cert-manager.io,resources=issuers,verbs=get;list;watch;
```
rabi, line 130: I think certs are managed in openstackdataplanedeployment_controller.go, so better add these rbac rules there.

tests/kuttl/tests/dataplane-deploy-tls-test/02-dataplane-deploy-services-override.yaml
```diff
@@ -0,0 +1,70 @@
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: custom-service1
```
rabi, line 5: Can we rename these services appropriately as `custom-service1` and `custom-service-2` are for different scenarios? What about scenario of default  services in NodeSet?

vakwetu: @rabi  sorry.  I'm not quite sure I understand what you are asking for here.

The TLS cert functionality needs services that request a TLS certificate by having a tlsCert attribute.  

custom-service1 and custom-service2 are examples of these kinds of services - just with slightly different contents requested in the tlsCert.  I'm not sure why we want to name them differently.  If we decide later to add another service to test some different cert content - like a non-default issuer , for instance, how should we name those?

Unless maybe you're asking for the services to be named more explicitly - like:
service-tls-dnsnames-ips-ctlplane-rootca-internal-issuer
service-tls-dnsnames-all-networks-default-issuer

These can get pretty long, but they would describe more explicitly whats being tested.  Though I suppose you can always look at the test to see whats failing.

As for default services, the reason I use custom services is because I don't think nodes are actually created when we run kuttl tests.  We need nodes to be defined in the nodeset with actual ips and dnsnames to generate the certs.  We then check that the certs are generated and mounted properly by the install-certs-override service.  But when the services run, we want to be sure that they don't actually reach out to the non-existent nodes (or the test will fail).  So they run a play that sleeps on localhost.

I think what you're asking for here is for me to create a new default-service with a tls-cert attribute  (that sleeps on localhost) , add that to my definition of the nodeset as a default service, and then do a deployment without service-overrides?

vakwetu: ok -- I think I addressed your comments .. lmk :)

rabi (commented):
Looks like some failures that did not mark the tests failed? 

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_dataplane-operator/651/pull-ci-openstack-k8s-operators-dataplane-operator-main-dataplane-operator-build-deploy-kuttl/1757128502743666688/artifacts/dataplane-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

vakwetu:
@rabi   No - everything passes in the test.  I was a little confused by the log at first too.  The test step in 2-dataplane-deploy and 3-dataplane-deploy-services override that show up in the log repeat until they succeed, and then they repeat until all the other assertions pass too.  You can see that the step passes because the line (for instance):

logger.go:42: 20:23:30 | dataplane-deploy-tls-test/3-dataplane-deploy-services-override | Error from server (NotFound): secrets "cert-cert-custom-service-tls-dns-edpm-compute-0" not found
    logger.go:42: 20:23:30 | dataplane-deploy-tls-test/3-dataplane-deploy-services-override | bad match: 

goes away.

vakwetu:
If the tests do in fact fail, they will ultimately set the result as failed.  I can confirm that through the bunches of testing I did to get this working.

rabi:
recheck

rabi (commented):
/lgtm

rabi (approved):

PR #655, opened 2024-02-09
Add make docs to pre-commit

bshephar:


fao89 (commented):
we need this https://github.com/openshift/release/pull/48637

fao89 (approved):

fao89:
installing ruby wasn't enough, I'm working on https://github.com/openshift/release/pull/48698

jpodivin (changes requested):
I'm not sure this is a good idea. Pre-commit shouldn't take too long to run. It shouldn't really build anything. Instead it should check compliance of newly created commit with existing standard.

jpodivin:
I've run it and it doesn't seem to cause noticeable slowdown. It still needs the prow action fixed, but that's comparatively easy to do.

jpodivin (approved):

PR #661, opened 2024-02-13
Bump k8s version for envtest

bshephar:
This change bumps the version of k8s used in Envtests to match the version used in OCP4.14


slagle (commented):
looks like we've already bumped to 1.28 in 5e6dd1ba69a45a730af241b67e93ad7ddab8664e

PR #664, opened 2024-02-13
Drop warning about existing netconfig during kuttl

slagle:
This warning is no longer needed since a netconfig resource is now
created as part of each kuttl test since commit
a9bab7acb46572f9a74187b9a2c942bc10bbfda6, so it's expected to exist.


rabi (approved):
/lgtm

PR #668, opened 2024-02-14
Bumping actions to use node.js-20

jpodivin:
As per the guidance from github[0] I'm bumping all of the affected actions to new versions.
There may be tangential benefit in new features, but most of all we are going to be using something that is actually supported.

[0] github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/


slagle (approved):

slagle (approved):
/retest

ci failure due to controller-runtime update

PR #669, opened 2024-02-14
Updating the dataplane docs

fao89:
You can find the preview at:
image
after clicking on `details`,  go to `dataplane-operator-docs-preview`
image
click on `logs` and then find the `index-upstream.html` link

https://logserver.rdoproject.org/69/669/528e1c195ccf388bd4f726bbaf4eac9d1bf51c8c/github-check/dataplane-operator-docs-preview/41a96fd/docs_build/docs_build/dataplane/index-upstream.html


docs/assemblies/deploying.adoc
slagle, line 1: There's a lot of overlapping content between this file and creating-the-data-plane.adoc so I don't think we want to just add it as a separate doc.

I think what we need to do with this content is to add any missing documentation into creating-the-data-plane.adoc directly, using separate procs/assemblies where it makes sense. 

@igallagh-redhat WDYT?

igallagh-redhat: Hey, yes, I agree. We don't need `deploying.adoc` - I think it was used as the source for creating the `creating-the-data-plane.adoc` content, so most of the content should be covered already. We should comb through `deploying.adoc` and tag any content that needs its own module (e.g. "=== NodeSet Config Changes", line 248, could be transferred to a concept module, e.g. `con_nodeset-config-changes.adoc` and included in `creating-the-data-plane.adoc`.

docs/assemblies/creating-the-data-plane.adoc
```diff
@@ -20,6 +20,9 @@ To create and deploy a data plane, you must perform the following tasks:
 ** CBO is installed and configured for provisioning.
 ** `BareMetalHosts` registered, inspected, and have the label `app:openstack`.
 * You are logged on to a workstation that has access to the RHOCP cluster as a user with `cluster-admin` privileges.
+* Review the available fields on the dataplane CRD's, further documentation on each field available in the dataplane CRD's is
+available under the <<custom-resources,Custom Resources>> documentation section.
+
```
igallagh-redhat, line 26: Hey, I think you can safely remove this as a prereq, because we point the user to that section in the procedure - see step 5 in https://openstack-k8s-operators.github.io/dataplane-operator/#proc_creating-a-set-of-data-plane-nodes_dataplane. 

I also think you can delete https://openstack-k8s-operators.github.io/dataplane-operator/#ref_OpenStackDataPlaneNodeSet-CR-properties_dataplane in favor of your more comprehensive resource `custom_resources.adoc`, and link to it from step 5 instead.

slagle (approved):

PR #670, opened 2024-02-15
Getting network name missing in dictionary issue

Jaganathancse:
This change is to fix network name missing issue during dataplane deployment.

### DIFF
--- a/config/samples/dataplane_v1beta1_openstackdataplanenodeset_ovs_dpdk.yaml
+++ b/config/samples/dataplane_v1beta1_openstackdataplanenodeset_ovs_dpdk.yaml
@@ -73,7 +73,7 @@ spec:
                   mtu: {{ min_viable_mtu }}
                   # force the MAC address of the bridge to this interface
                   primary: true
-              {% for network in role_networks if network not in ["External", "Tenant"] %}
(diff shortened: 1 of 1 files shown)


rabi (changes requested):
You don't need to add these as those would be auto generated. I've cleaned up the sample in https://github.com/openstack-k8s-operators/dataplane-operator/pull/671

Jaganathancse:
> You don't need to add these as those would be auto generated. I've cleaned up the sample in #671

Getting error due to Tenant network is unavailable in dict,
TASK [osp.edpm.edpm_network_config : Validating arguments against arg spec 'main' - The main entry point for the edpm_network_config role.] ***
Thursday 15 February 2024  01:49:59 +0000 (0:00:01.561)       0:00:10.449 ***** 
fatal: [edpm-compute-0]: FAILED! => {"msg": "'dict object' has no attribute 'Tenant'. 'dict object' has no attribute 'Tenant'"}

what could be the reason if no need to add manually?

slagle (approved):

Jaganathancse:
recheck

rabi (approved):
/lgtm

PR #672, opened 2024-02-15
Instruct install_certs to generate certificates for Telemetry agents

jlarriba:


slagle (approved):

jlarriba:
/hold

The certificate issuing is still under development and api may change

PR #676, opened 2024-02-15
Pass instanceSpec.CtlPlaneIP in CIDR notation

hjensas:
Get the ipPrefix from the Cidr in the ipSet reservation, combine address and prefix when setting CtlPlaneIP in the instanceSpec.

Depends-On: https://github.com/openstack-k8s-operators/openstack-baremetal-operator/pull/129


rabi (approved):
/lgtm

PR #686, opened 2024-02-18
Hash the entire NodeSet spec

bshephar:
This change removes the selective fields we were using to create the ConfigHash. Initially, this was implemented to ensure we only captured user provided values from the Spec, while removing those that are populated during reconcile by the controller. For example, the values in the baremetalSetTemplate. However, since we are now including the baremetalSetTemplate in the current hash, there is no need to be selective about the fields we include.

This change removes the selective spec fields and instead hashes the entire NodeSet spec section.


rabi (approved):
I think reason for this change should be to include services, changing of which would require a new deployment to be created. However, with the overrides for ansible vars and services in deployment there is no way to know what has been deployed on the nodeset looking at this config hash.

PR #687, opened 2024-02-18
Added `check-fips-status` service

rebtoor:
In order to check the status of FIPS enablement on compute nodes, we need to call the `fips_status` playbook added with the related edpm-ansible PR[1]

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/564

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/564


slagle (approved):

slagle:

this is just adding the service file and does not change any functionality.

PR #694, opened 2024-02-20
Add Tags for NodeSets and also cleanup samples

rabi:
- Removes usage of `all_networks` not relevant. The order of networks in the spec would be used as the order when configuring nics.

- Drops `network_skip_config`

Also changes the vars from `role_` to `nodeset_`. We're duplicating them at present to avoid CI issues. Once openstack-operator has been bumped with this change we can remove the `role_` ansible vars usage from
templates.
-


slagle (approved):
it would be good to add to the dataplane docs what specific values for node_tags enable which configurations

PR #697, opened 2024-02-20
Temporarily fix issuers

vakwetu:
The code currently sets the default issuer to
osp-rootca-issuer-internal, which is the label for the issuer - rather than the name of the issuer itself (rootca-internal).

The proper fix for this is to use the label and then call something like GetIssuerByLabel() to get the right issuer - while handling the error if the issuer is not yet available.  We plan to add such a function to lib-common by early next week - and will add code to use it then.

Until then, we can unblock folks trying to test out tls-e by setting the issuer name  correctly.

### DIFF
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -35,8 +35,10 @@ import (
 	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/certmanager"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 )
 
 // EnsureTLSCerts generates  a secret containing all the certificates for the relevant service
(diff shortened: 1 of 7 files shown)


pkg/deployment/cert.go
```diff
@@ -109,7 +109,9 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 
 		if service.Spec.TLSCert.Issuer == "" {
 			// by default, use the internal root CA
-			issuer = certmanager.RootCAIssuerInternalLabel
+			// issuer = certmanager.RootCAIssuerInternalLabel
+			// TODO(alee) Temporarily set this to the issuer name
+			issuer = "rootca-internal"
```
rabi, line 114: I think we can instead use `tls.DefaultCAPrefix + endpoint.EndpointInternal` to align with the pattern used for the issuer name in openstack-operator, rather than hard-coding here...

jlarriba (approved):
/lgtm

Even if we could apply what Rabi says about getting the issuer from const values, this MUST be considered a temporary fix as it is stated on the description, so hardcoding it makes this temporary thing more visible.

We must keep in mind that we really need the GetIssuerByLabel() func in lib-common as soon as possible.

rabi:
> Even if we could apply what Rabi says about getting the issuer from const values, this MUST be considered a temporary fix

Though it would be better to use the label (which  itself is a constant in lib-common), I don't think using the constants that're used in openstack-operator for the issuer name, would make it as bad as hard-coding it.

vakwetu:
recheck

pkg/deployment/cert.go
```diff
@@ -109,7 +111,9 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 
 		if service.Spec.TLSCert.Issuer == "" {
 			// by default, use the internal root CA
-			issuer = certmanager.RootCAIssuerInternalLabel
+			// issuer = certmanager.RootCAIssuerInternalLabel
```
rabi, line 114: We don't need this line here

rabi (approved):

PR #698, opened 2024-02-21
Replace samples with kustomize examples

fao89:


slagle (commented):
I like the direction.
Instead of examples/base/openstackdataplanenodeset.yaml, can just make that use config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml? 

That would further remove duplication between the values.yaml and the sample.

I'd also like to get all fields that are just default values removed from the samples, so that they are minimal. We could have a single sample for each resource that documents all defaults as well.

fao89:
> I like the direction. Instead of examples/base/openstackdataplanenodeset.yaml, can just make that use config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml?
> 

there is some customize logic for building the bundle, so I can't use `config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml`, what I can do is copy its content to `examples/base/openstackdataplanenodeset.yaml`

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

examples/baremetal/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+    baremetalsettemplate:
+      bmhLabelSelector:
+        app: openstack
+      ctlplaneInterface: enp1s0
+      cloudUserName: cloud-admin
+    ansible:
+      vars:
+        # edpm firewall, change the allowed CIDR if needed
+        edpm_sshd_allowed_ranges: ['192.168.111.0/24']
+    networks:
```
rabi, line 20: Would not this create more issues and confusion as values.yaml has networks as a subkey in nodeset where as it maps to spec.nodeTemplate.networks?

examples/bgp/values.yaml
```diff
@@ -0,0 +1,133 @@
...
+        dns_search_domains: []
+        # edpm firewall, change the allowed CIDR if needed
+        edpm_sshd_allowed_ranges: ['192.168.122.0/24']
+        edpm_frr_bgp_uplinks: ['nic2', 'nic3']
+        edpm_frr_bgp_neighbor_password: f00barZ
+        edpm_frr_bgp_ipv4_src_network: bgpmainnet
+        edpm_frr_bgp_ipv6_src_network: bgpmainnet6
+        edpm_ovn_bgp_agent_expose_tenant_networks: true
+    nodes:
+      edpm-compute-0:
```
rabi, line 87: Do we need to repeat the `nodes` section as they're in the base? only the delta can be kustomized I assume. Also, we're now going to maintain 3 files for a sample and then the dependency on the base.

jpodivin (changes requested):
I think documentation should be part of this PR. They target the same repo and make sense only together.

Plus if anything changes here you'll have to remember to make changes there.

fao89:
> I think documentation should be part of this PR. They target the same repo and make sense only together.
> 
> Plus if anything changes here you'll have to remember to make changes there.

I don't want to make this PR bigger, I'll keep the other PR on hold until we merge this one

fao89:
recheck

fultonj (commented):
I am in favor of this proposed change.

For someone who only wants to refernce samples things change a little. Instead of looking directly at the samples [1] they can clone the repo and run kustomize and see the same thing. E.g. I can run `kustomize build examples/pre_ceph_hci/` and see the same outputs I would see if I clicked on [2] and [3].

This makes the samples much more maintainable as all the different examples can inherit from the base and duplicate data is replaced by references. 

I also see that for any change  in the new examples diretory, that you make sure kustomize can build it [4].

Thanks @fao89 

[1] https://github.com/openstack-k8s-operators/dataplane-operator/tree/main/config/samples

[2] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml

[3] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanedeployment_pre_ceph_hci.yaml

[4] https://github.com/openstack-k8s-operators/dataplane-operator/actions/runs/8085585345/workflow?pr=698#L30-L33

slagle (commented):
LGTM, +2
How do we intend to merge this? Just merge this PR and the install_yamls PR as close together as possible?

fao89:
> LGTM, +2 How do we intend to merge this? Just merge this PR and the install_yamls PR as close together as possible?

I'll drop the breaking change on this PR, so it will be easier to coordinate with install_yamls

fao89:
recheck

slagle:
recheck

https://github.com/openstack-k8s-operators/install_yamls/pull/752

slagle (commented):
Can you restore:
config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp.yaml
config/samples/dataplane_v1beta1_openstackdataplanenodeset_bgp_ovn_cluster.yaml
config/samples/dataplane_v1beta1_openstackdataplanenodeset_networker.yaml
config/samples/dataplane_v1beta1_openstackdataplanedeployment_networker.yaml

that way we can merge this without breaking any part of install_yamls?

I think bumping the default KUSTOMIZE_VERSION to v5.0.1 in the Makefile is required as well, since this won't work with v3.8.7

fao89:
I kept the samples that were needed on CI, the other samples is more for development and I believe it won't take long for https://github.com/openstack-k8s-operators/install_yamls/pull/743 to get merged

slagle (approved):

PR #702, opened 2024-02-22
Add service certs to secret hash

vakwetu:
hashes.go creates a hash of all the configs and secrets associated with dataplane services.  This PR adds secrets that are generated when certs are requested for the services.

These secrets may change as the certs are renewed by cert-manager. Adding the cert secrets to hashes.go will help the deployer decide if a new deployment is needed.

Also, to reduce duplicate processing in hashes.go, rather than iterating across all services in all nodesets - we first find a union of the services across nodesets.


slagle (commented):
+2. Can you add some docs as well? Either here or in a follow up.

vakwetu:
kuttl tests fixed and documentation added.

fao89 (approved):

PR #709, opened 2024-02-26
Refer to issuers by label instead of name

vakwetu:
This PR changes cert.go to retrieve an issuer by label instead of by name.  We expect labels to be consistent for issuers, rather than names, and labels are defined in lib-common.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/467

### DIFF
--- a/docs/assemblies/tls.adoc
+++ b/docs/assemblies/tls.adoc
@@ -107,10 +107,11 @@ to the SAN.
 
 ==== issuer
 
-This attribute corresponds to the certmanager issuer that will be used to issue the certificate.
-If the issuers attribute is not set (as in the configuarion above for service1), certificates
-will be issued using the default root CA for internal TLS as defined in lib-common.
-(currently set to "osp-rootca-issuer-internal").
+This attribute corresponds to the label for the certmanager issuer that is used to issue the certificate.
+The label can be different from the name of the issuer. There can be only one issuer with the specified label.
+If more than one issuer has the label, an error is generated. If the issuers attribute is not set, as in the
+configuration for `service1`, the certificates are issued with the default root CA for internal TLS as defined
+in `lib-common`, which is set to the label "osp-rootca-issuer-internal" for the `rootca-internal` issuer.
 
 === addCertMounts
 
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -35,10 +35,8 @@ import (
 	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/certmanager"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 )
 
 // EnsureTLSCerts generates  a secret containing all the certificates for the relevant service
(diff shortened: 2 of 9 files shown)


vakwetu:
Needs lib-common to merge first.  But I'll add a build of that change soon to make sure the ci passes.

pkg/deployment/cert.go
```diff
@@ -112,15 +111,18 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 
 		if service.Spec.TLSCert.Issuer == "" {
 			// by default, use the internal root CA
-			// issuer = certmanager.RootCAIssuerInternalLabel
-			// TODO(alee) Temporarily set this to the issuer name
-			issuer = tls.DefaultCAPrefix + string(endpoint.EndpointInternal)
+			issuerLabelSelector = map[string]string{certmanager.RootCAIssuerInternalLabel: ""}
 		} else {
-			issuer = service.Spec.TLSCert.Issuer
+			issuerLabelSelector = map[string]string{service.Spec.TLSCert.Issuer: ""}
```
stuggi, line 116: should the issuer parameter in the service spec be the issuer name or the label?, e.g. https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_libvirt.yaml#L14C13-L14C26

I mean I am wondering if it should be something like this:

```
		if service.Spec.TLSCert.Issuer == "" {
			i, err := certmanager.GetIssuerByLabels(ctx, helper, "", issuerLabelSelector)
			if err != nil {
				return &result, err
			}
                        issuer = i.Name
		} else {
			issuer = service.Spec.TLSCert.Issuer
		}

		certSecret, result, err = GetTLSNodeCert(ctx, helper, instance, secretName,
			issuer, labels, hosts, ips, nil)
```

docs/assemblies/tls.adoc
```diff
@@ -107,10 +107,12 @@ to the SAN.
...
-This attribute corresponds to the certmanager issuer that will be used to issue the certificate.
-If the issuers attribute is not set (as in the configuarion above for service1), certificates
-will be issued using the default root CA for internal TLS as defined in lib-common.
-(currently set to "osp-rootca-issuer-internal").
+This attribute corresponds to a label for the certmanager issuer that will be used to issue the
+certificate.  Note that this can be different from the name of the issuer.  It is expected that there
+is only one issuer with the specified label.  If ore than one issuer has the label, an error will be
+generated.  If the issuers attribute is not set (as in the  configuration above for service1), certificates
+will be issued using the default root CA for internal  TLS as defined in lib-common. (currently set to
+"osp-rootca-issuer-internal" for the rootca-internal issuer).
```
igallagh-redhat, line 115:
```suggestion
This attribute corresponds to the label for the certmanager issuer that is used to issue the
certificate. The label can be different from the name of the issuer.  There can be only one issuer with the specified label.  If more than one issuer has the label, an error is generated. If the issuers attribute is not set, as in the  configuration for `service1`, the certificates are issued with the default root CA for internal  TLS as defined in `lib-common`, which is set to "osp-rootca-issuer-internal" for the `rootca-internal` issuer.
```
Hey, some suggestions for minimalism/clarity, and to align with style guide:

- Avoid future tense.
- Avoid parenthesis in text.

igallagh-redhat (changes requested):
Hey, I've added suggestions for the doc update.
Thanks
Irina

pkg/deployment/cert.go
vakwetu: Well, this depends on whether we plan to create labels for other "standard" CAs - like for example, for OVN or libvirt, and have those labels specified somewhere central like lib-common and used on the control plane.   If so, then we need to use labels.

Its relatively trivial to add a label to an issuer if someone wants to use a different issuer. So the use of names vs. labels is really a design/convenience choice.  Labels across the board would be consistent though.

@stuggi @olliewalsh  - so what do we plan to do for OVN etc, ?

docs/assemblies/tls.adoc
vakwetu: thanks!  will change as provided

pkg/deployment/cert.go
vakwetu: as discussed on slack, will use labels

slagle (commented):
+2

slagle (approved):

PR #714, opened 2024-02-27
Fix cert secret names

vakwetu:
The code in lib-common already adds "cert-" to the name of the secret
    for the cert.  With our code, we end up with secrets with the name
    "cert-cert-foo" which is redundant and at the least inconsistent with
    the control plane.
    
    Also, it looks like the check for the existence of the secret cert would
    always fail (as it was looking for the wrong secret name), so we fixed it.
    
    Finally, just modified the kuttl test to create a TLS combined-ca-bundle
    more consistent with what we expect from the
    openstack-operator/lib-common.

### DIFF
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -58,7 +58,7 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 		var ips []string
 		var issuer *certmgrv1.Issuer
 		var issuerLabelSelector map[string]string
-		var secretName string
+		var certName string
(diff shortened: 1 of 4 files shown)


vakwetu:
Note - this change includes changes from my other outstanding commit

https://github.com/openstack-k8s-operators/dataplane-operator/pull/709

Setting not to merge yet, till we resolve questions there.

igallagh-redhat (approved):
Doc change approved.

slagle (approved):

PR #721, opened 2024-02-29
Adding subscription-manager InitContainer

fao89:
Provide a way to execute `subscription-manager register` without exposing the password at `edpm_bootstrap_command`

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -1054,6 +1054,8 @@ spec:
                       - subnetName
                       type: object
                     type: array
+                  subscriptionManagerSecret:
+                    type: string
                   userData:
                     properties:
                       name:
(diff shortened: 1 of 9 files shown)


slagle (commented):
I think we can use a similar design as you've done here, but do it generically. We should avoid hard coding the special handling for the bootstrap service, and the subscription-manager command, etc.

I have 2 ideas how this might work:

1. Add a secret field on the NodeSet, where we set env vars in the ansibleEE based off of every key in that secret. This is very similar to what you've done here, but make it generic. Any ansible content could then consume those secret values from the env vars.

2. Add a secret to the `secrets` list on the bootstrap service. Modify the edpm_bootstrap ansible to read the the secret values when they're mounted into the pod. Make the secret key/values available as host vars with `set_fact`. Any command in `edpm_bootstrap_command` could then make use of those vars.

Both approaches require the user to know ahead of time the name of the keys they intend to consume with `edpm_bootstrap_command`. For instance, I'd need to know to create a secret that has data.username so that I can use `{{ username }}` (or whatever we call the env var) in `edpm_boostrap_command`. I think this is reasonable as there could be any secret data that could be passed in that we don't have to keep adding support for (org_id or activation key as an example).

fao89:
I'll try implement 1

fao89:
I've started this: https://github.com/openstack-k8s-operators/dataplane-operator/pull/723

PR #722, opened 2024-02-29
Adapted telemetry variable names to the dataplane guidelines

jlarriba:


slagle (commented):
these images also need to be added to config/default/manager_default_images.yaml. That may have been missed when they were first added.

jlarriba:
> these images also need to be added to config/default/manager_default_images.yaml. That may have been missed when they were first added.

Yeah, that was missed when they were added.

slagle (approved):
thanks

jlarriba:
recheck

`Error from server (NotFound): deployments.apps "nmstate-operator" not found`

jpodivin (approved):

PR #725, opened 2024-03-01
Expanding existing kuttl tests to cover case with multiple nodesets

jpodivin:
Tests addition of new nodeset into existing deployment and redeploying.

### DIFF
--- a/tests/kuttl/tests/dataplane-deploy-no-nodes-test/06-assert.yaml
+++ b/tests/kuttl/tests/dataplane-deploy-no-nodes-test/06-assert.yaml
@@ -1,0 +1,160 @@
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNodeSet
+metadata:
+  name: edpm-compute-beta-nodeset
+  namespace: openstack
+spec:
+  services:
+  - download-cache
+  - bootstrap
+  env:
+  - name: ANSIBLE_FORCE_COLOR
+    value: "True"
+  nodes: {}
+  nodeTemplate:
+    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+status:
+  conditions:
+  - message: NodeSet Ready
+    reason: Ready
+    status: "True"
+    type: Ready
+  - message: Deployment completed
+    reason: Ready
+    status: "True"
+    type: DeploymentReady
+  - message: Input data complete
+    reason: Ready
+    status: "True"
+    type: InputReady
+  - message: ServiceAccount created
+    reason: Ready
+    status: "True"
+    type: ServiceAccountReady
+  - message: Setup complete
+    reason: Ready
+    status: "True"
+    type: SetupReady
+  configHash: n696hcdh665h56ch74hc4h559hc5hdch5cdh5b8hb5h694hd8h65bh696h56chfhc4hcbh64fh5b8h5b5h9dh58h8bh98h97hf6h8h5cdhf9q
+  configMapHashes:
+    ovncontroller-config: n56h54bh9bhcbh65ch9fhdh66dh95h5dch569h678h7fh599h7ch84h597h59h54dh58dhf6h66bh565h4hc4h587h645hd7hcch5d8h5f4h55cq
+  deployedConfigHash: n696hcdh665h56ch74hc4h559hc5hdch5cdh5b8hb5h694hd8h65bh696h56chfhc4hcbh64fh5b8h5b5h9dh58h8bh98h97hf6h8h5cdhf9q
+  deploymentStatuses: {}
+  secretHashes:
+    neutron-dhcp-agent-neutron-config: n68h676h98h689hd4h575h5dbh694h6fh688h57h665h5c5h56dh5ddh65bh5d7h5cdh644hb8h8fh5d9h5b9h555h9ch56dh5fh6chd4h5c5h5c5h68q
+    neutron-ovn-agent-neutron-config: n5f4h89hb8h645h55bh657h9fh5d9h5c6h595h9dh667h5f4hfhffh7fh685h56ch57fh679h5ddh5ddh95h696hbch5c7h669h84h54dh685hfh85q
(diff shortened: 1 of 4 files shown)


slagle (commented):
Adding a NodeSet to a Deployment that is already deployed (Deployment.Status.Deployed=true) isn't going to do anything. You'd need to create a new Deployment instead.

I don't think your assert actually asserts that any OpenStackAnsibleEE's were created for the edpm-compute-beta-nodeset. So, nothing was actually deployed.

Looks like this did find a bug though, in that we shouldn't add the deployedConfigHash, ConfigMapHashes, and SecretHashes to the NodeSet if it's not actually deployed.

I think we need to update the Deployment reconciler to add NodeSetHashes to Deployment.Status. Then from the deploymentWatcher for the NodeSet, we can actually confirm if the Deployment deployed the actual NodeSet.Status.ConfigHash, before we copy any deployed hashes from the Deloyment to the NodeSet.

jpodivin:
> Adding a NodeSet to a Deployment that is already deployed (Deployment.Status.Deployed=true) isn't going to do anything. You'd need to create a new Deployment instead.
> 
> I don't think your assert actually asserts that any OpenStackAnsibleEE's were created for the edpm-compute-beta-nodeset. So, nothing was actually deployed.
> 
> Looks like this did find a bug though, in that we shouldn't add the deployedConfigHash, ConfigMapHashes, and SecretHashes to the NodeSet if it's not actually deployed.
> 
> I think we need to update the Deployment reconciler to add NodeSetHashes to Deployment.Status. Then from the deploymentWatcher for the NodeSet, we can actually confirm if the Deployment deployed the actual NodeSet.Status.ConfigHash, before we copy any deployed hashes from the Deloyment to the NodeSet.

Right. The nodeset reaches all the conditions but the deployment is already done.

slagle (changes requested):
The ansible execution for edpm-compute-beta-nodeset is not actually "done" though. So, the test you've added here tests a buggy behavior (and passes). That's not what we want.

slagle:
> The ansible execution for edpm-compute-beta-nodeset is not actually "done" though. So, the test you've added here tests a buggy behavior (and passes). That's not what we want.

This would fix that behavior:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/763
Can you give that a review and check my logic there? Let me know what you think.

jpodivin:
> > The ansible execution for edpm-compute-beta-nodeset is not actually "done" though. So, the test you've added here tests a buggy behavior (and passes). That's not what we want.
> 
> This would fix that behavior: #763 Can you give that a review and check my logic there? Let me know what you think.

I've rewritten the test so that it now uses different deployment. So it should no longer get "deployed" due to bug.

However I think the better option would be to prevent addition of new nodesets into deployment outright, like @SeanMooney suggests.

slagle (approved):

jpodivin:
/retest

Tests still appear to be flaky.

tests/kuttl/tests/dataplane-deploy-no-nodes-test/06-assert.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - download-cache
+  - bootstrap
+  env:
+  - name: ANSIBLE_FORCE_COLOR
+    value: "True"
+  nodes: {}
+  nodeTemplate:
+    ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+status:
+  conditions:
```
rabi, line 17: You need to add `ServiceAccountReady` condition here too.

        +  - lastTransitionTime: "2024-03-22T08:26:47Z"
        +    message: ServiceAccount created
        +    reason: Ready
        +    status: "True"
        +    type: ServiceAccountReady

jpodivin: Done, config hashes too.

jpodivin:
recheck

```
timed out waiting for the condition on openstackdataplanenodesets/openstack-edpm-ipam
```

tests/kuttl/tests/dataplane-deploy-no-nodes-test/06-assert.yaml
```diff
@@ -0,0 +1,60 @@
...
+    type: InputReady
+  - message: ServiceAccount created
+    reason: Ready
+    status: "True"
+    type: ServiceAccountReady
+  - message: Setup complete
+    reason: Ready
+    status: "True"
+    type: SetupReady
+  configHash: n696hcdh665h56ch74hc4h559hc5hdch5cdh5b8hb5h694hd8h65bh696h56chfhc4hcbh64fh5b8h5b5h9dh58h8bh98h97hf6h8h5cdhf9q
```
rabi, line 38: Is there any point of asserting these config hashes? I had removed them from other tests earlier as someone adding anything has to run the tests to find these hashes or calculate them manually.

rabi (commented):
Can we also assert the AnsibleEE resources for the nodesets as we do in other tests?

tests/kuttl/tests/dataplane-deploy-no-nodes-test/06-assert.yaml
jpodivin: The point is exactly that. If the hashes change it means something has changed in our code. If the change is expected there is nothing easier than filling the hash. If not, it indicates a potential issue.

This way the developer can't ignore necessity of revising existing tests.

rabi: Hashes are not meant for code change, but the spec changes after deployment.  It's only meant to compared when specs are updated.

jpodivin: I know hashes are computed from the data. You don't have to tell me that. But we are asserting on the result.
If, hypothetically, something changes in the way operator works with spec, or there is a change in kubebuilder defaults or one of many other things, the hash will change. Indicating that something is happening.

And if that happens in PR that shouldn't change the hash, we will know that it's touching something it shouldn't touch.

jpodivin:
Affected by ongoing CIX, do not recheck.

jpodivin:
recheck - certmanager

slagle (approved):

PR #730, opened 2024-03-05
Delete samples in favor of kustomize examples

fao89:


rabi (approved):
I assume nothing would break after https://github.com/openstack-k8s-operators/install_yamls/pull/743

PR #718, opened 2024-02-28
Initial code for multiple secrets

mauricioharley:
This code allows the creation of multiple secrets when either the number of nodes or of services or both of them create more certificates than the maximum size a secret can accommodate.

PR #731, opened 2024-03-05
Include ca cert in service cert secret

olliewalsh:
Libvirt/OVN need the specific CA cert, not a bundle

### DIFF
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -134,6 +134,7 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 		// To use this cert, add it to the relevant service data
 		certsData[hostName+"-tls.key"] = certSecret.Data["tls.key"]
 		certsData[hostName+"-tls.crt"] = certSecret.Data["tls.crt"]
+		certsData[hostName+"-ca.crt"] = certSecret.Data["ca.crt"]
 	}
 
 	// create a secret to hold the certs for the service


rabi (commented):
/lgtm

vakwetu:
@olliewalsh  

Ok - this isn't originally what I had in mind - I figured one would basically label the secret containing the libvirt cert and use this as the bundle - but I suppose we can make it work.

Part of the reason for this was to reduce the size of the secrets that are mounted in the ansibleEE container.  Recall that these secrets must contain all the certs, keys and now cacerts for all nodes in the nodeset.  But we have a PR up to handle the size of this - so it should be fine.  See https://github.com/openstack-k8s-operators/dataplane-operator/pull/718

vakwetu:
We'll have to adjust that - as well as the docs in https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/docs/assemblies/tls.adoc

olliewalsh:
> @olliewalsh
> 
> Ok - this isn't originally what I had in mind - I figured one would basically label the secret containing the libvirt cert and use this as the bundle - but I suppose we can make it work.
> 
> Part of the reason for this was to reduce the size of the secrets that are mounted in the ansibleEE container. Recall that these secrets must contain all the certs, keys and now cacerts for all nodes in the nodeset. But we have a PR up to handle the size of this - so it should be fine. See #718

ack, can change if you want. This was really just a quick fix for libvirt as what was in place was not secure.

olliewalsh:
Option 1 isn't an option. The CA bundle(s) must not be used to validate libvirt client certs, it would give everyone with a signed cert from one of the trusted public CAs admin access to all of the hypervisors.

olliewalsh:
If libvirt needs to trust a CA it should be added to the system-wide trusted CAs and mounted in the container

SeanMooney:
> Option 1 isn't an option. The CA bundle(s) must not be used to validate libvirt client certs, it would give everyone with a signed cert from one of the trusted public CAs admin access to all of the hypervisors (unless sasl auth has been implemented in the edpm role, not sure if that has yet).

sasl auth has not been implemented and its not currently planed for ga.

we could implement it for FP1 but for GA we need to go with option 2 based on what is currently implemented in the role.

vakwetu:
OK - I can see the benefits of going with option 2 in this case.  We can revisit later as needed, but for now, we'll adjust https://github.com/openstack-k8s-operators/dataplane-operator/pull/718 accordingly.

olliewalsh:
cool, don't think there is any other option for ovn as it doesn't support any auth so all we have is client certs and the ca cert to control access

PR #732, opened 2024-03-06
Switching deployOnAllNodeSets from *bool to bool

jpodivin:
There aren't many benefits to using pointer instead of a value. However, it does require us to check for pointer being nil, if we want it marked as `omitempty`. Since we are already using `bool` values in other places, I believe we should stick with one approach.


slagle (commented):
+2

fao89 (approved):

PR #734, opened 2024-03-06
Allow client auth in cert usage

olliewalsh:
Should be a param but for now continue hardcoding since both ovn and libvirt need client auth.


vakwetu:
+2  I see that the missing usage is ClientAuth.  I'll work on a a patch to make this configurable.

fao89 (approved):

SeanMooney:
yep we require this for libvirt/ovn/ovs perhapse else where.

PR #737, opened 2024-03-07
Don't try to mount all nodeset keys/inventories in AEE

rabi:
It tries to mount keys and inventories for all nodesets with the same mount path and fails when there are multiple nodesets in a deployment.

Jira: https://issues.redhat.com/browse/OSPRH-5371


jpodivin (approved):
Thanks @rabi.

PR #738, opened 2024-03-07
Don't check for services in nodeset when serviceOverride used

rabi:
This is a regression and if some services mentioned in the nodeset does not exist deployment/nodeset does not go to Ready inspite of all deployments done.

### DIFF
--- a/controllers/openstackdataplanedeployment_controller.go
+++ b/controllers/openstackdataplanedeployment_controller.go
@@ -226,15 +226,18 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context, 
 		globalSSHKeySecrets[nodeSet.Name] = nodeSet.Spec.NodeTemplate.AnsibleSSHPrivateKeySecret
 	}
 
-	if err := deployment.CheckGlobalServiceExecutionConsistency(ctx, helper, nodeSets.Items); err != nil {
-		util.LogErrorForObject(helper, err, "OpenStackDeployment error for deployment", instance)
-		haveError = true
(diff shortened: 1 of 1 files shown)


controllers/openstackdataplanedeployment_controller.go
```diff
@@ -226,15 +226,18 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
...
-	if err := deployment.CheckGlobalServiceExecutionConsistency(ctx, helper, nodeSets.Items); err != nil {
-		util.LogErrorForObject(helper, err, "OpenStackDeployment error for deployment", instance)
-		haveError = true
-		instance.Status.Conditions.MarkFalse(
-			condition.ReadyCondition,
-			condition.ErrorReason,
-			condition.SeverityError,
-			dataplanev1.DataPlaneNodeSetErrorMessage,
-			err.Error())
+	if instance.Spec.ServicesOverride == nil {
```
fao89, line 229: I think we need an else to check the services listed on ServicesOverride

fao89 (approved):

controllers/openstackdataplanedeployment_controller.go
rabi: That check is for repetition of  same global service in multiple nodesets (from the code) that has no meaning when using ServiceOverride. If services don't exist we would get error in https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanedeployment_controller.go#L184

abays (approved):
/lgtm

PR #739, opened 2024-03-07
Fix Deployment statuses

fao89:
`condition.ReadyCondition` gets reset at: https://github.com/openstack-k8s-operators/dataplane-operator/blob/785c63db66eb0fab41cdeffd4af897f99b8859a8/controllers/openstackdataplanedeployment_controller.go#L99-L101

For properly update the status we should use `condition.DeploymentReadyCondition` for deployment statuses


fao89:
recheck

slagle (approved):

jpodivin (approved):

PR #744, opened 2024-03-08
Selecting service by nodeset

jpodivin:
Rather than identifying service ansibleee resources by combination of deployment and nodeset, this PR sets prefix of service and nodeset names, appended with randomly generated string. While this approach does not guarantee true unique identification of Ansibleee resources, it makes namespace collisions substantially less likely. 
Furthermore, it will allow us to retrieve information about previously used ansibleee resources, provided they were not deleted, for purposes of diagnostics.

The service resources will now also be labeled with identification of nodeset they were applied to.

### DIFF
--- a/pkg/util/ansible_execution.go
+++ b/pkg/util/ansible_execution.go
@@ -41,12 +41,12 @@ import (
 func AnsibleExecution(
 	ctx context.Context,
 	helper *helper.Helper,
-	obj client.Object,
+	openStackDeployment client.Object,
 	service *dataplanev1.OpenStackDataPlaneService,
 	sshKeySecrets map[string]string,
 	inventorySecrets map[string]string,
 	aeeSpec *dataplanev1.AnsibleEESpec,
-	targetNodeset string,
+	targetNodeSet string,
 ) error {
 	var err error
 	var cmdLineArguments strings.Builder
@@ -59,20 +59,21 @@ func AnsibleExecution(
 
 	ansibleEEMounts := storage.VolMounts{}
 
-	ansibleEE, err := GetAnsibleExecution(ctx, helper, obj, service.Name)
+	ansibleEE, err := GetAnsibleExecution(ctx, helper, openStackDeployment, service.Name, targetNodeSet)
 	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
 	if ansibleEE == nil {
 
-		executionName := fmt.Sprintf("%s-%s", GetAnsibleExecutionNamePrefix(service), obj.GetName())
+		executionName := fmt.Sprintf("%s-%s-%s-", GetAnsibleExecutionNamePrefix(service), openStackDeployment.GetName(), targetNodeSet)
 		ansibleEE = &ansibleeev1.OpenStackAnsibleEE{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      executionName,
-				Namespace: obj.GetNamespace(),
+				GenerateName: executionName,
+				Namespace:    openStackDeployment.GetNamespace(),
 				Labels: map[string]string{
-					service.Name: string(obj.GetUID()),
-					"osdpd":      obj.GetName(),
+					"openstackdataplaneservice":    service.Name,
+					"openstackdataplanedeployment": openStackDeployment.GetName(),
+					"openstackdataplanenodeset":    targetNodeSet,
 				},
 			},
 		}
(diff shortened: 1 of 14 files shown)


pkg/util/ansible_execution.go
```diff
@@ -59,20 +59,21 @@ func AnsibleExecution(
...
 
-	ansibleEE, err := GetAnsibleExecution(ctx, helper, obj, service.Name)
+	ansibleEE, err := GetAnsibleExecution(ctx, helper, obj, service.Name, targetNodeset)
 	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
 	if ansibleEE == nil {
 
-		executionName := fmt.Sprintf("%s-%s", GetAnsibleExecutionNamePrefix(service), obj.GetName())
+		executionName := fmt.Sprintf("%s-%s-", GetAnsibleExecutionNamePrefix(service), targetNodeset)
```
fao89, line 68: why the extra `-`?

jpodivin: Otherwise the generated name will end up with the random part concatenated directly to the nodeset name. 
Which can get confusing, as the random suffix can include `s`. For example `mysuperservice-defaultnode-syba` would turn into `mysuperservice-defaultnodesyba`.

Is it plural `nodes` or singular `node`? Who can say. This way it's more obvious which part is generated and which is predetermined.

slagle: `obj` is the Deployment.

Don't we actually need the AnsibleEE name to be a combination of Service, NodeSet, and Deployment? Otherwise, just using Service and NodeSet might find completed AnsibleEE's from a previous Deployment wouldn't it?

pkg/util/ansible_execution.go
```diff
@@ -219,22 +220,25 @@ func AnsibleExecution(
...
 	var err error
 	ansibleEEs := &ansibleeev1.OpenStackAnsibleEEList{}
 
 	listOpts := []client.ListOption{
 		client.InNamespace(obj.GetNamespace()),
 	}
 	labelSelector := map[string]string{
-		label: string(obj.GetUID()),
+		label:     string(obj.GetUID()),
+		"nodeset": targetNodeSet,
```
slagle, line 241: can this be `openstackdataplanenodeset` since that is the actual name of the resource type, not `nodeset`

slagle: do we want also want to go ahead and set the labels in a more consistent way? Instead of using a dynamic label based on service name, we could use:

```
"openstackdataplaneservice": <serviceName>,
"openstackdataplanenodeset": <nodeSetName>,
"openstackdataplanedeployment": <deploymentName>
```

Or use UID's instead of names?

jpodivin: I think we should either use existing mix of name key and UID value. Or fully switch to names. Let's face it, names are more readable.

pkg/util/ansible_execution.go
jpodivin: I'll rename the var to make it clear. The function is being used for a single purpose anyway, so it's not like we are going to be supplying other objects to it.

pkg/util/ansible_execution.go
jpodivin: I went with your suggestion.

pkg/util/ansible_execution.go
jpodivin: New name is now set by all three.

slagle (commented):
LGTM, just need to fix a log message

pkg/util/ansible_execution.go
```diff
@@ -247,11 +252,11 @@ func GetAnsibleExecution(ctx context.Context, helper *helper.Helper, obj client.
 
 	var ansibleEE *ansibleeev1.OpenStackAnsibleEE
 	if len(ansibleEEs.Items) == 0 {
-		return nil, k8serrors.NewNotFound(appsv1.Resource("OpenStackAnsibleEE"), fmt.Sprintf("with label %s=%s", label, obj.GetUID()))
+		return nil, k8serrors.NewNotFound(appsv1.Resource("OpenStackAnsibleEE"), fmt.Sprintf("with label %s=%s", openStackService, openStackDeployment.GetUID()))
```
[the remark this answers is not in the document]
jpodivin: Right. Makes sense.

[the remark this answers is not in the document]
jpodivin: Done.

fao89 (approved):

tests/kuttl/tests/dataplane-deploy-global-service-test/01-assert.yaml
slagle, line 1: all of the OpenStackAnsibleEE.metadata.ownerReferences.name need to be updated here to edpm-deploy-global-service since you changed the name of the deployment.

jpodivin:
Abandoned in favor of #770

PR #751, opened 2024-03-11
Fix comment in config example

cschwede:
Node-specific ansibleVars override global ansibleVars, but they are not merged. Fixing the example comment to avoid confusion by users.


rabi (approved):
/lgtm

PR #753, opened 2024-03-11
Add owner to cert request

vakwetu:
Right now, the owner of the certificate object defaults to whatever object created it - which in this case is the deployment.  This of course means that the Certificate is deleted when the deployment is deleted.

The owner of the certificate should instead be set to the nodeset.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/476

### DIFF
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -185,7 +185,7 @@ func GetTLSNodeCert(ctx context.Context, helper *helper.Helper,
 			Usages:      usages,
 		}
 
-		certSecret, result, err = certmanager.EnsureCert(ctx, helper, request)
+		certSecret, result, err = certmanager.EnsureCert(ctx, helper, request, instance)
 		if err != nil {
 			return nil, ctrl.Result{}, err
 		} else if (result != ctrl.Result{}) {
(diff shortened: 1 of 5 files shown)


vakwetu:
Adding hold till the lib-common patch merges

slagle:
this will need to be combined with https://github.com/openstack-k8s-operators/dataplane-operator/pull/772 in order to pass CI

pkg/deployment/cert.go
```diff
@@ -185,7 +185,7 @@ func GetTLSNodeCert(ctx context.Context, helper *helper.Helper,
 			Usages:      usages,
 		}
 
-		certSecret, result, err = certmanager.EnsureCert(ctx, helper, request)
+		certSecret, result, err = certmanager.EnsureCert(ctx, helper, request, instance)
```
rabi, line 188: How do we handle scenario where multiple deployments have the same nodeset and we request certs for the nodes in the nodeset? Should we check for existing certs (are they even going to change) or just patch the certs without worrying like now[1]?

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L229

vakwetu: @rabi  the current logic checks for the existence of the secret that is created by cert-manager when the certificate is issued (i.e. in line 168).  If a secret is found, then err == nil and the code in 171-193 is not invoked.

So subsequent deployments on the same nodeset should not trigger any code to patch the certs.

If a secret is not found, then we enter the loop and either error out (because of some error other than not-found , or we issue the certs).

vakwetu:
recheck

rabi (commented):
/lgtm

PR #754, opened 2024-03-11
Enable TLS-e by default

SeanMooney:
TLS-e is intended to be on by default with an explict opt-out
when not required. This change has already been done for the
contolplane and this commit does the same for the dataplaneNodeSet

This change updates the dataplaneNodeset TLSEnabled default to true
the envtests are modifed to account for this change and
the dataplane-create-test kuttl test is exented to assert the default
behavior.

exisiting test covergeage fo tls supprot is unmodifed.

The base example cr is also updated to make it clear that tls is the
default however that change is not strictly requried.

With this change the tempest jobs should now default to useing
TLS unless they explictly opt out.

Related: [OSPRH-2382](https://issues.redhat.com//browse/OSPRH-2382)


vakwetu:
I'll be super happy when this lands, but we're not quite there yet.  Specifically we're waiting on some work to be completed on the control plane side.

The control plane is not yet enabled for internal TLS by default.

https://github.com/openstack-k8s-operators/openstack-operator/blob/81a98c76364917b7e555711505353e4a6f79b129/apis/core/v1beta1/openstackcontrolplane_types.go#L79

SeanMooney:
> I'll be super happy when this lands, but we're not quite there yet. Specifically we're waiting on some work to be completed on the control plane side.
> 
> The control plane is not yet enabled for internal TLS by default.
> 
> https://github.com/openstack-k8s-operators/openstack-operator/blob/81a98c76364917b7e555711505353e4a6f79b129/apis/core/v1beta1/openstackcontrolplane_types.go#L79

oh ok so we need a patch to enable internal tls on the contolplane first and then use depend on form this to that. to test them together before we can proceed with this

SeanMooney:
ok so this fails copying the cert the same way it was failing for me locally last week

https://logserver.rdoproject.org/54/754/c672cf974e19da1492e5cb219c8ae6348856c1d5/github-check/podified-multinode-edpm-deployment-crc/51a7438/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/install-certs-edpm-deployment-4jcgf/logs/openstackansibleee.log

i tried enabling internal tls after the contolane and daata plane was intillly deployed and it never seamed to create the tls keys for the the compute node

SeanMooney:
there does appar to be a cert for the compute with a .ci-rdo.local domain

https://logserver.rdoproject.org/54/754/c672cf974e19da1492e5cb219c8ae6348856c1d5/github-check/podified-multinode-edpm-deployment-crc/51a7438/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/cert-nova-custom-compute-0.ci-rdo.local.yaml

and another her 

https://logserver.rdoproject.org/54/754/c672cf974e19da1492e5cb219c8ae6348856c1d5/github-check/podified-multinode-edpm-deployment-crc/51a7438/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/openstack-edpm-nova-custom-certs.yaml


its not clear to me why the generic role for copying the cert failed but as the pod output shows it did not copy the libvirt service certs correctly and we can confirm that by looking at 
https://logserver.rdoproject.org/54/754/c672cf974e19da1492e5cb219c8ae6348856c1d5/github-check/podified-multinode-edpm-deployment-crc/51a7438/controller/ci-framework-data/logs/192.168.122.100/openstack/certs/libvirt/

SeanMooney:
ah there is a bug in the generic edpm_install_certs role @vakwetu 

the combined secte has compute-0.ci-rdo.local-tls.key but its trying to copy 
var/lib/openstack/certs/libvirt/compute-0-tls.key

so the generic role is copying using the hostname but the comnined secrert is using the fqdn

this is likely because the hostname filed is the fqdn but the node name is the hostname not the fqdn.

this can either be fixed in the ansible code or in the dataplane operator go code but we need to deciend on one or the other.

vakwetu:
@SeanMooney  nice catch.

The generic secret is using the node hostName as in here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/cert.go#L136 , whereas the install-certs role is copying using the inventory_hostname - https://github.com/openstack-k8s-operators/edpm-ansible/blob/f98ed5f09467620a00e2936c27a80120b634a56c/roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml#L49 

I'm not sure, but is inventory_hostname being populated here :  https://github.com/openstack-k8s-operators/dataplane-operator/blob/684e436f30f4d187082066937cda996f185b650b/pkg/deployment/inventory.go#L143  ?

I have no particular preference for either method.  However, as I'll be at a conference for the next few days - please go ahead and choose one accordingly.  I'll review as I can.

SeanMooney:
thanks this is exactly what i was expcting one is 
hostName := node.HostName
and the other is using 
strings.Split(node.HostName, ".")[0]

while i generally recommend having unique hostnames not just unique fqdns it might be better to update the role to use ansible_hostname instead of inventory_hostname

inventory_hostname in this case is edpm-compute-0

 nodes:
    edpm-compute-0:
      ansible:
        ansibleHost: 192.168.122.100
        ansibleUser: ""
      hostName: compute-0.ci-rdo.local
      networks:
      - defaultRoute: true
        fixedIP: 192.168.122.100
        name: CtlPlane
        subnetName: subnet1
      - name: InternalApi
        subnetName: subnet1
      - name: Storage
        subnetName: subnet1
      - name: Tenant
        subnetName: subnet1

is translated into the following in the inventory

hosts:
        compute-0:
            ansible_host: 192.168.122.100
            canonical_hostname: compute-0.ci-rdo.local
            ctlplane_cidr: 24


so if we use host_vars[inventory_hostname][canonical_hostname]
or we use ansible_hostname 

we will match the key in the cert secrete

i might just use host_vars[inventory_hostname][canonical_hostname]
to ensure its matching the hostname we declared rather then relying on it being discovert via ansibel gather_facts since we are always generating canonical_hostname and  i dont need toworry about zuul/molecue names being differnt then we expect.

SeanMooney:
check-rdo lets see if the package download issues have resolved themselves

SeanMooney:
ok so my fix as written wont quite work

given 

```
    nodes:
      edpm-compute-0:
        ansible:
          ansibleHost: 192.168.122.100
          ansibleUser: ""
        hostName: edpm-compute-0
        networks:
        - defaultRoute: true
          fixedIP: 192.168.122.100
          name: CtlPlane
          subnetName: subnet1
        - name: InternalApi
          subnetName: subnet1
        - name: Storage
          subnetName: subnet1
        - name: Tenant
          subnetName: subnet1
```
the dataplae operator calulats the host names for edpm-comptue-0 as
```
   allHostnames:
      edpm-compute-0:
        ctlplane: edpm-compute-0.ctlplane.example.com
        internalapi: edpm-compute-0.internalapi.example.com
        storage: edpm-compute-0.storage.example.com
        tenant: edpm-compute-0.tenant.example.com
```

and canonical host name is calculated as 

edpm-compute-0.ctlplane.example.com

the key in the libvirt secrete however is 
edpm-compute-0-tls.key

so canonical_hostname will not always match that key.

this is what the relevnet part of the inventory looks like

```
   "hosts": {
      "edpm-compute-0": {
        "ansible_host": "192.168.122.100",
        "canonical_hostname": "edpm-compute-0.ctlplane.example.com",
        "ctlplane_cidr": 24,
        "ctlplane_dns_nameservers": [
          "192.168.122.80"
        ],
```

so really to fix this we need a change to the go code.

we need to normalise the keys of the secrete to either rthe inventory_hostname or the canonical name

SeanMooney:
the Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/596 should now be flipped and that should depend on this patch but ill wait for the ci jobs to complete before doing that.

SeanMooney:
ok cool so this work for tempest i just need to fix the kuttl tests.
I'm going to keep the depndeinces they way i have them now.

rabi:
> TLS-e is intended to be on by default with an explict opt-out when not required

This is orthogonal to what we've had in earlier OSP versions where it was an opt-in. Have we discussed and agreed on the general and upgrade implications for customers and normal non-production deployments?

SeanMooney:
yes the aggree strage is that this will not be change as part of adoption.
meaning if your coming form a non tls-e env you will be expect to disabel tls in the contolplane and datapalne as part of the adoption procedure.

for greenfiled the intent is that all new greenfeild deployments should be tls-e but you can use the same option to opt our in greenfiled as is used for adoption.

i do not belive there is any intent to force tls-e only

we did discuss that as an option and decide the impact to customer was not warranted and default to on was sufficent.

olliewalsh:
> > TLS-e is intended to be on by default with an explict opt-out when not required
> 
> This is orthogonal to what we've had in earlier OSP versions where it was an opt-in. Have we discussed and agreed on the general and upgrade implications for customers and normal non-production deployments?

Previously had to be opt-in as it dependency on an external IPA deployment.
Can be enabled by default going forward. For adoption if TLSe is enabled then it will remain enabled, if it's disabled will remain disabled. For non-prod deployments I don't see an issue, OCP enables TLS by default too.

SeanMooney:
this was passing a few days ago https://review.rdoproject.org/zuul/build/c1dcb0bfb6b34f2787dc2e535caf3018 but now its complaining that there is no issuerer

```
   message: TLSInput error occured in TLS sources Issuer.apps "No Issuer object found
      in namespace " not found
```
so it looks like something regressed that in the dataplane-operator or consol plane operator in the last 3 days

SeanMooney:
check-rdo

vakwetu:
+2  All looks good here.  Lets see if @rabi change will fix the CI issues.

SeanMooney:
to unblock this while we figure out the ci issue im going to split this into two prs
one for the canonical_hostname change and one for enableing by default

rabi:
> For adoption if TLSe is enabled then it will remain enabled, if it's disabled will remain disabled

Is this a binary thing for adoption? As far as I recall most customers that have TLS enabled, use only for public endpoints.

SeanMooney:
> > For adoption if TLSe is enabled then it will remain enabled, if it's disabled will remain disabled
> 
> Is this a binary thing for adoption? As far as I recall most customers that have TLS enabled, use only for public endpoints.

our intent for the next release is to be secure by default, in this context that means that external internal and tls on edpm nodes will be enabled by default for all greenfield installs. we had discussed making this mandatory but decided not to in this release and instead will support disabling it primarily for existing deployments. 

because we are using cert manager to establish the root of trust the overhead to configuring tls in the new installer is much lower as we can simply use self signed certs if a signing cert/ca is not provided.

our intent woudl be that all deployments eventually converge on tls-e being the enabeld so that we can remove the optionaltiy at a later date but that is unlikely to happen until all customers have moved to the new installer. that effectivly means removing the ablity to opt out of tls is a osp 19 or 20 concern after we have gotten feedback form our custoemrs on the tls experince in 18 

from a security audit perspective not using tls for internal apis and db/rabbit is a security harding opertunrity that we want to adress in our default configs going forward as an enhancement over our existing installer.

rabi:
> from a security audit perspective not using tls for internal apis and db/rabbit is a security harding opertunrity that we want to adress in our default configs going forward as an enhancement over our existing installer.

Though it's good goal to have, it's probably not something we should force customers when upgrading. Anyway, switching to OSP18 is going to be enough pain (with no OCP knowledge) for them and we're seeing lots of push-back from customers on that.

SeanMooney:
> > from a security audit perspective not using tls for internal apis and db/rabbit is a security harding opertunrity that we want to adress in our default configs going forward as an enhancement over our existing installer.
> 
> Though it's good goal to have, it's probably not something we should force customers when upgrading. Anyway, switching to OSP18 is going to be enough pain (with no OCP knowledge) for them and we're seeing lots of push-back from customers on that.

correct which is why for adoption we are going to state that if you are adopting you should not change the tls configuration. we will be adapting the adoption procedure to specifically tell customer to override the new default to maintain the current tls state during the adoption

this is the same as fips.

we will have fips enabled by default for greenfiled but we will not support changing the fips state during adoption. environments without fips enabled will have to opt out of fips during adoption.

for fips unlike tls we have not committed to support enabling fips in place in existing clouds.
for tls that is much simpler to do and we intend to support that as a post adoption procedure that is manually done by a human if they choose too.

vakwetu:
The failing tests should be resolved by https://github.com/openstack-k8s-operators/dataplane-operator/pull/779

xek:
I think we need a fix to horizon before this can merge. Horizon is configured to connect to memcache without tls and it's session backend fails resulting in a 500 error https://github.com/openstack-k8s-operators/horizon-operator/issues/302

Furthermore memcached is configured for TLS only, and keystonemiddleware doesn't support TLS https://github.com/openstack-k8s-operators/infra-operator/issues/203

SeanMooney:
check-rdo the crc vm did not reboot

SeanMooney:
check-rdo cert manager failed to install

SeanMooney:
ok so we have some progress the last recheck atucllly ran the ansible but we timed out waiting for the deployment

2024-03-21 22:21:50,093 p=29345 u=zuul n=ansible | fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["oc", "wait", "OpenStackDataPlaneNodeSet", "openstack-edpm", "--namespace=openstack", "--for=condition=ready", "--timeout=40m"], "delta": "0:40:00.159936", "end": "2024-03-21 22:21:50.074034", "msg": "non-zero return code", "rc": 1, "start": "2024-03-21 21:41:49.914098", "stderr": "error: timed out waiting for the condition on openstackdataplanenodesets/openstack-edpm", "stderr_lines": ["error: timed out waiting for the condition on openstackdataplanenodesets/openstack-edpm"], "stdout": "", "stdout_lines": []}

checking the logs there were no errors we were running the last dataplane service when must gather was ran so this just took longer then the time out. if we waited longer it would have likely completed correctly.

this may have just been a slightly slow node. the ci performance was degraded yesterday with many random failures. I'm going to see if I can extend this wait to 1 hour instead of 40m
normally it does not take that long but in ci on a slow node it might

SeanMooney:
i have submitted a pr https://github.com/openstack-k8s-operators/ci-framework/pull/1325 to bump the default deploy timeout and added it as a depends on. then i rebased via the UI to get a new ci run with that change, there is not direct code change from the last version

slagle:
recheck

SeanMooney:
this is the second time the cert manager install has taken more than 2m so I'm going to find that timeout and extend it to ~5 to 10 mins 

TASK [cert_manager : Verify cert_manager api _raw_params={{ lookup('env', 'HOME') }}/bin/cmctl check api --wait=2m] ***
Monday 25 March 2024  13:33:37 -0400 (0:00:02.529)       0:31:45.739 ********** 
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["/home/zuul/bin/cmctl", "check", "api", "--wait=2m"], "delta": "0:02:00.084777", "end": "2024-03-25 13:35:38.162728", "msg": "non-zero return code", "rc": 124, "start": "2024-03-25 13:33:38.077951", "stderr": "", "stderr_lines": [], "stdout": "error: the cert-manager CRDs are not yet installed on the Kubernetes API server", "stdout_lines": ["error: the cert-manager CRDs are not yet installed on the Kubernetes API server"]}

vakwetu:
recheck

vakwetu:
recheck

fao89:
recheck

fao89:
both jobs show an error on telemetry service, but the logs show the telemetry didn't complete the execution (not sure if we need to bump the timeout or if must gather isn't collecting the failed pods)

vyzigold:
Looking at the ansible logs for the telemetry dataplane service, it looks like it just didn't have enough time to finish deploying. Increasing the timeout would probably help.

Look at the logs from the 2 failed jobs:
https://logserver.rdoproject.org/54/754/9b97cd18b47dd7fec164833c7caed0456d1d15e3/github-check/podified-multinode-edpm-deployment-crc/047ba1d/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/telemetry-edpm-deployment-openstack-edpm-dbg2s/logs/openstackansibleee.log

https://logserver.rdoproject.org/54/754/9b97cd18b47dd7fec164833c7caed0456d1d15e3/github-check/cifmw-crc-podified-edpm-baremetal/51b257d/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/telemetry-edpm-deployment-ipam-openstack-edpm-ipam-wddl2/logs/openstackansibleee.log

they each stop at a different task without any failures, this suggests to me it just needed a little bit more time.

fao89:
recheck

vyzigold:
I submitted a PR which fixes the issue with telemetry edpm service. We were missing a `become: true` in one of the tasks.

fao89:
recheck

vakwetu:
recheck

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
```diff
@@ -118,6 +118,7 @@ metadata:
   name: edpm-compute-global
 spec:
   preProvisioned: true
+  tlsEnabled: false
```
fao89, line 121: playbook changed, so we need to update the hashes for fixing kuttl

fao89:
I'll fix the hashes

olliewalsh:
> With this change the tempest jobs should now default to useing
TLS unless they explictly opt out.

Not sure how this PR affects tempest

fao89 (approved):

PR #755, opened 2024-03-12
This change introduces a struct for DNS configuration items.

bshephar:
This struct is used to hold values relevant to the DNS configuration and reduces the cognitive load required to track large numbers of unnamed return values through various functions. It will also reduce the memory footprint required as we aren't creating multiple copies of variables and passing them between functions. But instead we use a pointer to our struct containing values that are required to move between functions.


bshephar:
recheck

slagle (approved):

PR #756, opened 2024-03-12
Watch resources from ansibleVarsFrom

fao89:
Create an index field for user provided configmaps/secrets and watch them


fao89:
recheck

fao89:
recheck

fao89:
recheck

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -489,6 +603,10 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(mgr ctrl.Manager)
 			dnsMasqWatcher).
 		Watches(&dataplanev1.OpenStackDataPlaneDeployment{},
 			deploymentWatcher).
+		Watches(&corev1.ConfigMap{},
+			cmVarsWatcher).
```
rabi, line 607: I think we can use a single func for both secrets and config maps.  Also maybe good to just wtach for configmap/secret version changes.

rabi (commented):
I missed the earlier PR, but EnvFromSource is a type used in container envs and could be misleading in this context. We could have declared a similar struct `AnsibleVarsFromSource` instead here.

slagle (commented):
the added index/watch causes the inventory to get regenerated when any of the ConfigMaps/Secrets in AnsibleVarsFromSource change.

+2, LGTM

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -489,6 +583,10 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(mgr ctrl.Manager)
 			dnsMasqWatcher).
 		Watches(&dataplanev1.OpenStackDataPlaneDeployment{},
 			deploymentWatcher).
+		Watches(&corev1.ConfigMap{},
```
rabi, line 586: This could be expensive as we watch for all configmaps/secrets, though we filter nodes using the index we create. This seems to be the pattern used else where and I don't have a better solution.

rabi (approved):
/lgtm

PR #757, opened 2024-03-12
Set container image vars from OpenStackVersion

slagle:
When setting the container image ansible variables in the generated
inventory, the OpenStackVersion resource is now checked. If an
OpenStackVersion resource exists, values are taken from that resource if
the vars are not already set in AnsibleVars by the user.

If an OpenStackVersion resource does not exist, the default values are
used instead, with the default values still being set from the
$RELATED_IMAGE* environment variables. Again, only if not explicitly set
by the user in AnsibleVars.

A watch on OpenStackVersion is also added to the NodeSet Manager, which
will trigger a reconcile for all NodeSets when OpenStackVersion changes.
The reconcile will cause the inventory to get regenerated.

The previous dnsMasqWatcher is renamed to genericWatcher, since both the
DNSMasq and OpenStackVersion watcher need to use the same logic of
generating reconcile requests for all NodeSets. genericWatcher is then
used as the watch function for both DNSMasq and OpenStackVersion
resources.


slagle:
This depends on https://github.com/dprince/openstack-operator/tree/version_update2. There isn't yet an openstack-operators PR for that branch.

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -44,47 +45,14 @@ import (
...
-		OvnBgpAgent:                util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_OVN_BGP_AGENT_IMAGE", OvnBgpAgentDefaultImage),
-		TelemetryCeilometerCompute: util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_CEILOMETER_COMPUTE_IMAGE", TelemetryCeilometerComputeDefaultImage),
-		TelemetryCeilometerIpmi:    util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_CEILOMETER_IPMI_IMAGE", TelemetryCeilometerIpmiDefaultImage),
-		TelemetryNodeExporter:      util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_NODE_EXPORTER_IMAGE", TelemetryNodeExporterDefaultImage),
-	}
+// NodeConfigElements is a struct containing just the elements used for Ansible executions
+type NodeConfigElements struct {
+	BaremetalSetTemplate baremetalv1.OpenStackBaremetalSetSpec
+	NodeTemplate         dataplanev1.NodeTemplate
+	Nodes                map[string]dataplanev1.NodeSection
```
bshephar, line 55: Looks like we need to rebase on main. This struct has been removed.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -181,16 +185,58 @@ func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
...
+		Logrotate:                  util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_LOGROTATE_CROND_DEFAULT_IMG", LogrotateDefaultImage),
+		NeutronMetadataAgent:       util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_NEUTRON_METADATA_AGENT_DEFAULT_IMG", NeutronMetadataAgentDefaultImage),
+		NovaCompute:                util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_NOVA_COMPUTE_DEFAULT_IMG", NovaComputeDefaultImage),
+		OvnControllerAgent:         util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_OVN_CONTROLLER_AGENT_DEFAULT_IMG", OvnControllerAgentDefaultImage),
+		OvnBgpAgent:                util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_OVN_BGP_AGENT_IMAGE", OvnBgpAgentDefaultImage),
+		TelemetryCeilometerCompute: util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_TELEMETRY_CEILOMETER_COMPUTE_IMAGE", TelemetryCeilometerComputeImage),
+		TelemetryCeilometerIpmi:    util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_TELEMETRY_CEILOMETER_IMPI", TelemetryCeilometerIpmiImage),
+		TelemetryNodeExporter:      util.GetEnvVar("RELATED_IMAGE_OPENSTACK_EDPM_TELEMETRY_NODE_EXPORTER_IMAGE", TelemetryNodeExporterImage),
+	}
```
bshephar, line 241: Nice, this looks better now in this PR.

bshephar (commented):
Looking good so far. Just needs a rebase.

Probably be nice to have some functional tests to ensure that logical hierarchy works as expected in different scenarios.

api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
```diff
@@ -197,6 +197,29 @@ spec:
                 - ctlplaneInterface
                 - deploymentSSHSecret
                 type: object
+              containerImages:
+                properties:
+                  frr:
+                    type: string
+                  iscsiD:
+                    type: string
+                  logrotate:
```
SeanMooney, line 206: as an aside logroate should hopefully go away when we complete the move to having all services log to the journal

api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
```diff
@@ -197,6 +197,29 @@ spec:
...
+                    type: string
+                  neutronMetadataAgent:
+                    type: string
```
SeanMooney, line 209: the neutron sriov nic agent container appears to be missing.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -181,16 +185,58 @@ func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
...
+	IscsiD                     string `json:"iscsiD,omitempty"`
+	Logrotate                  string `json:"logrotate,omitempty"`
+	NeutronMetadataAgent       string `json:"neutronMetadataAgent,omitempty"`
+	NovaCompute                string `json:"novaCompute,omitempty"`
+	OvnControllerAgent         string `json:"ovnControllerAgent,omitempty"`
+	OvnBgpAgent                string `json:"ovnBgpAgent,omitempty"`
+	TelemetryCeilometerCompute string `json:"telemetryCeilometerCompute,omitempty"`
+	TelemetryCeilometerIpmi    string `json:"telemetryCeilometerIpmi,omitempty"`
+	TelemetryNodeExporter      string `json:"telemetryNodeExporter,omitempty"`
+}
```
SeanMooney, line 200: using a stuct instead of a hashmap of sting to sting will mean that this upgrade mechanium cannot be used for thrid party vendor integrations. for example juniper vrouter or any ohter integration that would require addtional containers.

controllers/openstackdataplanenodeset_controller.go
slagle: Done.

api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
slagle: Looks like no one ever added it on main. I'll add it there and then rebase this.

slagle: https://github.com/openstack-k8s-operators/dataplane-operator/pull/760

api/v1beta1/openstackdataplanenodeset_types.go
dprince: wouldn't juniper vrouter deployments get deployed with their own Ansible playbooks? And any env variables there can be set in the Ansible inventory or whatever. The interface we are building here can accomidate custom backends where needed (storage backends). But on the dataplane it unless it is our code that is orchestrating the deployment (i.e. our Ansible) I'm not sure we need to have a parameter to represent anything yet.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -190,17 +198,55 @@ func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
...
-	Logrotate                  string
-	NeutronMetadataAgent       string
-	NeutronSRIOVAgent          string
-	NovaCompute                string
-	OvnControllerAgent         string
-	OvnBgpAgent                string
-	TelemetryCeilometerCompute string
-	TelemetryCeilometerIpmi    string
-	TelemetryNodeExporter      string
+const (
```
slagle, line 201: TODO: rename these constants to match the names better from OpenStackVersion

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -687,13 +659,15 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(mgr ctrl.Manager)
...
-			dnsMasqWatcher).
+			genericWatcher).
 		Watches(&dataplanev1.OpenStackDataPlaneDeployment{},
 			deploymentWatcher).
 		Watches(&corev1.ConfigMap{},
 			varsFromWatcher, builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
 		Watches(&corev1.Secret{},
 			varsFromWatcher, builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
+		Watches(&openstackv1.OpenStackVersion{},
+			genericWatcher).
```
slagle, line 670: TODO: this watcher seems to be causing infinite reconcile loops. Not sure why yet as I don't see the OpenStackVersion resource actually changing.

pkg/deployment/deployment.go
```diff
@@ -113,6 +116,17 @@ func (d *Deployer) Deploy(services []string) (*ctrl.Result, error) {
...
 
 		log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
+
+		// (TODO) Only considers the container image values from the Version
+		// for the time being. Can be expanded later to look at the actual
+		// values used from the inventory, etc.
+		if d.Version != nil {
+			vContainerImages := reflect.ValueOf(d.Version.Status.ContainerImages)
+			for _, cif := range foundService.Spec.ContainerImageFields {
+				d.Deployment.Status.ContainerImages[cif] = reflect.Indirect(vContainerImages.FieldByName(cif)).String()
```
slagle, line 126: TODO: It's not guaranteed that the container image value read here is what was actually used. We haven't considered ansible group/host/extra vars. We might need to check that.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -190,17 +198,55 @@ func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
...
+)
+
+var ContainerImageDefaults openstackv1.ContainerImages
+
+// SetupDefaults - initializes any CRD field defaults based on environment variables
+// called from main.go
+func SetupDefaults() {
+	// Acquire environmental defaults and initialize dataplane defaults with them
+	ContainerImageDefaults = openstackv1.ContainerImages{
+		ContainerTemplate: openstackv1.ContainerTemplate{
```
slagle, line 233: I would have expected that I could use the ContainerTemplate fields directly in the ContainerImages struct since ContainerTemplate is embedded, but that did not work.

controllers/openstackdataplanenodeset_controller.go
dprince: instead of a 'genericWatcher' function you may try something more specific to OpenStackVersion where you can return false if no update is required.

slagle: I tried that and it doesn't make a difference. The cause is actually because when I run openstack-operator locally, it's looping endlessly as well causing OpenStackVersion to constantly reconcile, and thus the NodeSet. That is probably something specific to my environment with the control plane I have deployed, so I'll consider this one a non-issue for now.

api/v1beta1/openstackdataplanenodeset_types.go
slagle: this is addressed in the latest commit

api/v1beta1/openstackdataplanenodeset_types.go
gibizer: nope, this is the way to initialize the embedded fields. The fields will be accessible directly on the ContainerImage though.

api/v1beta1/openstackdataplanedeployment_types.go
```diff
@@ -79,6 +79,12 @@ type OpenStackDataPlaneDeploymentStatus struct {
 
 	// NodeSetHashes
 	NodeSetHashes map[string]string `json:"nodeSetHashes,omitempty" optional:"true"`
+
+	// DeployedVersion
```
gibizer, line 83: Is the the version string form OpenStackVersion.Spec.TargetVersion at the point of time when this Deployment was run?

api/v1beta1/openstackdataplanedeployment_types.go
```diff
@@ -79,6 +79,12 @@ type OpenStackDataPlaneDeploymentStatus struct {
...
+	DeployedVersion string `json:"deployedVersion,omitempty"`
+
+	// ContainerImages
+	ContainerImages map[string]string `json:"containerImages,omitempty"`
```
gibizer, line 87: Does this field consider if a service override was used in this deployment and therefore only a subset of the images get deployed?

config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
```diff
@@ -18,3 +18,6 @@ spec:
     - ctlplane
     issuer: osp-rootca-issuer-internal
   caCerts: combined-ca-bundle
+  containerImageFields:
+  - NovaComputeImage
+  - EdpmIscsidImage
```
gibizer, line 23: I'm not sure this is the complete list. The osp.edpm.nova_storge playbook uses not just the iscsid but also the 
openstack-multipathd image as well.

pkg/util/version.go
```diff
@@ -0,0 +1,86 @@
...
+	openstackv1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+)
+
+// GetVersion
+func GetVersion(ctx context.Context, helper *helper.Helper) (*openstackv1.OpenStackVersion, error) {
+	client := helper.GetClient()
+	log := helper.GetLogger()
+	var version *openstackv1.OpenStackVersion
+	versions := &openstackv1.OpenStackVersionList{}
+	if err := client.List(ctx, versions); err != nil {
```
gibizer, line 34: Is it intentional that we are not filtering to the namespace when looking up the OpenStackVersion object?

slagle: no. I'll add it.

config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
slagle: I don't see that image in config/default/manager_default_images.yaml either. Maybe we need to add it?

api/v1beta1/openstackdataplanedeployment_types.go
slagle: Yes. ContainerImageFields is on the Service, which is the resource being used to get the fields, whether they're listed on the NodeSet or Deployment.

api/v1beta1/openstackdataplanedeployment_types.go
slagle: It's OpenStackVersion.Spec.TargetVersion at the point when the Deployment is successful, that's when it's read from OpenStackVersion and set here.

slagle:
https://github.com/openstack-k8s-operators/openstack-operator/pull/712 has merged, so setting this ready for review

config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
slagle: Added here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/811
and here for openstack-operator: https://github.com/openstack-k8s-operators/openstack-operator/pull/741
once the openstack-operator PR merges, I can update this one.

controllers/openstackdataplanedeployment_controller.go
```diff
@@ -357,6 +367,9 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
 	Log.Info("Set DeploymentReadyCondition true")
 	instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
 	instance.Status.Deployed = true
+	if version != nil {
+		instance.Status.DeployedVersion = version.Spec.TargetVersion
+	}
```
dprince, line 372: There will always be an OpenStackVersion resource in the namespace of the OpenStackControlplane. While it is okay to check for nil here, it might be better to have it wait for that resource to exist, and re-queue if it doesn't

fao89 (commented):
+2

config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
fao89: shouldn't it be appended here?

controllers/openstackdataplanedeployment_controller.go
fao89: +1

slagle:
kuttl error is:

`
2024-04-17T00:44:02Z	ERROR	setup	problem running manager	{"error": "failed to wait for openstackdataplanenodeset caches to sync: timed out waiting for cache to be synced for Kind *v1beta1.OpenStackVersion"}
`

slagle:
> kuttl error is:
> 
> `2024-04-17T00:44:02Z ERROR setup problem running manager {"error": "failed to wait for openstackdataplanenodeset caches to sync: timed out waiting for cache to be synced for Kind *v1beta1.OpenStackVersion"}`

i suppose we'll need to add a running openstack-operator as a hard dependency

slagle:
> kuttl error is:
> 
> `2024-04-17T00:44:02Z ERROR setup problem running manager {"error": "failed to wait for openstackdataplanenodeset caches to sync: timed out waiting for cache to be synced for Kind *v1beta1.OpenStackVersion"}`
> 
> i suppose we'll need to add a running openstack-operator as a hard dependency

attempting to resolve this with: https://github.com/openstack-k8s-operators/install_yamls/pull/808

slagle:
/retest

added Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/808

slagle:
recheck

`
dataplane-operator-docs-preview
2 mins 9 secs
POST_FAILURE
`

slagle:
recheck

slagle:
recheck

slagle:
barmetal job failed on

`
TASK [openshift_setup : Wait for openshift-marketplace pods to be running kubeconfig={{ cifmw_openshift_kubeconfig }}, kind=Pod, namespace=openshift-marketplace, wait=True, wait_condition={'type': 'Ready', 'status': 'True'}, wait_timeout=300] ***
Thursday 18 April 2024  13:36:51 -0400 (0:00:00.925)       0:01:51.173 ******** 
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'NoneType' object has no attribute 'status'
fatal: [localhost]: FAILED! => {"changed": false, "module_stderr": "Traceback (most recent call 
`

looks related to https://github.com/openstack-k8s-operators/ci-framework/pull/1508

slagle:
recheck

fao89:
recheck

go.mod
```diff
@@ -22,7 +22,8 @@ require (
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.3.1-0.20240417144545-d24e7a32d33b
 	github.com/openstack-k8s-operators/openstack-ansibleee-operator/api v0.3.1-0.20240417073905-8802c89ece05
 	github.com/openstack-k8s-operators/openstack-baremetal-operator/api v0.3.1-0.20240417104943-679bfaa3afc1
-	golang.org/x/exp v0.0.0-20240213143201-ec583247a57a
+	github.com/openstack-k8s-operators/openstack-operator/apis v0.0.0-20240418083003-41031f2a1340
```
gibizer, line 25: This will cause a dependency loop because we have the /api and the operator in the same repo and we use git commit based pseudoversions.

1. bump something in openstack-operator/api
2. dataplane-operator bumps to the new openstack-operator/api, this results in a new pseudoversion of dataplane-operator/api as well as they are in the same repo
3. now there is a need to bump openstack-operator to the new dataplane-operator/api version, but this will lead to a new pseudoversion of openstack-operator/api as well and you go to step 2 again
I afraid this will lead to pain in the future.

So if we want to keep the logic that the dataplane-operator reads the OpenStackVersion but want to break the cycle then we either need to

* stop using pseudoversions
* move the /api module to a separate repo
* move the openstack-operator and dataplane-operator into the same repo and use replace lines in the go.mod files to use the dependency from locally and not from github (like we do in all the main go.mod files towards the /api module already)

fao89: +1 to move to the seame repo

PR #763, opened 2024-03-13
Add NodeSetHashes to OpenStackDataPlaneDeployment.Status

slagle:
Add the OpenStackDataPlaneNodeSet.Status.ConfigHash to the
OpenStackDataPlaneDeployment.Status.NodeSetHashes when a deployment
finishes.

Ensure the NodeSet's ConfigHash is in NodeSetHashes from the Deployment
when checking if the deployment was done for the Node from the NodeSet
controller.

This fixes an issue where the NodeSet would be considered deployed if it
were added to a Deployment after the Deployment was already done
(Deployment.Status.Deployed). Now, the NodeSet is only considered
deployed if the Deployment is done and the NodeSet's ConfigHash is in
the Deployment's NodeSetHashes.


SeanMooney:
the meta comment i have in general is that you are not intented to ever modify a deployment CR


its effectivly ment to be write once and then all spec filds should be treated as read only.

so adding the node set hash is not wrong but the i think the usecase

"This fixes an issue where the NodeSet would be considered deployed if it
were added to a Deployment after the Deployment was already done
(Deployment.Status.Deployed). Now, the NodeSet is only considered
deployed if the Deployment is done and the NodeSet's ConfigHash is in
the Deployment's NodeSetHashes."

is in correct and we shoudl reject the addtion of a nodeset to the deployment instead with a webhook.

the only way to "modify" a deployment should be a delete and recreate
so i would adress the problem by rejecting the update.

jpodivin (commented):
I think proposal by @rabi is better fit for our situation.

slagle:
> I think proposal by @rabi is better fit for our situation.

that WFM. I can close this one.

PR #767, opened 2024-03-14
Adding cleanup steps to kuttl tests

jpodivin:


tests/kuttl/common/99-cleanup.yml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+  - kind: OpenStackDataPlaneNodeSet
+  - kind: OpenStackDataPlaneDeployment
+  - kind: Secret
```
slagle, line 6: could also add ConfigMap here

jpodivin: Done.

fao89 (approved):

mhuin:
recheck

morucci:
recheck

rabi:
These cleanup step files are ignored as the file extension is not `yaml`.  Also, if a test fails the cleanup won't happen for that test and other tests could be impacted if running parellel. As we're using single openstack namespace, we have to anyway ensure that we don't use the same name for objects in multiple tests. 

logger.go:42: 10:41:52 | dataplane-create-test | Ignoring 99-cleanup.yml as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$

jpodivin:
> These cleanup step files are ignored as the file extension is not `yaml`. Also, if a test fails the cleanup won't happen for that test and other tests could be impacted if running parellel. As we're using single openstack namespace, we have to anyway ensure that we don't use the same name for objects in multiple tests.
> 
> logger.go:42: 10:41:52 | dataplane-create-test | Ignoring 99-cleanup.yml as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:.yaml)?$

Good point. Although we have number of parallel tests set to 1. So we shouldn't have an issue with this sort of contamination. Unfortunately, since we are declaring our own namespace, we need to handle object cleanup etc. ourselves.

rabi:
> Unfortunately, since we are declaring our own namespace, we need to handle object cleanup etc. ourselves.

Not really, I think they would be cleaned up automatically unless there are failures.. That's what I've noticed locally.

jpodivin:
Even with number of parallel tests set to `1`,  indicating no parallel executions should take place, they still do.
Therefore we can't clean up resources in this way.

PR #768, opened 2024-03-14
DDP package option service label changes

Jaganathancse:
Adding label for ddp-package-option openstack
dataplane service file.


config/services/dataplane_v1beta1_openstackdataplaneservice_ddp_package_option.yaml
```diff
@@ -3,4 +3,5 @@ kind: OpenStackDataPlaneService
 metadata:
   name: ddp-package-option
 spec:
+  label: ddp-package-option
```
fao89, line 6: we don't use label anymore https://github.com/openstack-k8s-operators/dataplane-operator/pull/713

Jaganathancse: @fao89 thanks. will close this PR.

PR #769, opened 2024-03-14
Test samples from inline docs

fao89:
test inline samples with kuttl

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,10 @@
 repos:
 - repo: local
   hooks:
+    - id: doc-sample
+      name: doc-sample
+      language: system
+      entry: docs/docs_to_sample.sh
     - id: gotidy
       name: gotidy
       language: system
--- a/docs/assemblies/creating-the-data-plane.adoc
+++ b/docs/assemblies/creating-the-data-plane.adoc
@@ -35,8 +35,8 @@ include::proc_creating-a-custom-service.adoc[leveloffset=+2]
 include::proc_configuring-a-node-set-for-a-Compute-feature-or-workload.adoc[leveloffset=+2]
 include::proc_building-a-custom-ansible-runner-image.adoc[leveloffset=+2]
 
-include::ref_example-OpenStackDataPlaneNodeSet-CR-for-preprovisioned-nodes.adoc[leveloffset=+1]
-include::ref_example-OpenStackDataPlaneNodeSet-CR-for-bare-metal-nodes.adoc[leveloffset=+1]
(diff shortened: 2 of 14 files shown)


lewisdenny:
recheck

lewisdenny:
recheck

fao89:
recheck tempest

tests/kuttl/tests/docs-test-00/00-assert.yaml
```diff
@@ -0,0 +1,148 @@
...
+      ctlplane: 192.168.122.100
+      internalapi: 172.17.0.100
+      storage: 172.18.0.100
+      tenant: 172.19.0.100
+  ctlplaneSearchDomain: ctlplane.example.com
+  conditions:
+  - message: NodeSetBaremetalProvisionReady error occurred
+    reason: Error
+    severity: Error
+    status: "False"
```
fao89, line 117: I believe it would require a CI change to enable BMO_SETUP

fao89: https://github.com/openshift/release/pull/50180

slagle (commented):
I think this inverts the logic slighly on how we want to approach the samples.

IMO, it ought to be:

1. minimal samples under config/samples that serve as kustomize base
2. kustomize examples that kustomize the samples from 1 
3. A specific doc kustomize examples builds on the other examples
4. kuttl tests use and docs both use what is generated from 3

This PR results in another top level CRs we have to maintain and continue to test. Instead, it should just all come from the base samples and kustomize examples.

Whether we end up documenting how to use kustomize, or a fully generated samples CRs is a separate question, as long as the source is kustomize for all of it.

fao89:
> I think this inverts the logic slighly on how we want to approach the samples.
> 
> IMO, it ought to be:
> 
> 1. minimal samples under config/samples that serve as kustomize base
> 2. kustomize examples that kustomize the samples from 1
> 3. A specific doc kustomize examples builds on the other examples
> 4. kuttl tests use and docs both use what is generated from 3
> 
> This PR results in another top level CRs we have to maintain and continue to test. Instead, it should just all come from the base samples and kustomize examples.
> 
> Whether we end up documenting how to use kustomize, or a fully generated samples CRs is a separate question, as long as the source is kustomize for all of it.

1 and 2 are being addressed here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/764

I started this with docs to kuttl, for keeping the callouts, 
I'll try to see if I can make kustomize add the callouts as comments to implement 3 and 4

fao89:
I realize that if we are taking preprovisioned/baremetal samples, they are already tested on zuul jobs, so we don't need to add kuttl tests for them.
I'll update https://github.com/openstack-k8s-operators/dataplane-operator/pull/764 to address 3 and 4

PR #770, opened 2024-03-15
Fix how we use labels and aee name for services

rabi:
Use deployment and nodeset name/uuid in AEE name/label for normal services. 

Use deployment name/uuid in AEE name/label for global services.

This would fix the intermittent failures in functional and kuttl tests due to name collision of AEEs.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -149,7 +149,8 @@ func (d *Deployer) ConditionalDeploy(
 	}
 
 	if nsConditions.IsFalse(readyCondition) {
-		ansibleEE, err := dataplaneutil.GetAnsibleExecution(d.Ctx, d.Helper, d.Deployment, foundService.Name)
+		_, labelSelector := dataplaneutil.GetAnsibleExecutionNameAndLabel(&foundService, d.Deployment, d.NodeSet)
+		ansibleEE, err := dataplaneutil.GetAnsibleExecution(d.Ctx, d.Helper, d.Deployment, labelSelector)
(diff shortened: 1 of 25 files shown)


jpodivin:
I'm actually already working on this https://github.com/openstack-k8s-operators/dataplane-operator/pull/744

rabi:
> I'm actually already working on this #744

This is mostly ready other than kuttl assert mismatches which has weird hashes in them. I had looked at your PR and I think we should not complicate it with auto-generated ids. I'm fine if that one is ready earlier.. We need to fix the gate asap. All my prs are blocked by the issue and it's happening frequently.

jpodivin:
> > I'm actually already working on this #744
> 
> This is mostly ready other than kuttl assert mismatches which has weird hashes in them. I had looked at your PR and I think we should not complicate it with auto-generated ids. I'm fine if that one is ready earlier.. We need to fix the gate asap. All my prs are blocked by the issue and it's happening frequently.

If you don't want to collaborate on this it's fine. If your fix works I'm all for it.

rabi:
> If you don't want to collaborate on this it's fine. If your fix works I'm all for it.

Not sure I understand. My intent is to fix the issue ASAP and no way ensure that this PR is merged :smile: From the discussion yesterday I understood that we still don't know what the issue is, that's why I looked at it this morning. Isn't that collaboration where I looked at the issue and trying to fix it?  Anyway, I can drop this PR if you want that PR to go in. The other option is we merge this if this fixes the immediate gate issue and is green. We can refine it in your PR.

PR #776, opened 2024-03-18
Service names

jpodivin:


fao89 (approved):

pkg/util/ansible_execution.go
```diff
@@ -266,23 +266,20 @@ func getAnsibleExecutionNamePrefix(serviceName string) string {
...
+	nodeSetName string) (string, map[string]string) {
+	executionName := fmt.Sprintf("%s-%s", getAnsibleExecutionNamePrefix(service.Name), deploymentName)
 	if !service.Spec.DeployOnAllNodeSets {
-		executionName = fmt.Sprintf("%s-%s", executionName, nodeset.GetName())
+		executionName = fmt.Sprintf("%s-%s", executionName, nodeSetName)
 	}
 	if len(executionName) > AnsibleExcecutionNameLabelLen {
 		executionName = executionName[:AnsibleExcecutionNameLabelLen]
 	}
-	labelValue := string(deployment.GetUID())
```
rabi, line 278: I had continued using uid for labels to ensure that long deployment and nodeset names don't make us truncate the labels for AEE. We could have used uid for AEE names too, but we assert the names in kuttl tests (and the plan was to change those). There could be still be name collision, if the service, deployment and nodeset  names are not kept small and unique.

jpodivin: That's a good point. Labels need to follow [RFC-952](https://www.rfc-editor.org/rfc/rfc952), or used to, now the [documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) specifies something more like [RFC-1123](https://www.rfc-editor.org/rfc/rfc1123).

There are ways to handle that though, by setting validation for nodeset, deployment, and service name. As it happens, we already have a package for that.

jpodivin: I've added validation for nodeset, deployment and service names.

morucci:
recheck

morucci:
recheck

morucci:
recheck

pkg/util/ansible_execution.go
rabi: k8s object names can be 253 chars. By limit deployment/nodeset names to 63 chars would not achieve anything as we use a combination of deployment and nodeset name to build the label in this PR.

jpodivin:
recheck

pkg/util/ansible_execution.go
jpodivin: Some of them do, yes. But the validation here is meant to sanitize strings we'll eventually use as labels. These are limited to 63 chars.

controllers/openstackdataplanedeployment_controller.go
```diff
@@ -68,6 +69,12 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
 	Log := r.GetLogger(ctx)
 	Log.Info("Reconciling Deployment")
 
+	// Check if deployment name matches RFC1123 for use in labels
+	validate := validator.New()
+	if err := validate.Var(req.Name, "hostname_rfc1123"); err != nil {
+		Log.Error(err, "error validating OpenStackDataPlaneDeployment name, the name must follow RFC1123")
```
fao89, line 75: maybe we can move this logic to the webhook validations

jpodivin: That would make a lot of sense. Although right now it would mean making service and deployment webhooks just for these.

fao89: yes

fao89 (approved):

jpodivin:
Hit by the go version mismatch error.

slagle (approved):

fao89:
recheck cert_manager

jpodivin:
recheck cert-manager

fao89:
recheck

api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -126,16 +127,25 @@ func (r *OpenStackDataPlaneNodeSet) ValidateCreate() (admission.Warnings, error)
 	if err != nil {
 		return nil, err
 	}
+	// Check if OpenStackDataPlaneNodeSet name matches RFC1123 for use in labels
+	validate := validator.New()
+	if err = validate.Var(r.Name, "hostname_rfc1123"); err != nil {
+		openstackdataplanenodesetlog.Error(err, "Error validating OpenStackDataPlaneNodeSet name, name must follow RFC1123")
+		errors = append(errors, field.Invalid(
+			field.NewPath("Name"),
```
gibizer, line 135: In general you want to return the json path of the CRD field that failed the validation. I think in this case that is field.NewPath("metadata").Child("name").

PR #778, opened 2024-03-20
ensure that cert keys are prefixed by the canonical_hostname

SeanMooney:
previously the cert keys used node.hostname directly

node.hostName could be empty in which case it defaults to the
node name, a hostname or an fqdn

As a result the secret field name of the tls.key, tls.crt and
ca.crt were not stable or predicable.

ansible has 3 possible values it could use to corralate to the
prefix. inventory_hostname which is the the short name
i.e. node.hostname.split('.')[0]

{{ hostvars['inventory_hostname']['canonical_hostname'] }}
This is set to the controlplane network fqdn

{{ ansible_hostname }} this is only aviable after gathering facts
and is the hostname reported by the host as if hostname -f was run
{{ ansible_hostname }} is not always avlaible which leave the first
two options.

we could normalise on the hostname i.e. the short hostname and use
inventory_hostname, this patch uses canonical_hostname instead
to ensure that if we have two nodes with the same shortname but differnt
fqdn that we can support that in the future.

### DIFF
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -19,6 +19,7 @@ package deployment
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


SeanMooney:
the kuttl test failure looks unreatled to this but ill take a look at that later today

vakwetu:
+2

jpodivin:
I've seen similar kuttl test failures recently on other patches. I'm going to say that it's unrelated.

PR #779, opened 2024-03-20
Fix issuer for ovn

vakwetu:
The issuer label for ovn is set incorrectly in the service definition. In lib-common, the label is set to OvnDbCaName = tls.DefaultCAPrefix + "ovn", but we have ovndb instead.

Also, added an error message to help troubleshoot these in future.

### DIFF
--- a/config/services/dataplane_v1beta1_openstackdataplaneservice_ovn.yaml
+++ b/config/services/dataplane_v1beta1_openstackdataplaneservice_ovn.yaml
@@ -12,7 +12,7 @@ spec:
     - ips
     networks:
     - ctlplane
-    issuer: osp-rootca-issuer-ovndb
+    issuer: osp-rootca-issuer-ovn
     keyUsages:
       - digital signature
       - key encipherment
--- a/pkg/deployment/cert.go
+++ b/pkg/deployment/cert.go
@@ -116,8 +116,9 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 			issuerLabelSelector = map[string]string{service.Spec.TLSCert.Issuer: ""}
 		}
 
-		issuer, err = certmanager.GetIssuerByLabels(ctx, helper, "", issuerLabelSelector)
+		issuer, err = certmanager.GetIssuerByLabels(ctx, helper, instance.Namespace, issuerLabelSelector)
 		if err != nil {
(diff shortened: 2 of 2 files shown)


fao89 (commented):
/approve

pkg/deployment/cert.go
```diff
@@ -116,8 +116,9 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 			issuerLabelSelector = map[string]string{service.Spec.TLSCert.Issuer: ""}
 		}
 
-		issuer, err = certmanager.GetIssuerByLabels(ctx, helper, "", issuerLabelSelector)
+		issuer, err = certmanager.GetIssuerByLabels(ctx, helper, instance.Namespace, issuerLabelSelector)
 		if err != nil {
+			helper.GetLogger().Info("Error retrieving issuer by label", "issuerLabelSelector", issuerLabelSelector)
```
SeanMooney, line 121: +1 this will help with future debugging

ade and i briefly chatted about the fact this should be detected as part of the nodeset reconciliation not the deployment reconciliation

moving the validation fo the issuers is out of scope of this pr.

pkg/deployment/cert.go
```diff
@@ -116,8 +116,9 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 			issuerLabelSelector = map[string]string{service.Spec.TLSCert.Issuer: ""}
 		}
 
-		issuer, err = certmanager.GetIssuerByLabels(ctx, helper, "", issuerLabelSelector)
+		issuer, err = certmanager.GetIssuerByLabels(ctx, helper, instance.Namespace, issuerLabelSelector)
```
SeanMooney, line 119: this is currently in the tls-e defaulting patch but it makes more sense to include that here.
for now I'm going to rebase my pr on this pr so we can get a ci run without waiting for this to merge.

we can merge this patch indepently however.

config/services/dataplane_v1beta1_openstackdataplaneservice_ovn.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     - ips
     networks:
     - ctlplane
-    issuer: osp-rootca-issuer-ovndb
+    issuer: osp-rootca-issuer-ovn
```
SeanMooney, line 15: ack so this was the main error
nice find

SeanMooney (approved):

PR #780, opened 2024-03-20
add addtional logging to dataplane-deploy-global-service-test

SeanMooney:
this patch add collectors to the 00-assert.yaml file to collect
the deployment cr and dataplane operator logs on assert failure.

### DIFF
--- a/tests/kuttl/tests/dataplane-deploy-global-service-test/00-assert.yaml
+++ b/tests/kuttl/tests/dataplane-deploy-global-service-test/00-assert.yaml
@@ -1,3 +1,14 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 600
+collectors:
+- type: command
+  command: oc get OpenStackDataPlaneDeployment -n openstack edpm-compute-global -o yaml
+  name: edpm-compute-global-deployment
+- type: command
+  command: oc logs -n openstack-operators -l openstack.org/operator-name=dataplane
+  name: dataplane-operator-logs
+---
 apiVersion: dataplane.openstack.org/v1beta1
 kind: OpenStackDataPlaneNodeSet
 metadata:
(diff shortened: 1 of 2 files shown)


fao89 (approved):

SeanMooney:
im not actuly sure this will work so i just want to see the ci run before removing the hold
knowing our luck the kuttl test will pass and wont trigger this

if it does we can merge this and hope that the next time this fails this will give us detailed logs.

tests/kuttl/tests/dataplane-deploy-global-service-test/00-assert.yaml
```diff
@@ -1,3 +1,14 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 600
+collectors:
+- type: command
+  command: oc get OpenStackDataPlaneDeployment -n openstack edpm-compute-global -o yaml
```
SeanMooney, line 6: on thing im not sure about is if in ci we are expecting to use oc or kubectl I'm also not sure if its available in the default path or if we need to pass the kubeconfig.

we wont really know any of that until we see this run in the ci

tests/kuttl/tests/dataplane-deploy-global-service-test/00-assert.yaml
```diff
@@ -1,3 +1,14 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 600
```
jpodivin, line 3: Isn't default timeout enough?

SeanMooney:
this obvouly didnt break anythin when the test past so ill retrigger the kuttl job and zuul jobs if both pass ill remove the hold and next time we have a random ci failure in this test hopefully it will give us more detailed logs

SeanMooney:
recheck

tests/kuttl/tests/dataplane-deploy-global-service-test/00-assert.yaml
slagle: the timeout in kuttl-test.yaml applies to all tests afaik

slagle (approved):

PR #789, opened 2024-03-25
Remove reboot-os from default services

fao89:


fultonj (commented):
I'm +2 on this as I don't think we should be doing a reboot for every service by default. Only where it's needed. Just wanted to CC @ciecierski since he had done work on that service.

fao89:
recheck

slagle (commented):
aren't other services relying on this being included?
playbooks/bootstrap.yml for instance includes kernelargs.yml from edpm_kernel which marks a reboot being needed, and leaves it up to the edpm_reboot role to handle it.

fao89:
I 

> aren't other services relying on this being included? playbooks/bootstrap.yml for instance includes kernelargs.yml from edpm_kernel which marks a reboot being needed, and leaves it up to the edpm_reboot role to handle it.

I didn't know about that, I'll close this PR

PR #790, opened 2024-03-26
Use nodeset_networks and nodeset_tags

rabi:
There is no concept of role. Let's use the new variables.

Depends-On: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/381
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1338
Depends-On: https://github.com/openstack-k8s-operators/architecture/pull/142

PR #793, opened 2024-03-27
Add nodeset_networks to hostvars

rabi:
Some spec files use snake case networks in nodeset template section and small case networks in node section. Without this it would cause issues when switching to use nodeset_networks.


rabi:
> should we file a jira to clean this up later?

We already have a PR for cleanup https://github.com/openstack-k8s-operators/dataplane-operator/pull/790. When merging the Depends-On PRs, we saw the issue that this trying to resolve.

slagle:
> > should we file a jira to clean this up later?
> 
> We already have a PR for cleanup #790. When merging the Depends-On PRs, we saw the issue that this trying to resolve. I can create a jira if we want.

don't worry about it. I didn't realize it was taken care of already

PR #795, opened 2024-03-28
WIP: Testing behavior with undefined nodeset

jpodivin:


jpodivin:
Implementing negative tests in kuttl appears to be overly complicated with only limited support. Furthermore, given the status of kuttl project, we can't expect a significant change on their side.

jpodivin:
Implementing negative tests in kuttl appears to be overly complicated with only limited support. Furthermore, given the status of kuttl project, we can't expect a significant change on their side.

PR #796, opened 2024-03-28
Reset the state of conditions to Unknown for each reconcile

slagle:
All conditions are set to Unknown at the beginning of the reconcile, as
their state is not yet known for that particular reconcile.

Otherwise, stale and incorrect condition state could be persisted on the
resources indefinitely if they were never reset.

Jira: [OSPRH-5698](https://issues.redhat.com//browse/OSPRH-5698)

### DIFF
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -226,6 +226,10 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		// in the cli
 		return ctrl.Result{}, nil
 	}
+	// Reset all conditions to Unknown as the state is not yet known for
+	// this reconcile loop.
+	instance.InitConditions()
+
 	if instance.Status.ConfigMapHashes == nil {
 		instance.Status.ConfigMapHashes = make(map[string]string)
 	}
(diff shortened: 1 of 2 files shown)


controllers/openstackdataplanenodeset_controller.go
```diff
@@ -226,6 +226,10 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		// in the cli
 		return ctrl.Result{}, nil
 	}
+	// Reset all conditions to Unknown as the state is not yet known for
+	// this reconcile loop.
+	instance.InitConditions()
```
rabi, line 231: I don't see this being done across operators. I had added it in openstack-operator[1] and was dropped in a later PR. I think we can remove L240 if we're doing this.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/155/files#diff-9762aaf2db7cfe1469f3437adc492cb509f3bc85933dddf91a95ca12cd63a639R152

slagle: It's being rolled out to the other operators: https://issues.redhat.com/browse/OSPRH-5698.

If I drop L240, then anywhere between that line and before we mark SetupReady=True, I'd need to set SetupReady=False. That's already a happening in a few places, such as L255. I think the result would be the same either way though.

rabi: It's better to set it the conditions when there is an error. This way there is probably no difference between `unknown` and `false`.

rabi (approved):
/lgtm

PR #801, opened 2024-04-02
Moving kustomize-to-docs invocation to make target

jpodivin:
#764 introduced dependencies on `yq` and `oc` for pre-commit. These were not documented, nor were they installed automatically during pre-commit run and cached. Potentially breaking workflow for developers or forcing them to skip checks.

Since we don't need to build examples for every commit, this script should instead be invoked by make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ docs-dependencies: .bundle
 	bundle config set --local path 'local/bundle'; bundle install
 
 .PHONY: docs
-docs: manifests docs-dependencies crd-to-markdown ## Build docs
+docs: manifests docs-dependencies crd-to-markdown  docs-kustomize-examples ## Build docs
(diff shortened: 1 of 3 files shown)


fao89 (commented):
the idea of having them on pre-commit was to ensure we have updated samples in our docs, if we move it to the makefile, it should be a dependency of `manifests` target

Makefile
```diff
@@ -113,6 +113,10 @@ docs-watch: docs-preview
 docs-clean:
 	rm -r docs_build
 
+.PHONY: docs-examples
+docs-kustomize-examples: yq kustomize ## Generate updated docs from examples using kustomize
```
fao89, line 117: I think you can add it here:
```
.PHONY: docs
docs: manifests docs-dependencies docs-kustomize-examples crd-to-markdown ## Build docs
```
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/Makefile#L96-L97

jpodivin: That makes sense.

fao89 (approved):
thank you!

fao89:
recheck

jpodivin:
/retest

out of memory

jpodivin:
otoh the virtual env will be better.

fao89 (approved):

PR #803, opened 2024-04-03
Drop leftovers for removed CRDs

karelyatin:
openstackdataplanes, openstackdataplanenodes and openstackdataplaneroles are legacy CRDs, cleans up their remaining references.


karelyatin:
/test dataplane-operator-build-deploy-kuttl

secret already exists

PR #815, opened 2024-04-10
Fix kuttl tests when running locally without webhooks

slagle:
The tests that assert hashes rely on webhooks modifying the spec so that
the asserted hash value matches the calculated value. When running
locally without webhooks, these tests would fail. This commit adds the
relevant parts to the specs that the webhooks add so that the caluclated
hashes are equivalent when running locally.

### DIFF
--- a/tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
+++ b/tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
@@ -117,6 +117,22 @@ kind: OpenStackDataPlaneNodeSet
 metadata:
   name: edpm-compute-global
 spec:
+  baremetalSetTemplate:
+    automatedCleaningMode: metadata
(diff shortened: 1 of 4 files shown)


jpodivin (approved):

tests/kuttl/tests/dataplane-deploy-global-service-test/00-dataplane-create.yaml
```diff
@@ -117,6 +117,22 @@ kind: OpenStackDataPlaneNodeSet
 metadata:
   name: edpm-compute-global
 spec:
+  baremetalSetTemplate:
```
rabi, line 120: It's weird that we've to add defaulted fields when creating a CR. This is one of the reasons why I did not want hashes to be asserted in tests unless we're testing the use-cases for them and my feedback was ignored in https://github.com/openstack-k8s-operators/dataplane-operator/pull/725.

PR #817, opened 2024-04-11
WIP: ServiceType field for OpenStackDataPlaneService resource

jpodivin:


slagle (commented):
I'd like to see more of how this will be used before adding it.

jpodivin:
> I'd like to see more of how this will be used before adding it.

We've kind of opened the discussion on monday but didn't finish it. Essentially, some services require special handling.
At this time its mostly nova, which requires TLS to be enabled. Right now we don't have clear way of distinguishing between nova and standard services. We could try to make some sort of match on the playbook used ... but our customers are expected to define their own, plus it's not really a nice solution. Better be explicit about it.

Alternatively we can just document the fact. But that may cause issues down the line.

PR #819, opened 2024-04-11
Add validator for issuers and cacert secrets

vakwetu:
This adds validation code to confirm that any issuers or cacert secrets that are referenced in a dataplane service actually exist before the dataplane service is created.

This will be required even if tls-e is not enabled on the nodeset.

This is WIP for now as I need to add tests, but I'd like feedback on the approach.

See https://issues.redhat.com/browse/OSPRH-5893 for more discussion on the approach.  If we want to use the deployment webhook instead, it will be easy to move this over.

### DIFF
--- a/api/v1beta1/openstackdataplaneservice_webhook.go
+++ b/api/v1beta1/openstackdataplaneservice_webhook.go
@@ -17,20 +17,58 @@ limitations under the License.
 package v1beta1
 
 import (
+	"context"
+	"fmt"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/certmanager"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


api/v1beta1/openstackdataplaneservice_webhook.go
```diff
@@ -97,9 +187,31 @@ func (r *OpenStackDataPlaneService) ValidateUpdate(original runtime.Object) (adm
...
+	}
 
-	return field.ErrorList{}
+	// Validate cacerts if present
+	if secretName, err := r.ValidateCACerts(); err != nil {
+		openstackdataplaneservicelog.Error(
+			err, "Error validating OpenStackDataPlaneService caCerts secret",
+			"secretName", secretName)
+		errors = append(errors, field.Invalid(
+			field.NewPath("Spec").Child("CACerts"),
```
jpodivin, line 209: We should keep the `Spec` in lower case for sake of consistency. We didn't really do that before, but from now on we should.

jpodivin (commented):
The test suite will need some work. As  will the one for my PR. The webhooks don't seem to mesh well with our functional tests.

vakwetu:
Closing because we don't want to do this now.  We can reopen later if we decide to take on https://issues.redhat.com/browse/OSPRH-5893

PR #821, opened 2024-04-12
Validating dataplane/controlplane TLS consistency

jpodivin:
Webhook verifies that TLS settings for nodeset are consistent with those of existing control plane, if there is one.

adding dependency on openstack-operator

New field for nodeset specifying control plane name.

### DIFF
--- a/api/v1beta1/openstackdataplanenodeset_types.go
+++ b/api/v1beta1/openstackdataplanenodeset_types.go
@@ -17,9 +17,11 @@ limitations under the License.
 package v1beta1
 
 import (
+	"context"
 	"fmt"
 
 	"golang.org/x/exp/slices"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
@@ -289,4 +291,50 @@ func (r *OpenStackDataPlaneNodeSetSpec) duplicateNodeCheck(nodeSetList *OpenStac
 	}
 
 	return
+}
+
+// Compare TLS settings of control plane and data plane
+// if control plane name is specified attempt to retrieve it
+// otherwise get any control plane in the namespace
+func (r *OpenStackDataPlaneNodeSetSpec) ValidateTLS(namespace string, reconcilerClient client.Client, ctx context.Context) error {
+	var err error
+	controlPlanes := openstackv1.OpenStackControlPlaneList{}
+	opts := client.ListOptions{
+		Namespace: namespace,
+	}
+
+	// Attempt to get list of all ControlPlanes fail if that isn't possible
+	if err = reconcilerClient.List(ctx, &controlPlanes, &opts); err != nil {
+		return err
+	}
+	// Verify TLS status of control plane only if there is a single one
+	// report error if there are multiple, or proceed if there are none
+	if len(controlPlanes.Items) > 1 {
+		err = fmt.Errorf("multiple control planes found in the namespace %s", namespace)
+	} else if len(controlPlanes.Items) == 1 {
+		controlPlane := controlPlanes.Items[0]
+		fieldErr := r.TLSMatch(controlPlane)
+		if fieldErr != nil {
+			err = fmt.Errorf("%s", fieldErr.Error())
+		}
+	}
+
+	return err
+}
+
+// Do TLS flags match in control plane ingress, pods and data plane
+func (r *OpenStackDataPlaneNodeSetSpec) TLSMatch(controlPlane openstackv1.OpenStackControlPlane) *field.Error {
+
+	if controlPlane.Spec.TLS.Ingress.Enabled != r.TLSEnabled || controlPlane.Spec.TLS.PodLevel.Enabled != r.TLSEnabled {
+
+		return field.Forbidden(
+			field.NewPath("spec.tlsEnabled"),
+			fmt.Sprintf(
+				"TLS settings on Data Plane node set and Control Plane %s do not match, Node set: %t Control Plane Ingress: %t Control Plane PodLevel: %t",
+				controlPlane.Name,
+				r.TLSEnabled,
+				controlPlane.Spec.TLS.Ingress.Enabled,
+				controlPlane.Spec.TLS.PodLevel.Enabled))
+	}
+	return nil
 }
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -188,6 +188,14 @@ rules:
 - apiGroups:
   - core.openstack.org
   resources:
+  - openstackcontrolplanes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - core.openstack.org
+  resources:
   - openstackversions
   verbs:
   - get
--- a/controllers/openstackdataplanedeployment_controller.go
+++ b/controllers/openstackdataplanedeployment_controller.go
@@ -62,6 +62,7 @@ func (r *OpenStackDataPlaneDeploymentReconciler) GetLogger(ctx context.Context) 
 //+kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=cert-manager.io,resources=issuers,verbs=get;list;watch;
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete;
+//+kubebuilder:rbac:groups=core.openstack.org,resources=openstackcontrolplanes,verbs=get;list;watch;
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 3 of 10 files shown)


jpodivin:
`failed to wait for the created cluster claim to become ready: timed out waiting for the condition `

api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -198,9 +234,63 @@ func (r *OpenStackDataPlaneNodeSetSpec) ValidateUpdate(oldSpec *OpenStackDataPla
...
+	if len(r.ControlPlane) > 0 {
+		controlPlaneName := types.NamespacedName{
+			Name:      r.ControlPlane,
+			Namespace: namespace,
+		}
+		err := webhookClient.Get(context.TODO(), controlPlaneName, &controlPlane)
+		if err != nil {
+			errors = append(errors, field.InternalError(field.NewPath("spec.tlsEnabled"), err))
+		}
+
```
vakwetu, line 250: You forget to check the TLSMatch in case you do find the right control plane!

vakwetu: Ditto above in the Create validate

api/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -198,9 +234,63 @@ func (r *OpenStackDataPlaneNodeSetSpec) ValidateUpdate(oldSpec *OpenStackDataPla
...
+			field.NewPath("spec.tlsEnabled"),
+			fmt.Sprintf(
+				"TLS settings on Data Plane node set and Control Plane %s do not match, Node set: %t Control Plane Ingress: %t Control Plane PodLevel: %t",
+				controlPlane.Name,
+				r.TLSEnabled,
+				controlPlane.Spec.TLS.Ingress.Enabled,
+				controlPlane.Spec.TLS.PodLevel.Enabled))
+	}
+	return nil
+}
```
vakwetu, line 292: I wonder if this should just be a validation.  Or whether we should make this a mutating webhook and match the setting in the control plane.  Thoughts?

jpodivin: Not relevant anymore

jpodivin:
There seems to be an issue with openstack-operator dependency forcing downgrade in the lib-common dependency.
Until they are sync this PR is on hold.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -238,3 +246,56 @@ func (r *OpenStackDataPlaneNodeSetSpec) duplicateNodeCheck(nodeSetList *OpenStac
...
+func (r *OpenStackDataPlaneNodeSetSpec) ValidateTLS(namespace string) error {
+	var err error
+
+	controlPlane := openstackv1.OpenStackControlPlane{}
+	if len(r.ControlPlane) > 0 {
+		controlPlaneName := types.NamespacedName{
+			Name:      r.ControlPlane,
+			Namespace: namespace,
+		}
+		err = webhookClient.Get(context.TODO(), controlPlaneName, &controlPlane)
```
vakwetu, line 262: You've changed the code to call this function from the controller - but you're still using webhook related variables here.  What happens if someone decides to disable webhooks ?  -- you'll end up with an error here.  Instead this should have the ctx and client passed in -- and maybe even should be defined in the controller file itself - like other such functions.

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -238,3 +246,56 @@ func (r *OpenStackDataPlaneNodeSetSpec) duplicateNodeCheck(nodeSetList *OpenStac
...
+		if err == nil {
+			err = r.TLSMatch(controlPlane)
+		}
+
```
vakwetu, line 266: In the case where the control plane is not found, we should requeue and wait, right?

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -238,3 +246,56 @@ func (r *OpenStackDataPlaneNodeSetSpec) duplicateNodeCheck(nodeSetList *OpenStac
...
+		_ = webhookClient.List(context.TODO(), &controlPlanes, &opts)
+		// Verify TLS status of control plane only if there is a single one
+		// report error if there are multiple, or proceed if there are none
+		if len(controlPlanes.Items) > 1 {
+			err = fmt.Errorf("multiple control planes found in the namespace %v", namespace)
+		} else if len(controlPlanes.Items) == 1 {
+			controlPlane = controlPlanes.Items[0]
+			err = r.TLSMatch(controlPlane)
+		}
```
vakwetu, line 281: Again, if there is no control plane , we should requeue and wait, right?

api/v1beta1/openstackdataplanenodeset_types.go
vakwetu: ah good point.  I forgot about the no control plane case.  @slagle  - this is a legit case, right?

api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
```diff
@@ -199,6 +199,8 @@ spec:
                 - ctlplaneInterface
                 - deploymentSSHSecret
                 type: object
+              controlPlane:
+                type: string
```
vakwetu, line 203: Just a thought.  Right now, the expectation is that there is only one control plane per namespace - and its not clear that this will change anytime soon.  For simplicity sake, does it make sense to just assume this is the case for now.  In that case, this code becomes even simpler.

We don't need to add a parameter for the controlPlane.  The logic would be:
-- look for a control plane
  -- if one exists, check tls settings for it
  -- if more than one exists, error out
  -- if none exists, just continue  (this is assuming that deploying a dataplane is possible and makes sense without 
     a control plane)

When we have a requirement for more than one control plane, we can add this parameter?

What do you think @slagle @gibizer @SeanMooney

SeanMooney: makeing crd changes after we release if they are not purely additive will be expensive

in general i think we should keep this simple and be explicit by pass it in by name

if its not present we can either error, look it up or assume we are deploying the dataplane by itself for testing and no tls will be used.

i would prefer to error and make it a required field.

you are correct wew are unlikely to have more the one instance of the conolplane in a given namespaces however i think we would be better served by not having complex logic to try and discover the control plane.
i would add it if for no other reason then we are being consistent with the pattern we have already established where depencies should be passed by name.
we avoid trying to look up things automaitcally.

api/go.mod
```diff
@@ -15,67 +15,89 @@ require (
 require (
 	github.com/cert-manager/cert-manager v1.13.5
 	github.com/go-playground/validator/v10 v10.19.0
-	golang.org/x/exp v0.0.0-20240213143201-ec583247a57a
+	github.com/openstack-k8s-operators/openstack-operator/apis v0.0.0-20240418083003-41031f2a1340
```
gibizer, line 18: This will cause a dependency loop because we have the /api and the operator in the same repo and we use git commit based pseudoversions.

1. bump something in openstack-operator/api
2. dataplane-operator bumps to the new openstack-operator/api, this results in a new pseudoversion of dataplane-operator/api as well as they are in the same repo
3. now there is a need to bump openstack-operator to the new dataplane-operator/api version, but this will lead to a new pseudoversion of openstack-operator/api as well and you go to step 2 again

I afraid this will lead to pain in the future.


So if we want to keep the logic that the dataplane-operator reads the OpenStackVersion but want to break the cycle then we either need to
* stop using pseudoversions
* move the /api module to a separate repo
* move the openstack-operator and dataplane-operator into the same repo and use replace lines in the go.mod files to use the dependency from locally and not from github (like we do in all the main go.mod files towards the /api module already)

jpodivin: That is a problem. I wonder if it would be possible to fashion some sort of quasi ControlPlane struct here and fill it with values obtained by client? I don't know how would that work, but it seems like the only way forward.

gibizer: We agreed on slack to allow the cycle to be added. We will disable automatic version bumping for this dependency in renovate and mid term we intend to merge the two operator repos together

jpodivin: Ok. Let's do that then.

api/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
vakwetu: ^^ @jpodivin @slagle  @gibizer  thoughts?  I don't have any real preference here.  Also, https://github.com/openstack-k8s-operators/openstack-operator/pull/760 landed recently.  Should we be specifying a control plane name or a label?

slagle: > We don't need to add a parameter for the controlPlane. The logic would be: -- look for a control plane -- if one exists, check tls settings for it -- if more than one exists, error out -- if none exists, just continue (this is assuming that deploying a dataplane is possible and makes sense without a control plane)

That is how the check for OpenStackVersion is implemented, and I would be ok with the same for the ControlPlane.

Erroring when it doesn't exist (whether there is a field or not), adds another dependency on the ControlPlane itself, not just openstack-operator. Effectively, we'd need to add the `openstack_deploy` target as a dep for `dataplane_kuttl` in install_yamls.

jpodivin: The field is a good idea, making the connection explicit. It also makes sense since in the future we are going to have even tighter coupling with control plane.

I'm a bit less convinced about making it required. But it is a workable approach.

As for the question of logic. There are couple of options. If we want too keep the coupling loose:

1. Is ControlPlane field set? Yes: Try to get it and use it. Error if it doesn't exist.
2. Is there a 1 control plane in namespace? Yes: use it. No: go 3
3. Is there more than one control plane? Yes: error. No: go 4
4. Proceed without control plane.

If we want it slightly tighter:
1. Is ControlPlane field set? Yes: Try to get it and use it. Error if it doesn't exist.
2. Proceed without control plane.

And finally, the simplest option is to just make the field required and error out if control plane it points to doesn't exist.

We can, hypothetically, proceed from option 1 to 3 with minor modifications of the code.
But all of them require the controlPlane field to be present.

jpodivin:
/recheck

multiple timeouts

api/v1beta1/openstackdataplanenodeset_types.go
jpodivin: ditto

PR #829, opened 2024-04-16
Fail the deployment if nodeset changes while deploying

fao89:


fao89:
recheck

pkg/util/ansible_execution.go
```diff
@@ -64,14 +64,24 @@ func AnsibleExecution(
...
-				Namespace: deployment.GetNamespace(),
-				Labels:    labels,
+				Name:        executionName,
+				Namespace:   deployment.GetNamespace(),
+				Labels:      labels,
+				Annotations: annotations,
 			},
 		}
+	} else if ansibleEE.Annotations["nodesethash"] != annotations["nodesethash"] {
+		return fmt.Errorf("openstackdataplanenodeset %s hash changed while deploying", nodeSet.GetName())
```
fao89, line 84: fail if a nodeset changes in the middle of the deploy

fao89: should we requeue instead?

slagle: a requeue wouldn't re-execute the ansibleee jobs that have already completed for the deployment. so, we'd need to either force that somehow (such as by deleting the openstackansibleee's that had already completed), or just stick with the error.

i think given the existing patterns that we don't re-execute ansible automatically, we probably stick with what you have here and just fail the deployment. if the user wants to re-execute all the ansible with the updated nodeset(s), they need to create a new deployment.

rabi: Why not just reject update to nodeset spec when deployment(s) are running in the webhook?

jpodivin: Is it possible without querying the deployment? We did have a big discussion about webhooks not so long ago and there was considerable opposition to implementing checks relying on other resources than the one the webhook belongs to.

pkg/util/ansible_execution.go
```diff
@@ -64,14 +64,24 @@ func AnsibleExecution(
...
-				Name:      executionName,
-				Namespace: deployment.GetNamespace(),
-				Labels:    labels,
+				Name:        executionName,
+				Namespace:   deployment.GetNamespace(),
+				Labels:      labels,
+				Annotations: annotations,
 			},
 		}
+	} else if ansibleEE.Annotations["nodesethash"] != annotations["nodesethash"] {
```
jpodivin, line 83: The commit message needs to be updated to describe the new behavior. The implementation appears fine, but if you look at history it isn't clear it was added with this patch, or even why.

Which is another thing.

jpodivin (changes requested):

pkg/util/ansible_execution.go
fao89: I like the idea of rejecting the nodeset spec

fao89: the webhook doesn't cover the case where we oc apply the nodeset and deployment together, 
we can try to block nodeset change, but it is good to have this validation as a fallback

jpodivin (commented):
+2

pkg/util/ansible_execution.go
slagle: There is a difference of opinion on whether validations that rely on the state of other resources belong in the webhook or reconcile. I see this as reconcile logic.

If we were working with native k8s concepts of a workload Deployment (and not the EDPM special case with ansible), then webhooks don't even come into the picture. You update a resource, and if a Deployment is already in progress it gets updated in the reconcile, and not blocked at the web layer. EDPM should work the same way, except we only trigger Deployments manually.

Since EDPM is so different, I understand there are different views on how we'd implement the same sort of behavior. To me, I would keep the ux layer the same in that the request is accepted, and then handle it in the reconcile. We'll know if the NodeSet spec has changed (can use either hash or generation), if the deployment sees a change it can either fail, or keep deploying with the old version.

To keep deploying with the old version, we just need to requeue early from the NodeSet reconcile if  we detect a change and a deployment is in progress. Once the in progress deployment finishes the next nodeset reconcile would update the inventory and the NodeSet.Status would reflect that the old version was deployed, and a new Deployment is needed.

In either case, the user then needs to (re)create a deployment to deploy again.

rabi: I think the suggestion here was to use the `deploymentStatuses`  from the old nodeset object in the webhook and not query and use status from any other resource. Webhooks though can be expensive at times (i.e if they make too many api calls), are always used to check if some conditions are met before updating the spec. 

Talking about core k8s concepts, if we can ensure that after updating a nodeset (when a deployment is already running),  deployment happens with the new nodeset spec without any kind of user intervention (like creating a new deployment) then we don't need to block patching resources with webhooks.

fao89:
I started these 2 approaches:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/839
https://github.com/openstack-k8s-operators/dataplane-operator/pull/834

PR #831, opened 2024-04-16
Set default secret in libvirt service

olliewalsh:
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/806

Related: [OSPRH-6172](https://issues.redhat.com//browse/OSPRH-6172)

### DIFF
--- a/config/services/dataplane_v1beta1_openstackdataplaneservice_libvirt.yaml
+++ b/config/services/dataplane_v1beta1_openstackdataplaneservice_libvirt.yaml
@@ -4,6 +4,10 @@ metadata:
   name: libvirt
 spec:
   playbook: osp.edpm.libvirt
+  secrets:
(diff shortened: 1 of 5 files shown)


gibizer (commented):
Looks OK. It seems the install_yamls depends on was picked up the CI as the deployment did not fail on the missing secret form the service.

I sugges to document the need for this secret generation in https://github.com/openstack-k8s-operators/dataplane-operator/blob/0230e251a207684d3a0642487281704db65174d6/docs/assemblies/proc_creating-the-SSH-key-secrets.adoc

olliewalsh:
@gibizer ack, will add docs in a follow-up as this is in a depends-on chain

gibizer (approved):
this make sense to me and the install_yamls dependency has been landed.

PR #832, opened 2024-04-17
Also add domain from hostname to search domains

rabi:
If the hostName is fqdn in nodeset spec, add the domain from it search domains.


rabi:
closed by mistake.

PR #833, opened 2024-04-17
Replacing python yq with go yq

jpodivin:
We are using it in install yamls already.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1520
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/811

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -343,9 +343,9 @@ endif
 
 .PHONY: yq
 yq: ## Download and install yq in local env
-	python -m venv $(LOCALBIN)/.virtualenv && \
-	source $(LOCALBIN)/.virtualenv/bin/activate && \
-	pip install yq
+	test -s $(LOCALBIN)/yq || ( cd $(LOCALBIN) &&\
+	wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64.tar.gz -O - |\
+	tar xz && mv yq_linux_amd64 $(LOCALBIN)/yq )
 
 # A comma-separated list of bundle images (e.g. make catalog-build BUNDLE_IMGS=example.com/operator-bundle:v0.1.0,example.com/operator-bundle:v0.2.0).
 # These images MUST exist in a registry and be pull-able.
--- a/docs/kustomize_to_docs.sh
+++ b/docs/kustomize_to_docs.sh
@@ -1,13 +1,11 @@
 #!/usr/bin/env bash
 set -ex pipefail
 
-source $LOCALBIN/.virtualenv/bin/activate
-
 BAREMETAL=docs/assemblies/ref_example-OpenStackDataPlaneNodeSet-CR-for-bare-metal-nodes.adoc
 FOOTER=$(sed '0,/----/d' $BAREMETAL | sed -e '0,/----/d')
 sed -i '/----/q' $BAREMETAL
 sed -i 's/preprovisioned/baremetal/' examples/no_vars_from/kustomization.yaml
-$KUSTOMIZE --load-restrictor LoadRestrictionsNone build examples/no_vars_from | yq ' select(.kind == "OpenStackDataPlaneNodeSet")' -Y >> $BAREMETAL
+$KUSTOMIZE --load-restrictor LoadRestrictionsNone build examples/no_vars_from | $LOCALBIN/yq ' select(.kind == "OpenStackDataPlaneNodeSet")' >> $BAREMETAL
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -344,9 +344,9 @@ endif
 
 .PHONY: yq
 yq: ## Download and install yq in local env
-	python -m venv $(LOCALBIN)/.virtualenv && \
-	source $(LOCALBIN)/.virtualenv/bin/activate && \
-	pip install yq
+	cd $(LOCALBIN) &&\
+	wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64.tar.gz -O - |\
+	tar xz && mv yq_linux_amd64 $(LOCALBIN)/yq
```
fao89, line 349:
```suggestion
	test -s $(LOCALBIN)/yq || GOBIN=$(LOCALBIN) github.com/mikefarah/yq/v4@latest
```

jpodivin: I don't think this would work. The test is useful yes. But the `GOBIN=$(LOCALBIN) github.com/mikefarah/yq/v4@latest` doesn't actually invoke any command. I assume you meant `go install`, but bear in mind that the v4 of `yq` is using go v1.21. Whereas we are on v1.20.

That's bound to cause issues. By simply downloading the compiled binary we can avoid the compatibility issue, not to mention pulling more dependencies.

fao89 (approved):

Makefile
SeanMooney: i agree with that logic downloading the binary is preferable in this case.

docs/kustomize_to_docs.sh
```diff
@@ -1,13 +1,11 @@
 #!/usr/bin/env bash
 set -ex pipefail
 
-source $LOCALBIN/.virtualenv/bin/activate
-
 BAREMETAL=docs/assemblies/ref_example-OpenStackDataPlaneNodeSet-CR-for-bare-metal-nodes.adoc
 FOOTER=$(sed '0,/----/d' $BAREMETAL | sed -e '0,/----/d')
 sed -i '/----/q' $BAREMETAL
 sed -i 's/preprovisioned/baremetal/' examples/no_vars_from/kustomization.yaml
-$KUSTOMIZE --load-restrictor LoadRestrictionsNone build examples/no_vars_from | yq ' select(.kind == "OpenStackDataPlaneNodeSet")' -Y >> $BAREMETAL
```
SeanMooney, line 10: the -Y by the way is what makes this incompatioable

the python version now required -y or -Y  when modifying a yaml input to specify that
it should render yaml output and if you want to preserve the original order ectra during the transformation

the go version does not support those flags and does not require them

SeanMooney (approved):

docs/kustomize_to_docs.sh
jpodivin: Yeah, that is true.

jpodivin:
zuul failed on dep pull with `HTTP Error 502: Bad Gateway", "status_code": 502, "url": "https://github.com/kudobuilder/kuttl/releases/download/v0.15.0/kubectl-kuttl_0.15.0_linux_x86_64`

jpodivin:
recheck

slagle (approved):

jpodivin:
recheck - `/home/zuul/src/github.com/openstack-k8s-operators/dataplane-operator/bin/controller-gen rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases && \
rm -f api/bases/* && cp -a config/crd/bases api/
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa2d40f]`

fao89 (approved):

jpodivin:
recheck - depends on changed

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck - rebase in dependency.

jpodivin:
recheck - pin with repo now

jpodivin:
/retest 

` Unable to establish connection to https://keystone-public-openstack.apps.oko-08.oooci.ccitredhat.com/v3/auth/tokens: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')) `

PR #834, opened 2024-04-17
Do not update the inventory while deployment is in progress

fao89:
webhook approach: https://github.com/openstack-k8s-operators/dataplane-operator/pull/839

### DIFF
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -384,6 +384,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
+	isDeploymentReady, isDeploymentRunning, isDeploymentFailed, err := checkDeployment(helper, instance)
+	if err != nil {
+		instance.Status.Conditions.MarkFalse(
+			condition.DeploymentReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityError,
+			condition.DeploymentReadyErrorMessage,
+			err.Error())
+		Log.Error(err, "Unable to get deployed OpenStackDataPlaneDeployments.")
+		return ctrl.Result{}, err
+	}
+
+	if isDeploymentRunning {
+		Log.Info("Deployment still running...", "instance", instance)
+		return ctrl.Result{}, nil
+
+	}
+
 	// Generate NodeSet Inventory
 	_, err = deployment.GenerateNodeSetInventory(ctx, helper, instance,
 		allIPSets, dnsDetails.ServerAddresses, containerImages)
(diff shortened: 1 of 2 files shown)


controllers/openstackdataplanenodeset_controller.go
```diff
@@ -417,21 +417,39 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
+				condition.ErrorReason,
+				condition.SeverityError,
+				dataplanev1.DataPlaneNodeSetErrorMessage,
+				errorMsg)
+			return ctrl.Result{}, err
+		}
+	}
+
 	// all setup tasks complete, mark SetupReadyCondition True
 	instance.Status.Conditions.MarkTrue(dataplanev1.SetupReadyCondition, condition.ReadyMessage)
```
slagle, line 454: when this `MarkTrue` is run, it's possible the inventory wasn't updated if `!isDeploymentReady`. That seems wrong. The last deployment could also be failed, and `isDeploymentReady` would be `false` there also, and the inventory wouldn't get regenerated.

what about checking if a deployment is in progress, and if so, then requeue much earlier than this before we reset any conditions or set observedGeneration, etc?

slagle (commented):
this looks reasonable to me now, but do we still want to proceed given https://github.com/openstack-k8s-operators/dataplane-operator/pull/839 merged?

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -417,6 +417,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
+			condition.SeverityError,
+			condition.DeploymentReadyErrorMessage,
+			err.Error())
+		Log.Error(err, "Unable to get deployed OpenStackDataPlaneDeployments.")
+		return ctrl.Result{}, err
+	}
+
+	if isDeploymentRunning {
+		Log.Info("Deployment running, requeueing...", "instance", instance)
+		return ctrl.Result{RequeueAfter: time.Second * 15}, nil
```
rabi, line 434: Though this probably won't be required after the restriction in the webhook, we can still have this if we want. However, re-queuing after 15 seconds, when a deployment is running is probably not ideal.

jpodivin: In deployment we have been using `DeploymentRequeueTime`[0] which is set by default to 15 sec. That already takes care of cases when nodeset doesn't exist or when it doesn't have the right conditions, but that is largely orthogonal to this issue.

I would suggest that this delay is set to some other time. Perhaps turning it into parameter so we can adjust it later, if it turns out our estimates were off.


[0] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanedeployment_types.go#L57

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -417,6 +417,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
+	if err != nil {
+		instance.Status.Conditions.MarkFalse(
+			condition.DeploymentReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityError,
+			condition.DeploymentReadyErrorMessage,
+			err.Error())
+		Log.Error(err, "Unable to get deployed OpenStackDataPlaneDeployments.")
+		return ctrl.Result{}, err
+	}
```
jpodivin, line 430: Test case for this would be a good idea.

jpodivin (commented):
I do have some suggestions, but overall I do think this is a good patch.
+1

controllers/openstackdataplanenodeset_controller.go
fao89: it is already covered in the kuttl test:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/tests/kuttl/tests/dataplane-deploy-no-nodes-test/05-assert.yaml

controllers/openstackdataplanenodeset_controller.go
```diff
@@ -384,6 +384,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
+
+	if isDeploymentRunning {
+		Log.Info("Deployment running, requeueing...", "instance", instance)
+		return ctrl.Result{RequeueAfter: time.Second * time.Duration(deploymentRequeueTime)}, nil
```
rabi, line 401: Just for record, as I mentioned earlier, this `deploymentRequeueTime` default of 15 secs could make sense when deployment has not started deploying services but probably not when a deployment is already deploying services.

controllers/openstackdataplanenodeset_controller.go
jpodivin: ack.

controllers/openstackdataplanenodeset_controller.go
slagle: Given that the NodeSet is already watching the Deployment, I don't think we need to explicitly requeue here at all.

rabi (commented):
/lgtm

PR #852, opened 2024-04-23
Validate fields of OpenstackDataPlaneServiceCert in OpenStackDataPlan…

jpodivin:
…eServiceSpec


api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -42,8 +42,8 @@ type OpenstackDataPlaneServiceCert struct {
 	Issuer string `json:"issuer,omitempty"`
 
 	// KeyUsages to be added to the issued cert
-	// +kubebuilder:validation:Optional
-	KeyUsages []certmgrv1.KeyUsage `json:"keyUsages,omitempty" yaml:"keyUsages,omitempty"`
+	// +kubebuilder:default={"key encipherment","digital signature","server auth"}
+	KeyUsages []certmgrv1.KeyUsage `json:"keyUsages" yaml:"keyUsages"`
```
vakwetu, line 46: Making the default explicit is good.  I thought though, that when you mentioned this tightening of the spec would include turning the Contents field into an enum - with "dnsnames" and "ips" as the two possible options.  That would result in validation errors in the same way that KeyUsages currently does.

api/v1beta1/openstackdataplaneservice_webhook.go
```diff
@@ -123,3 +136,22 @@ func (r *OpenStackDataPlaneServiceSpec) ValidateDelete() field.ErrorList {
...
+		"ips",
+	}
+	for _, val := range r.Contents {
+
+		if !slices.Contains(allowedContents, val) {
+			errs = append(errs, field.Invalid(field.NewPath("spec.tlsCert.keyUsages"),
+				r.KeyUsages,
+				fmt.Sprintf("error validating contents of TLSCert, %s, only valid contents are %v ", val, allowedContents)))
+		}
+	}
```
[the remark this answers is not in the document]
jpodivin: That's because it started is function for keyusages. I wanted to see the test results, before rewriting the comment's, logs etc.

api/v1beta1/openstackdataplaneservice_types.go
jpodivin: Contents is list, not a string, enum wouldn't work there because you need  list a approved values. Hence the `ValidateContents` method.

vakwetu: Yes but we can -- and that what I thought you wanted to do -- redefine Contents as a list of enums - just like keyusages is a list of certmgrv1.KeyUsage.  We can define a new enum of CertificateContents for example.

jpodivin: That isn't really the point. Kubebuilder `enum` [0] should be used in cases when field can have any value out of a list. Not when the field is supposed to be a list of approved values.

Enum validation just doesn't fit our use case here, and we shouldn't try to change the API just to make it fit.
Putting the validation in a function is, imho, a cleaner approach. 
[0] https://book.kubebuilder.io/reference/markers/crd-validation.html

vakwetu: Maybe enum wasn't the right thing here.  My point is that we expect contents to be one of a set of approved values.  KeyUsages (https://github.com/cert-manager/cert-manager/blob/eb9f8880fdf9a02c2f85e73cc08617c350717378/pkg/api/util/usages.go#L26)  is a similar thing, and we can use an equivalent structure to make that explicit - and thereby enforce it in the same way that KeyUsages are enforced.

PR #853, opened 2024-04-23
Add CertsFrom and EDPMServiceName to Service

slagle:
Some services will have different names, but effectively be managing the same
EDPM service, using the same ansible content, and need the same cert data as
another service.

Jira: [OSPRH-6532](https://issues.redhat.com//browse/OSPRH-6532)

### DIFF
--- a/api/v1beta1/openstackdataplaneservice_types.go
+++ b/api/v1beta1/openstackdataplaneservice_types.go
@@ -96,6 +96,10 @@ type OpenStackDataPlaneServiceSpec struct {
 	// ContainerImages struct field names from
 	// github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1
 	ContainerImageFields []string `json:"containerImageFields,omitempty" yaml:"containerImageFields,omitempty"`
+
+	// EDPMServiceName - name to use for edpm_service_name ansible variable
+	// +kubebuilder:validation:Optional
+	EDPMServiceName string `json:"edpmServiceName,omitempty" yaml:"edpmServiceName,omitempty"`
 }
 
 // OpenStackDataPlaneServiceStatus defines the observed state of OpenStackDataPlaneService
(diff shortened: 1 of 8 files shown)


api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -90,6 +96,10 @@ type OpenStackDataPlaneServiceSpec struct {
 	// ContainerImages struct field names from
 	// github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1
 	ContainerImageFields []string `json:"containerImageFields,omitempty" yaml:"containerImageFields,omitempty"`
+
+	// EDPMServiceName - name to use for edpm_service_name ansible variable
+	// +kubebuilder:validation:Optional
+	EDPMServiceName string `json:"edpmServiceName,omitempty" yaml:"edpmServiceName,omitempty"`
```
fao89, line 102: shouldn't it be:
```suggestion
	DataplaneServiceName string `json:"edpmServiceName,omitempty" yaml:"dataplaneServiceName,omitempty"`
```

slagle: well, it already is a DataPlaneService, but with a different name, so I thought calling this field something different (and aligned with the ansible variable) would make sense. But, I'm happy to change it, let's see if there's any other input.

fao89: if it is aligned with the ansible variable, then it is good (it was never a blocker to me, but I wonder when we should use `dataplane` or `EDPM`  for naming things)

fao89 (approved):

vakwetu:
+2   Please also update the tls.doc to document this new functionality. It will be useful to add an example in that doc as well.  https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/docs/assemblies/tls.adoc

PR #854, opened 2024-04-23
expose services list to ansibleee containers

rebtoor:
`edpm_services` will be exposed as inventory variable
`edpm_services_override` will be exposed as ansible extra var

### DIFF
--- a/pkg/deployment/inventory.go
+++ b/pkg/deployment/inventory.go
@@ -127,6 +127,9 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 		nodeSetGroup.Vars["nodeset_tags"] = instance.Spec.Tags
 	}
 
+	// add services list
+	nodeSetGroup.Vars["edpm_services"] = instance.Spec.Services
+
 	nodeSetGroup.Vars["ansible_ssh_private_key_file"] = fmt.Sprintf("/runner/env/ssh_key/ssh_key_%s", instance.Name)
 
 	for _, node := range instance.Spec.Nodes {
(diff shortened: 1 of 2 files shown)


pkg/deployment/inventory.go
```diff
@@ -127,6 +127,9 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 		nodeSetGroup.Vars["nodeset_tags"] = instance.Spec.Tags
 	}
 
+	// add services list
+	nodeSetGroup.Vars["edpm_services"] = instance.Spec.Services
```
fao89, line 131: we need to consider deployment.ServicesOverride

fao89 (approved):

pkg/util/ansible_execution.go
```diff
@@ -124,6 +124,10 @@ func AnsibleExecution(
 		}
 		ansibleEE.Spec.ExtraVars["edpm_service_name"] = json.RawMessage([]byte(fmt.Sprintf("\"%s\"", service.Name)))
 
+		if len(deployment.Spec.ServicesOverride) > 0 {
+			ansibleEE.Spec.ExtraVars["edpm_services_override"] = json.RawMessage([]byte(fmt.Sprintf("\"%s\"", deployment.Spec.ServicesOverride)))
+		}
+
```
slagle, line 130: whenever a variable is set in ansibleEE.Spec.ExtraVars by our code, we should probably first check if it isn't already set, so that the code does not overwrite something set by the user in Deployment.Spec.AnsibleExtraVars which gets passed through as ansibleEE.Spec.ExtraVars.

i'd be fine to merge this as-is though, and we can discuss if that is something we want to check for and apply it to the earlier statements above as well.

jpodivin: In principle we should trust that our customer, if they define the var, knows what they are doing. But I agree that it's not a deal breaker.

jpodivin (commented):
If it's fine with @slagle I think we can merge it.

slagle (approved):

PR #856, opened 2024-04-24
Drop code path to deploy without IPAM and DNS

rabi:
This PR does the following:

- Removes code path for broken deployment without IPAM and DNS and it now errors and requeues if netconfig and dnsmasq service CRs do not exist.
- Refactors the code to not use `dnsdata` to mean different things at different times. `DNSData` type/struct contains the dns records that dnsmasq service uses and not where dns server config and IPAM details are stored for dataplane. Also makes sure all IPAM or DNS functions are together for easy reference. It also changes to not use receivers and creates a DNSDetail struct that's returned from EnsureDNSData().
- Fixes functional tests that used to run without DNS/IPAM and adds tests when there is no netconfig or dns service.
- Fixes kuttl tests that used to run without IPAM/DNS.


pkg/deployment/ipam.go
```diff
@@ -31,74 +32,64 @@ import (
...
-		}
-	}
-	instance.Status.Conditions.MarkTrue(
-		dataplanev1.NodeSetIPReservationReadyCondition,
-		dataplanev1.NodeSetIPReservationReadyMessage)
-	return allIPSets, true, nil
-}
-
-// DataplaneDNSData holds information relevant to the OpenStack DNS configuration of the cluster.
-type DataplaneDNSData struct {
```
jpodivin, line 67: Is changing the struct name and order of fields really necessary? The patch could be substantially smaller/cleaner with just the field addition. Furthermore, I would like to see the comment amended, not removed. Code should be made more readable, not less.

pkg/deployment/ipam.go
```diff
@@ -31,74 +32,64 @@ import (
...
+func checkDNSService(ctx context.Context, helper *helper.Helper,
+	instance client.Object, dnsDetails *DNSDetails,
+) error {
+	dnsmasqList := &infranetworkv1.DNSMasqList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.GetNamespace()),
+	}
+	err := helper.GetClient().List(ctx, dnsmasqList, listOpts...)
+	if err != nil {
+		return err
```
jpodivin, line 60: log line here would be a good idea.

pkg/deployment/ipam.go
```diff
@@ -31,74 +32,64 @@ import (
...
+		util.LogForObject(helper, "DNS control plane service exists, but not ready yet ", instance)
+		return nil
+	}
+	dnsDetails.ClusterAddresses = dnsmasqList.Items[0].Status.DNSClusterAddresses
+	dnsDetails.ServerAddresses = dnsmasqList.Items[0].Status.DNSAddresses
+	return nil
 }
 
 // createOrPatchDNSData builds the DNSData
-func (dns *DataplaneDNSData) createOrPatchDNSData(ctx context.Context, helper *helper.Helper,
```
jpodivin, line 85: Can't this stay as a method of DataplaneDNSData? You still have to use the same fields and the structure remains, it's just inflating size of the patch.

pkg/deployment/ipam.go
```diff
@@ -158,41 +148,39 @@ func (dns *DataplaneDNSData) createOrPatchDNSData(ctx context.Context, helper *h
 }
 
 // EnsureDNSData Ensures DNSData is created
-func (dns *DataplaneDNSData) EnsureDNSData(ctx context.Context, helper *helper.Helper,
```
jpodivin, line 161: method -> function ?

pkg/deployment/ipam.go
```diff
@@ -208,24 +196,50 @@ func (dns *DataplaneDNSData) EnsureDNSData(ctx context.Context, helper *helper.H
...
 		dataplanev1.NodeSetDNSDataReadyCondition,
 		dataplanev1.NodeSetDNSDataReadyMessage)
-	dns.Ready = true
+	dnsDetails.IsReady = true
+	return dnsDetails, nil
+}
 
-	return nil
+// EnsureIPSets Creates the IPSets
+func EnsureIPSets(ctx context.Context, helper *helper.Helper,
```
jpodivin, line 218: Why are you moving this?

pkg/deployment/ipam.go
rabi: DNSData means[1] something else and it's confusing to use DataplaneDnsData along with it in the same code. The comment is partially incorrect (does not have DNS configuration, it's just a struct used internally for convience) and not an exposed type.

As per why not using reciever, it only makes sense to use a receiver if the function is part of an interface implementation.

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/dnsdata_types.go#L68

pkg/deployment/ipam.go
rabi: It's arranging the DNS and IPSet functions be at one place. Not mix them like before.

pkg/deployment/ipam.go
rabi: https://grisha.org/blog/2016/09/22/golang-receiver-vs-function/

rabi:
recheck

pkg/deployment/ipam.go
jpodivin: Where are we going to stub this function? Or are we going to do that in the future? Other than that the article doesn't have much in a way of argument. It even admits the result looks worse:

> the only benefit you are getting from using a method is syntactic sweetness.

jpodivin: It provides only one argument, not valid in our case, as to why it may be better to use functions. And then it says the result looks worse if you do. In my book that's not exactly an endorsement of the "function first" approach. 

Also, I do consider populating struct fields to be a type of operation. But that may just be me.

rabi:
recheck

jpodivin:
Could you at least split this into two patches? One for refactor and other for the stated change `dropping IPAM and DNS deploy path`

rabi:
> Could you at least split this into two patches? One for refactor and other for the stated change `dropping IPAM and DNS deploy path`

I don't think they're separate. Both refactoring and  code removal are touching the same files/functions and also those require changes to the tests (both functional and kuttl). Hence all of them are in the same PR. I've updated the commit message to reflect all changes. Normally folks create multiple commits in the same PR and that's reviewed together. I prefer a single commit.

bshephar (approved):
Personally, I prefer receiver functions as I think it makes things easier to follow and work with. But that's a personal preference and not a functional complaint about this code. Since it seems we have differing opinions on this, I would suggest we don't block this PR over it, and rather we all have that debate elsewhere and decide on how we move forward with this design detail in the future.

Functionality of the code looks good, and I like that we're taking this IPAM only path rather than trying to mix the functionality which had the potential to cause confusion and possible issues. It's a +1 from me here, with a suggestion to follow up on the method v function debate.

PR #758, opened 2024-03-13
Improve field alignment of structs

bshephar:
This change improves the field alignment in our structs to reduce the overall memory footprint required for each struct. We do this by aligning like types within the structs to reduce the amount of padding required to align each object in memory.

It's a simple re-ordering of objects with no functional change to how the operator works from the user perspective. See the explanation provided here regarding maligned structs:
https://medium.com/@sebassegros/golang-dealing-with-maligned-structs-9b77bacf4b97

PR #857, opened 2024-04-24
Remove config hashes from kuttl tests

rabi:
We changed order of fields in the structs[1] that would change these hashes. As we did not test the PR in prow kuttl job tests did not fail. Now we need to remove these so that we can test with PRs and merge [2].

Dropping `dataplane-deploy-nodehashes-test` as we've to remove the hashes and there is no place we're asserting that hashes are different after the edit. The test can be reworked and added later.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/758
[2] https://github.com/openshift/release/pull/51299


tests/kuttl/tests/dataplane-deploy-nodehashes-test/00-assert.yaml
```diff
@@ -7,8 +7,6 @@ spec:
   - edpm-compute-no-nodes
 status:
   deployed: true
-  nodeSetHashes:
-    edpm-compute-no-nodes: n595h74h98h5d8h579h5c7h84h5dh599h544hf8h65bh686h5f7h99h669h676h97h678h646h9fh5cch649h66fhcch5fh688h95h7bhbh675hd9q
 ---
```
slagle, line 10: I pushed a commit to also drop the above to your branch, hope you don't mind. that should fix the kuttl tests here hopefully.

slagle (approved):

rabi:
Re-adding the LGTM label as not many around this time.

PR #858, opened 2024-04-25
Show deploy name/condition in error message when nodeset update blocked

abays:
The webhook validation that prevents updates to `OpenStackDataPlaneNodeSet` when an associated `OpenStackDataPlaneDeployment` is running needs more detail in its error message.  For instance, we are currently seeing that VA1 is having trouble with this webhook but we don't know which `OpenStackDataPlaneDeployment` is blocking the update nor which particular condition was problematic.  This PR adds the name of the deployment and the bad condition to the error message to assist in debugging.


bshephar:
This will fix the issue:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/859

But if we want to merge that, then rebase this change on top of it to ensure we update that message. Just note that my PR adds functional tests and we'll need to update the `ContainsSubstring()` check. We can probably simply the check to be less explicit.

bshephar (approved):

PR #859, opened 2024-04-26
Allow NodeSet updates when associated Deployment is Ready

bshephar:
Ensure NodeSets can accept valid updated when Deployment is Complete

DeploymentStatuses never get the final Ready condition for the deployment as we're
only looking at a specific nodeset to be deployment ready. Use NodeSetDeploymentReadyCondition
which we set explicitly.

Also we should not wait for the whole deployment that could have multiple nodesets.
We're concerned about a specific nodeset here.

Resolves: https://issues.redhat.com/browse/OSPRH-6611


rabi:
Can we please update the commit message as in https://github.com/openstack-k8s-operators/dataplane-operator/pull/860

rabi (commented):
/lgtm

PR #860, opened 2024-04-26
Check for NodeSetDeploymentReadyCondition in deploymentStatuses

rabi:
DeploymentStatuses never get the final Ready condition for the deployment as we're only looking at a specific nodeset to be deployment ready. Use NodeSetDeploymentReadyCondition which we set explicitly.

Also we should not wait for the whole deployment that could have multiple nodesets. We're concerned about a specific
nodeset here.


rabi:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/859 was merged.

PR #861, opened 2024-04-26
Remove `deployed` from nodeset status

rabi:
We don't set it after reconciliation. We check Ready condition to understand that nodeset has been deployed. Also this field is set as optional which now omits it from the status.


slagle (commented):
+2, Looks like we stopped using this when we added the Deployment CRD, so lgtm to remove it.

bshephar (approved):

PR #868, opened 2024-04-29
Use BaremetalSet images from OpenStackVersion

rabi:


dprince (commented):
This looks good. Should we do a similar approach for the Apache, and Agent images as well? https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/controllers/openstackbaremetalset_controller.go#L492

api/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -245,6 +246,7 @@ func SetupDefaults() {
 			CeilometerIpmiImage:           getImageDefault("RELATED_IMAGE_CEILOMETER_IPMI_IMAGE_URL_DEFAULT", ContainerImageDefaults.CeilometerIpmiImage),
 			NovaComputeImage:              getImageDefault("RELATED_IMAGE_NOVA_COMPUTE_IMAGE_URL_DEFAULT", ContainerImageDefaults.NovaComputeImage),
 			OvnControllerImage:            getImageDefault("RELATED_IMAGE_OVN_CONTROLLER_AGENT_IMAGE_URL_DEFAULT", ContainerImageDefaults.OvnControllerImage),
+			OsContainerImage:              getImageDefault("RELATED_IMAGE_OVN_CONTROLLER_AGENT_IMAGE_URL_DEFAULT", ContainerImageDefaults.OsContainerImage),
```
slagle, line 249: I think you need to add a new env var to config/default/manager_default_images.yaml and use it here.

rabi: I think they're already in OBMO default images, so we don't have to add them here. https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/config/default/manager_default_images.yaml#L14

bshephar (approved):
lgtm

PR #872, opened 2024-05-03
Removing duplicated import in the test module

jpodivin:
This isn't really anything but a stylistic adjustment. Instead of importing twice, we will do it once.
Functionally the code stays the same. I've decided to go with the `dataplanev1` to make source of the code more obvious.


fao89 (approved):
/unhold

PR #877, opened 2024-05-07
Fix default image name for Apache

abays:
`current-podified` tag does not exist in `registry.redhat.io/ubi9/httpd-24`.  I don't know if `latest` is the proper alternative, but I am proposing it here since it is available in the registry.


fao89 (approved):
I would say to use `1-321`, but:
image

fao89:
Actually, there is the `1` tag:
image

```
docker run -it registry.redhat.io/ubi9/httpd-24:1 bash
Unable to find image 'registry.redhat.io/ubi9/httpd-24:1' locally
1: Pulling from ubi9/httpd-24
6b20e0c6c1e1: Pull complete 
2909ba4f32f6: Pull complete 
b6b37972c974: Pull complete 
Digest: sha256:80f1282b07b094831416c68373242d1fa17e118f39b190b95a47818cd28ff208
Status: Downloaded newer image for registry.redhat.io/ubi9/httpd-24:1
bash-5.1$ 
```

PR #883, opened 2024-05-13
Ensure required lists have one item at least

fao89:


gibizer (commented):
Looks OK to me. I don't think this is a beta blocker so we don't need to land it during the freeze.

jpodivin (commented):
Reasonable change. But I'm setting hold on this to until the beta is out.

jpodivin (approved):

PR #884, opened 2024-05-15
[tls] use cert duration/renewbefore from issuer annotation

stuggi:
use cert duration and renewBefore from annotations set on issuer
- if no duration annotation is set, use the default from certmanager lib-common module,
- if no renewBefore annotation is set, the cert-manager default is used.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/506


vakwetu (approved):

stuggi:
rebased

PR #886, opened 2024-05-17
Use dnf instead of rpm

bshephar:
This change updates the hotfix docs to use the dnf command instead of rpm directly

### DIFF
--- a/docs/assemblies/proc_hotfixing-the-data-plane-rpm-content.adoc
+++ b/docs/assemblies/proc_hotfixing-the-data-plane-rpm-content.adoc
@@ -34,10 +34,9 @@ Repeat this step for each data plane node that the hotfix must be applied to.
 +
 ----
 $ ssh <ssh_user>@<data_plane_node>
-$ sudo rpm -F /tmp/<hotfix_id>/rpms/*.rpm
+$ sudo dnf in -y /tmp/<hotfix_id>/rpms/*.rpm
 ----
 +
-* The `-F` (`--freshen`) `rpm` command option updates the RPM content if earlier versions are already installed.
 * Replace `<ssh_user>` with the SSH user name.
 * Replace `<data_plane_node>` with the hostname or IP for the data plane node.
 * Replace `<hotfix_id>` with a hotfix identifier such as a Jira issue, for example `osprh-0000`.


docs/assemblies/proc_hotfixing-the-data-plane-rpm-content.adoc
```diff
@@ -34,10 +34,9 @@ Repeat this step for each data plane node that the hotfix must be applied to.
 +
 ----
 $ ssh <ssh_user>@<data_plane_node>
-$ sudo rpm -F /tmp/<hotfix_id>/rpms/*.rpm
+$ sudo dnf in -y /tmp/<hotfix_id>/rpms/*.rpm
```
slagle, line 37: is there a specific reason why? are you expecting hotfix rpms to need to be able to resolve deps and perhaps install other updates? It seems to me that would be something we would *not* want to do. I'm pretty sure we have documented using rpm in past releases.

jpodivin (changes requested):

.github/workflows/docs.yaml
```diff
@@ -30,6 +30,8 @@ jobs:
 
       - name: Install Asciidoc
         run: make docs-dependencies
+      - name: Install ruby-dev
+        run: sudo apt install ruby-dev -y
```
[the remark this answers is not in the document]
bshephar: Yeah, I'll drop it. It ended up being fixed in Prow and not necessary for the github action:
https://github.com/openshift/release/pull/52224

docs/assemblies/proc_hotfixing-the-data-plane-rpm-content.adoc
bshephar: It just aligns with RHEL documented processes for installing and managing packages. AFAIK there's no RHEL recommendations to use rpm commands directly. I think it's probably less confusing for users if we align our recommendations with those used in RHEL documentation.

If there _are_ dependencies, they should probably be installed right? Would we expect that an RPM with dependencies would even work if the dependencies were to be skipped?

The main objective was to align with recommendations for RHEL. But if we feel there's a compelling case to not do that, we can drop this PR.

jpodivin: I think that this rationale for the change is valid. After all, the entire move to asciidocs was performed to harmonize docs across RH products. It makes sense to continue in that direction.
That being said, in the interest of clarity and for posterity, it may be better to explicitly state this in the commit message.

slagle: > The main objective was to align with recommendations for RHEL. But if we feel there's a compelling case to not do that, we can drop this PR.

I think that explains it, so I'm ok with it. Are there downstream CI considerations that need to be updated as well, or was this all manually tested?

slagle (approved):

docs/assemblies/proc_hotfixing-the-data-plane-rpm-content.adoc
bshephar: Manually tested with a random RPM. As far as I'm aware, there's currently no downstream jobs that test this process yet.

PR #896, opened 2024-05-23
Add EDPMServiceType, Drop EDPMServiceName

slagle:
Adds a top level EDPMServiceType field and drops EDPMServiceName.
EDPMServiceType is used to construct the volume mount paths for TLS
certs and CA certs, and should correspond to the ansible role name used
to managing the service. When not set, the service name will be used
instead.

EDPMRoleServiceName is kept in case there is a need to override the path
for a specific TLS cert.

Jira: https://issues.redhat.com/browse/OSPRH-7112

### DIFF
--- a/api/v1beta1/openstackdataplaneservice_types.go
+++ b/api/v1beta1/openstackdataplaneservice_types.go
@@ -50,8 +50,8 @@ type OpenstackDataPlaneServiceCert struct {
 	// the edpm-ansible role where this certificate is used. For example if the
 	// certificate is for edpm_ovn from edpm-ansible, EDPMRoleServiceName must be
 	// ovn, which matches the edpm_ovn_service_name variable from the role.  If
-	// not set, OpenStackDataPlaneService.Spec.EDPMServiceName is used. If
-	// OpenStackDataPlaneService.Spec.EDPMServiceName is not set, then
+	// not set, OpenStackDataPlaneService.Spec.EDPMServiceType is used. If
+	// OpenStackDataPlaneService.Spec.EDPMServiceType is not set, then
 	// OpenStackDataPlaneService.Name is used.
 	EDPMRoleServiceName string `json:"edpmRoleServiceName,omitempty"`
 }
@@ -107,9 +107,12 @@ type OpenStackDataPlaneServiceSpec struct {
 	// github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1
 	ContainerImageFields []string `json:"containerImageFields,omitempty" yaml:"containerImageFields,omitempty"`
 
-	// EDPMServiceName - name to use for edpm_service_name ansible variable
-	// +kubebuilder:validation:Optional
-	EDPMServiceName string `json:"edpmServiceName,omitempty" yaml:"edpmServiceName,omitempty"`
+	// EDPMServiceType - service type, which typically corresponds to one of
+	// the default service names (such as nova, ovn, etc). Also typically
+	// corresponds to the ansible role name (without the "edpm_" prefix) used
+	// to manage the service. If not set, will default to the
+	// OpenStackDataPlaneService name.
+	EDPMServiceType string `json:"edpmServiceType,omitempty" yaml:"edpmServiceType,omitempty"`
 }
 
 // OpenStackDataPlaneServiceStatus defines the observed state of OpenStackDataPlaneService
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -280,10 +280,8 @@ func (d *Deployer) addCertMounts(
 				},
 			}
 			certMountDir := service.Spec.TLSCert.EDPMRoleServiceName
-			if certMountDir == "" && service.Spec.EDPMServiceName != "" {
-				certMountDir = service.Spec.EDPMServiceName
-			} else {
-				certMountDir = service.Name
+			if certMountDir == "" {
+				certMountDir = service.Spec.EDPMServiceType
 			}
 			certVolumeMount := corev1.VolumeMount{
 				Name:      GetServiceCertsSecretName(d.NodeSet, service.Name, 0),
@@ -315,7 +313,7 @@ func (d *Deployer) addCertMounts(
 
 			cacertVolumeMount := corev1.VolumeMount{
 				Name:      fmt.Sprintf("%s-%s", service.Name, service.Spec.CACerts),
-				MountPath: path.Join(CACertPaths, service.Name),
+				MountPath: path.Join(CACertPaths, service.Spec.EDPMServiceType),
 			}
 
 			volMounts.Volumes = append(volMounts.Volumes, cacertVolume)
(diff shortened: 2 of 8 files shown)


pkg/deployment/deployment.go
```diff
@@ -280,8 +280,8 @@ func (d *Deployer) addCertMounts(
 				},
 			}
 			certMountDir := service.Spec.TLSCert.EDPMRoleServiceName
-			if certMountDir == "" && service.Spec.EDPMServiceName != "" {
-				certMountDir = service.Spec.EDPMServiceName
+			if certMountDir == "" && service.Spec.EDPMServiceType != "" {
+				certMountDir = service.Spec.EDPMServiceType
 			} else {
```
karelyatin, line 285: shouldn't it be else if used here? as even with EDPMRoleServiceName set else condition would get executed.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -107,9 +107,11 @@ type OpenStackDataPlaneServiceSpec struct {
...
 	ContainerImageFields []string `json:"containerImageFields,omitempty" yaml:"containerImageFields,omitempty"`
 
-	// EDPMServiceName - name to use for edpm_service_name ansible variable
-	// +kubebuilder:validation:Optional
-	EDPMServiceName string `json:"edpmServiceName,omitempty" yaml:"edpmServiceName,omitempty"`
+	// EDPMServiceType - service type, which typically corresponds to one of
+	// the default service names (such as nova, ovn, etc). Also typically
+	// corresponds to the ansible role name (without the "edpm_" prefix) used
+	// to manage the service.
+	EDPMServiceType string `json:"edpmServiceType,omitempty" yaml:"edpmServiceType,omitempty"`
```
gibizer, line 114: As far I see all the usage of EDPMServiceType considers the service.Name as a default value if no explicit value is set. If so then I think we can simplify the implementation by doing such defaulting logic in the defaulting webhook. That would allow us to remove a couple of condition in the controller implementation.

pkg/deployment/deployment.go
SeanMooney: it depends.
yes if we intend to keep EDPMRoleServiceName but I'm not sure that is the intent.

at least I'm not convinced we should but it would not be incorrect to make this an elif for now

karelyatin: yes i was just referring current scenario on how these used currently as it doesn't match what described in the comments.

api/v1beta1/openstackdataplaneservice_types.go
slagle: I've switched it to use a webhook

pkg/deployment/deployment.go
slagle: i fixed the if/else in the webhook change i made.

For now, I have kept EDPMRoleServiceName, I think we can drop that in a future PR if we want. But for now, it's the only way to override the mount dir for a specific cert (if needed).

api/v1beta1/openstackdataplaneservice_types.go
SeanMooney: so now that we have this field can we update the EDPMServiceType for all of the standard dataplane services.

i.e. set it to nova in https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml

that will enable this to work transparently for install yamls since it is just modifying the default one to generate the nova-custom one but for ci frame work since they start form scratch wee will still need to update there definition.

SeanMooney: we could also do that in a followup if we want to keep this small and move forward with it as is

slagle:
The rdo job failed due to the ovn-metadata-agent container not starting:

https://logserver.rdoproject.org/96/896/826d7d0e54d46a853f4bc6f4ef6f986aaf97a3c9/github-check/podified-multinode-edpm-deployment-crc/9895ed8/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/neutron-metadata-edpm-deployment-openstack-edpm-ipam-4rj9l/logs/openstackansibleee.log

`
[1;35m/var/lib/openstack/certs/neutron_metadata/tls.crt: no such file or directory[0m
[0;31mfatal: [compute-0]: FAILED! => {"changed": false, "msg": "Failed containers: ovn_metadata_agent"}[0m
`

It's trying to use `/var/lib/openstack/certs/neutron_metadata/tls.crt` with neutron_metadata with underscore instead of dash. I don't see where the underscore is coming from as of now. 

It should be using dash https://github.com/openstack-k8s-operators/edpm-ansible/blob/3fa954c66dc7c607b7907de7fe3d3b69fedb1b45/roles/edpm_neutron_metadata/defaults/main.yml#L5 And I don't see that var overridden anywhere.

SeanMooney:
i was wondering if that was the case but https://github.com/openstack-k8s-operators/edpm-ansible/pull/658 merged 5 days ago so it should be on quay.io

i wonder is this a regression due to the upgrade support.
i.e. are we not using latest by default any more?

the ci job should be pulling the latest images and building a new bundel

karelyatin:
> > The rdo job failed due to the ovn-metadata-agent container not starting:
> > https://logserver.rdoproject.org/96/896/826d7d0e54d46a853f4bc6f4ef6f986aaf97a3c9/github-check/podified-multinode-edpm-deployment-crc/9895ed8/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/neutron-metadata-edpm-deployment-openstack-edpm-ipam-4rj9l/logs/openstackansibleee.log
> > `�[1;35m/var/lib/openstack/certs/neutron_metadata/tls.crt: no such file or directory�[0m �[0;31mfatal: [compute-0]: FAILED! => {"changed": false, "msg": "Failed containers: ovn_metadata_agent"}�[0m`
> > It's trying to use `/var/lib/openstack/certs/neutron_metadata/tls.crt` with neutron_metadata with underscore instead of dash. I don't see where the underscore is coming from as of now.
> > It should be using dash https://github.com/openstack-k8s-operators/edpm-ansible/blob/3fa954c66dc7c607b7907de7fe3d3b69fedb1b45/roles/edpm_neutron_metadata/defaults/main.yml#L5 And I don't see that var overridden anywhere.
> 
> It's using too old runner image and that's why neutron_metadata being seen https://github.com/openstack-k8s-operators/edpm-ansible/blob/b8753605dbb1caeb8f38a91cd3a99f7e01afa01e/roles/edpm_neutron_metadata/defaults/main.yml#L60

Ok this was an old run, latest run should be good since image is now updated with https://github.com/openstack-k8s-operators/openstack-operator/commit/3069cfac53b767125df30bd2cdf2d03bf92c489f

api/v1beta1/openstackdataplaneservice_types.go
gibizer: thanks for the webhook change, looks good.

viroel:
Since dependencies already merged, we need https://github.com/openstack-k8s-operators/openstack-operator/pull/811 in, so the job gets the updated dataplane-operator.

PR #901, opened 2024-05-28
Add missing related images

fao89:


config/default/manager_default_images.yaml
```diff
@@ -29,3 +31,15 @@ spec:
           value: quay.io/prometheus/node-exporter:v1.5.0
         - name: RELATED_IMAGE_EDPM_MULTIPATHD_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-multipathd:current-podified
+        - name: RELATED_IMAGE_AGENT_IMAGE_URL_DEFAULT
+          value: quay.io/openstack-k8s-operators/openstack-baremetal-operator-agent:current-podified
+        - name: RELATED_IMAGE_APACHE_IMAGE_URL_DEFAULT
```
rabi, line 36: Why do we want to add agent/apache default images in this operator? They're in openstack-baremetal-operator[1].

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/config/default/manager_default_images.yaml#L14-L21

config/default/manager_default_images.yaml
```diff
@@ -29,3 +31,15 @@ spec:
...
+          value: registry.redhat.io/ubi9/httpd-24:latest
+        - name: RELATED_IMAGE_CEILOMETER_COMPUTE_IMAGE_URL_DEFAULT
+          value: quay.io/podified-antelope-centos9/openstack-ceilometer-compute:current-podified
+        - name: RELATED_IMAGE_CEILOMETER_IPMI_IMAGE_URL_DEFAULT
+          value: quay.io/podified-antelope-centos9/openstack-ceilometer-ipmi:current-podified
+        - name: RELATED_IMAGE_NOVA_COMPUTE_IMAGE_URL_DEFAULT
+          value: quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified
+        - name: RELATED_IMAGE_OS_CONTAINER_IMAGE_URL_DEFAULT
```
rabi, line 44: Same for this one.

config/default/manager_default_images.yaml
fao89: [getImageDefault](https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanenodeset_types.go#L236) is currently not finding these vars, therefore we are using the fallbacks: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/openstackdataplanenodeset_types.go#L205-L208
I believe by the way the CSV is built, it only gets the vars that are set on the current operator (the only exception is the openstack-operator which gets the related images from all the operators)

fao89: actually, the OpenstackVersion should catch it, I'll reprovision my env to test

rabi: Hmm.. There is no dataplane-operator csv, datapplane-controller-manager is part of openstack-operator csv and the images should be part of the ENV I think.

fao89: it is the same CSV, but the dataplane-operator does not get the same vars: https://gist.github.com/fao89/187b1b61545a5ab340e599f7fd419282
but the OpenstackVersion gets the right values: https://gist.github.com/fao89/14acb4057665f4dbe64b8704e4d6cd6a

PR #902, opened 2024-05-28
 Add data-plane-adoption job to the zuul dataplane operator github-check layout

marios:
This adds cifmw-data-plane-adoption-osp-17-to-extracted-crc so we get
adoption coverage in github-check.

https://issues.redhat.com/browse/OSPRH-5770

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -4,5 +4,15 @@
     github-check:
       jobs:
         - dataplane-operator-docs-preview
+        - openstack-k8s-operators-content-provider
+        - cifmw-data-plane-adoption-osp-17-to-extracted-crc:
+            dependencies:
+              - openstack-k8s-operators-content-provider
+            files:
+              - ^api/*
+              - ^config/*
+              - ^controllers/*
+              - ^pkg/*
+
     templates:
       - podified-multinode-edpm-baremetal-pipeline


slagle (changes requested):
I have concerns with adding this job, which takes 2.5 to 3 hours. Originally, our goal was 1 hour for check/gate jobs. We've grown to 1.5 hours organically with the crc jobs, and now up to 3 hours if this is added. This has a big effect on our velocity and estimation of how long it takes to get to dev complete.

I'd like to see further discussion on this.

zuul.d/projects.yaml
```diff
@@ -6,3 +6,4 @@
         - dataplane-operator-docs-preview
     templates:
       - podified-multinode-edpm-baremetal-pipeline
+      - data-plane-adoption-pipeline
```
marios, line 9: as discussed in yesterday's adoption call:

I'll add one of the jobs here explicitly instead of adding the data-plane-adoption-pipeline template. That way we will only run one job instead of both (the job with ceph, non tls will run here) and we can restrict when this job will run using files:

marios:
> I have concerns with adding this job, which takes 2.5 to 3 hours. Originally, our goal was 1 hour for check/gate jobs. We've grown to 1.5 hours organically with the crc jobs, and now up to 3 hours if this is added. This has a big effect on our velocity and estimation of how long it takes to get to dev complete.
> 
> I'd like to see further discussion on this.



> I have concerns with adding this job, which takes 2.5 to 3 hours. Originally, our goal was 1 hour for check/gate jobs. We've grown to 1.5 hours organically with the crc jobs, and now up to 3 hours if this is added. This has a big effect on our velocity and estimation of how long it takes to get to dev complete.
> 
> I'd like to see further discussion on this.

as discussed yesterday, yes these jobs do take longer. It is currently around 2.5 hours (build history at [1] for convenience and the 3 hours applies to the new multinode job).

As mentioned yesterday I'll add just one job (not both) and we can target it with files so it will only run when necessary.

zuul.d/projects.yaml
```diff
@@ -4,5 +4,15 @@
...
         - dataplane-operator-docs-preview
+        - openstack-k8s-operators-content-provider
+        - cifmw-data-plane-adoption-osp-17-to-extracted-crc:
+            dependencies:
+              - openstack-k8s-operators-content-provider
+            files:
+              - ^api/*
+              - ^config/*
+              - ^controllers/*
+              - ^pkg/*
```
slagle, line 15: this is all of the code, so using files doesn't really help here. I don't think we want a 2.5 hour job on this repo.

marios: well there is also /docs /examples /hack ;)

can we further target this though? Looking under those directories it was not clear to me that we could.

We need to check if changes proposed to this operator will break our understanding of the adoption workflow (to this point in time), as codified in the adoption tests run in this job.

slagle: > We need to check if changes proposed to this operator will break our understanding of the adoption workflow (to this point in time), as codified in the adoption tests run in this job.

I don't think that running it on every PR matches some of the earlier agreed upon context setting for which jobs we should run where. Perhaps we can revisit those discussions, but I'd like it to be more broadly understood so there's some acknowledgement about the impact of this change.

marios: the proposal is that it will not run on changes to /tests or /docs or the other directories.
as previously mentioned I didn't think we could make it more specific than I posted. However any coverage is better than no coverage so... can you propose an improvement to the files: list. Can/should we consider removing any of those? Can we be more specific with files and sub-directories? 

The idea is that we are constantly evolving our understanding of the adoption workflow, when new services are added (earlier on) or with new breaking features like the UpdateVersion resource and the latest work with OSPCIX-308. We have to update our "implementation" of the adoption workflow codified in the /tests/ subdirectory of the data-plane-adoption repo (and run in the adoption ci jobs).

It would be best to catch these breaking changes as close to the source as possible, which means gating on dataplane-operator and edpm-ansible.

We currently run adoption jobs in data-plane-adoption, ci-framework, install_yamls and openstack-operator repos. It seems to me that the omission of the actual dataplane related code is wrong.

slagle: We've agreed to hold this PR for now, and then when dataplane-operator is merged into openstack-operator it will be covered by the adoption integration job. We can continue to refine with files/etc once that merge is complete.

marios: ok. for right now and as also agreed, we will proceed with the edpm-ansible so at least that is more than nothing :D https://github.com/openstack-k8s-operators/edpm-ansible/pull/662

Closing this pull.

PR #909, opened 2024-05-30
Add DataSources to OpenStackDataPlaneService

slagle:
Add DataSources to OpenStackDataPlaneService

The DataSource struct is added to OpenStackDataPlaneService as an
additional way to specify ConfigMaps and Secrets.

This allows for specifying additional properties of the ConfigMaps and
Secrets, and for using the optional field on ConfigMapRef and SecretRef.
Deploying the service will no longer fail if the ConfigMap or Secret
doesn't exist when optional is true.

The existing ConfigMaps and Secrets fields are kept with the same
functionality for now for backwards compatibility with CI. Once CI
resources are updated to use OpenStackDataPlaneService.Spec.DataSources,
then these fields may be dropped.

AnsibleVarsFrom is also refactored to use the common DataSource code.


pkg/deployment/deployment.go
```diff
@@ -334,15 +334,23 @@ func (d *Deployer) addCertMounts(
...
+	for _, serviceCM := range service.Spec.ConfigMaps {
 
 		volMounts := storage.VolMounts{}
 		cm := &corev1.ConfigMap{}
-		err := client.Get(d.Ctx, types.NamespacedName{Name: cmName, Namespace: service.Namespace}, cm)
-		if err != nil {
+		err := client.Get(d.Ctx, types.NamespacedName{Name: serviceCM.Name, Namespace: service.Namespace}, cm)
+		if err != nil && k8s_errors.IsNotFound(err) {
+			if serviceCM.Required {
+				return d.AeeSpec, err
```
gibizer, line 348: OK this looks good. If a required CM is missing when we start the deployment the deployment will fail. That still allows creating the service first with a CM name, then later generating the CM itself, and finally starting the deployment. This is needed as the nova service might be created earlier than the control plane generating the cell config secret for it.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -56,15 +56,37 @@ type OpenstackDataPlaneServiceCert struct {
...
+// OpenStackDataPlaneSecret defines the properties of the Secrets used by
+// the service.
+type OpenStackDataPlaneSecret struct {
+	// Name of the Secret
+	Name string `json:"name"`
+
+	// Required is whether the Secret is required to exist
+	// +kubebuilder:validation:Optional
+	Required bool `json:"required,omitempty"`
+}
```
gibizer, line 79: As a follow up I would try to validate in the webhook that a service with EDPMServiceType=nova:
* runs the nova playbook
* has the cell secret as required
  * nova-operator generates the secret with a name that is based on the name of the cell. But knowing the cell names require looking up the OpenStackControlPlane CR or more likely the NovaCell CR which would be an unwanted dependency. 
* has the ssh migration key secret as required. This is even more problematic as the name of the secret does not matter for the edpm-ansible code. The only real requirement is to have an ssh keypair with filenames (secret keys) ssh-privatekey and ssh-publickey provided.

Maybe this is still impractical to do these kind of validations :/ I guess I need to think more how can we express the needs of the edm-nova role during the service creation / service deployment.  Anyhow this is not a real comment on the current struct proposal so feel free to ignore it.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -56,15 +56,37 @@ type OpenstackDataPlaneServiceCert struct {
 	EDPMRoleServiceName string `json:"edpmRoleServiceName,omitempty"`
 }
 
+// OpenStackDataPlaneConfigMap defines the properties of the ConfigMaps used
+// by the service.
+type OpenStackDataPlaneConfigMap struct {
```
gibizer, line 61: One thing we can think about is if we want to create something that is more similar how the volume in k8s expresses the CM/Secret sources. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#configmapvolumesource-v1-core

Copying that patter would mean something like:
```golang

// this type needs a validation that only one of the fields can be non nil
type ServiceDataSource struct {
    ConfigMap *ConfigMapRef
    Secret *SecretRef
}

type ConfigMapRef struct {
    Name string
    Optional bool
}

type SecretRef struct {
    Name string
    Optional bool
}
//...

type OpenStackDataPlaneServiceSpec struct {
    // then we can have a single list instead of two lists
    dataSources []ServiceDataSource
}
```

* This would make the interface more familiar for a k8s user. 
* Adding different data source types later would be a bit more structured
* maybe it represents other hard earned knowledge of the k8s API designers but I'm not sure.

Overall this is not a big difference so not doing this is also OK to me now.

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -56,15 +56,37 @@ type OpenstackDataPlaneServiceCert struct {
...
+	// Name of the ConfigMap
+	Name string `json:"name"`
+
+	// Required is whether the ConfigMap is required to exist
+	// +kubebuilder:validation:Optional
+	Required bool `json:"required,omitempty"`
```
gibizer, line 67: We might want to make these references required by default as that is the more common usage of a reference.

Just be careful that adding a default true here will not work out of the box as omitempty will prevent the false value to be ever set.

gibizer (commented):
I have couple of conce

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -56,15 +56,37 @@ type OpenstackDataPlaneServiceCert struct {
...
 type OpenStackDataPlaneServiceSpec struct {
 	// ConfigMaps list of ConfigMap names to mount as ExtraMounts for the OpenStackAnsibleEE
 	// +kubebuilder:validation:Optional
-	ConfigMaps []string `json:"configMaps,omitempty" yaml:"configMaps,omitempty"`
+	ConfigMaps []OpenStackDataPlaneConfigMap `json:"configMaps,omitempty" yaml:"configMaps,omitempty"`
 
 	// Secrets list of Secret names to mount as ExtraMounts for the OpenStackAnsibleEE
 	// +kubebuilder:validation:Optional
-	Secrets []string `json:"secrets,omitempty"`
+	Secrets []OpenStackDataPlaneSecret `json:"secrets,omitempty"`
```
fao89, line 89: instead of having 2 lists, maybe we can reuse the struct from AnsibleVarsFrom: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/common.go#L28-L38

SeanMooney: we shoudl not do that unless we entirely rename that class

the use case we are trying to enable is the creation of optional configmaps/secrets that will contain file that will be copied to the edpm nodes 

so without removing the AnsibleVars part for he class name this would be very confusing to anyone looking at the crd.

we are not trying to pass info to ansibele via the inventory we are trying to pass file to the deployed service.

fao89: yes, definitely we would need to rename the class

api/v1beta1/openstackdataplaneservice_types.go
slagle: I've combined this suggestion with the one from @fao89 to use the same struct as AnsibleVarsFrom, and refactored the common code. I've also left the existing ConfigMaps/Secrets fields, so that we can merge this with backwards compatibility, then clean those up after migrating cifmw/install_yamls/etc.

fao89 (commented):
the service samples need to be updated

api/v1beta1/openstackdataplaneservice_types.go
gibizer: the new struct looks OK to me. Also agree about the backward compatible change strategy.

pkg/deployment/hashes.go
```diff
@@ -59,15 +59,15 @@ func GetDeploymentHashesForService(
 			helper.GetLogger().Error(err, "Unable to retrieve ConfigMap %v")
 			return err
 		}
-		configMapHashes[cmName], err = configmap.Hash(cm)
+		configMapHashes[serviceCM.Name], err = configmap.Hash(cm)
 		if err != nil {
 			helper.GetLogger().Error(err, "Unable to hash ConfigMap %v")
```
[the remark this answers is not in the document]
slagle: this is dealing with the backwards compatible ConfigMaps, so I'll leave it as-is for now. I've put the variable names back to what they were.

pkg/util/datasource.go
```diff
@@ -0,0 +1,68 @@
...
+		cm := dataSource.ConfigMapRef
+		optional := cm.Optional != nil && *cm.Optional
+		configMap := &v1.ConfigMap{}
+		err := client.Get(ctx, types.NamespacedName{Name: cm.Name, Namespace: namespace}, configMap)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) && optional {
+				// ignore error when marked optional
+				utils.LogErrorForObject(helper, err, "Optional ConfigMap not found", configMap)
+			}
+			return configMap, secret, err
```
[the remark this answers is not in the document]
slagle: Indeed. Thanks for spotting :).

slagle:
> the service samples need to be updated

i may do that in the followup that drops the backwards compatibility if you don't object. Just to make sure that's still working.

fao89 (approved):

PR #922, opened 2024-06-06
Rename Play parameter to PlaybookContents

bshephar:
We have a "Play" parameter in our Services that is leveraged to provide a in-line Ansible playbook. This parameter is ultimately passed into the AEE container using the RUNNER_PLAYBOOK env var. Given that an Ansible Play is a single component of a playbook. This change rectifies our use of Ansible terminology and provides a more accurate description of what the field expects to receive from the user.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/847

### DIFF
--- a/api/v1beta1/openstackdataplaneservice_types.go
+++ b/api/v1beta1/openstackdataplaneservice_types.go
@@ -74,8 +74,8 @@ type OpenStackDataPlaneServiceSpec struct {
 	// +kubebuilder:validation:Optional
 	TLSCerts map[string]OpenstackDataPlaneServiceCert `json:"tlsCerts,omitempty" yaml:"tlsCerts,omitempty"`
 
-	// Play is an inline playbook contents that ansible will run on execution.
-	Play string `json:"play,omitempty"`
+	// PlaybookContents is an inline playbook contents that ansible will run on execution.
+	PlaybookContents string `json:"playbookContents,omitempty"`
 
 	// Playbook is a path to the playbook that ansible will run on this execution
 	Playbook string `json:"playbook,omitempty"`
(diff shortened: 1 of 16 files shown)


fao89:
CI will probably need a change here: https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/edpm/services/dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml#L14

bshephar:
recheck

api/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -75,7 +75,7 @@ type OpenStackDataPlaneServiceSpec struct {
 	TLSCerts map[string]OpenstackDataPlaneServiceCert `json:"tlsCerts,omitempty" yaml:"tlsCerts,omitempty"`
 
 	// Play is an inline playbook contents that ansible will run on execution.
```
slagle, line 77: Need to update the comment as well.

docs/assemblies/custom_resources.adoc
```diff
@@ -319,11 +319,6 @@ OpenStackDataPlaneServiceSpec defines the desired state of OpenStackDataPlaneSer
 | map[string]<<openstackdataplaneservicecert,OpenstackDataPlaneServiceCert>>
 | false
 
-| play
-| Play is an inline playbook contents that ansible will run on execution.
-| string
-| false
-
```
slagle, line 326: I spent a couple of minutes looking into why the new field doesn't show up here. And afaict, it looks like there might be a bug in crd-to-markdown in how it's looking for the "inline" string. Because if that string is part of the field name, it won't render the field.

we  might need to change the name of the field to work around this. PlaybookContents perhaps.

bshephar:
recheck

rabi (commented):
/lgtm

PR #926, opened 2024-06-07
Order nodes  when creating dnsdata

rabi:
Otherwise dnsdata could be different in every reconciliation and this would result in a reconciliation loop as
- NodeSet watches dnsmasq changes
- dnsmasq watches dnsdata changes

Also adds a multinode kuttl test where we can test these things.


rabi:
recheck

gibizer (approved):
make sense.

PR #928, opened 2024-06-07
Make TLS cert mount order consisitent

vakwetu:
We iterate over the TLSCerts map to define the cert mounts.  Because this is a map, though, the order is undefined.  This causes kuttl tests to fail every so often because they expect a consistent mount order.

The simplest solution is to sort the map keys and then add the mounts in a consistent way.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -273,7 +273,14 @@ func (d *Deployer) addCertMounts(
 		}
 
 		if service.Spec.TLSCerts != nil {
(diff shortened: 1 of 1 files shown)


fao89 (approved):

slagle (approved):

vakwetu:
@slagle @fao89  the kuttl test failure reported above looks totally unrelated to my change.  The failure occurs in a test suite in which tlsEnabled is set to False, which would mean that the code I changed wouldn't even be executed.

PR #931, opened 2024-06-10
Cleanup/fix how we manage DeploymentReadyCondition

rabi:
- Don't use deployemnt.Deployed as there could be more than one nodeset in a deployment.
- Check NodeSetDeploymentReadyCondition for the nodeset rather than Ready condition for the deployment.
- Break when a failed deployment for a nodeset is encountered. Continuing can overwrite the nodeset deployment conditions.

Jira: https://issues.redhat.com/browse/OSPRH-7617
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/847

### DIFF
--- a/controllers/openstackdataplanenodeset_controller.go
+++ b/controllers/openstackdataplanenodeset_controller.go
@@ -396,41 +396,32 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		Log.Error(err, "Unable to get deployed OpenStackDataPlaneDeployments.")
 		return ctrl.Result{}, err
 	}
-	deployErrMsg := ""
-	if isDeploymentFailed {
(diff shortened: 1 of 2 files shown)


rabi:
zuul jobs require https://github.com/openstack-k8s-operators/install_yamls/pull/847

rabi:
recheck

rabi:
> > Break when a failed deployment for a nodeset is encountered
> 
> this aspect looks missing in the source jira issue. Wouldn't it change the users expectations about tolerating a partial failure of some edpm nodes during rhoso deployment, which also had been demoed recently?

Partial failure is handled in ansible and AEE jobs won't fail if max_fail_percentage  > 0 and failed node count for a play is within the tolerated value.

fao89 (approved):

fao89:
recheck

PR #933, opened 2024-06-13
Fix remaining kuttl tests "play" fields

bshephar:
These should now be "playbookContents"


bshephar:
Eh, these are the AEE resources we're checking. So it's still play until we fix openstack-ansibleee-operator...

